PR #51, opened 2014-07-01
Encrypt content type

dkg:


dkg:
This pull request has been updated to the current master

draft-ietf-tls-tls13.md
```diff
@@ -1010,17 +1010,14 @@ of {{RFC5116}}. The key is either the client_write_key or the server_write_key.
 
 %%% Record Layer
        struct {
-           ContentType type;
```
ekr: I thought we had agreed to have a dummy type outside the record, so that it doesn't break everyone's parser.

draft-ietf-tls-tls13.md
```diff
@@ -1028,11 +1025,15 @@ record_version
 
 length
 : The length (in bytes) of the following TLSCiphertext.fragment.  The length
-  MUST NOT exceed 2^14 + 256.  An endpoint that receives a record that exceeds
+  MUST NOT exceed 2^14 + 257.  An endpoint that receives a record that exceeds
```
ekr: I suggest that we leave this as 256, since it was rough anway.

draft-ietf-tls-tls13.md
```diff
@@ -1083,19 +1085,29 @@ In order to decrypt and verify, the cipher takes as input the key, nonce, the
...
-
-As a special case, we define the NULL_NULL AEAD cipher which is simply
-the identity operation and thus provides no security. This cipher
-MUST ONLY be used with the initial TLS_NULL_WITH_NULL_NULL cipher suite.
+An AEAD cipher MUST NOT produce an expansion of greater than 256
+bytes.  An endpoint that receives a record from its peer with
+TLSCipherText.length larger than 2^14 + 257 octets MUST generate a
+fatal "record_overflow" alert.  This limit is derived from the maximum
+TLSPlaintext length of 2^14 octets + 1 octet for ContentType + the
+maximum AEAD expansion of 256 octets.
```
ekr: See above.

draft-ietf-tls-tls13.md
davegarrett: I thought the plan was just to do all the changes and see if anything important breaks that needs a kludge to deal with. I think you said this at one of the meetings, if I'm remembering from the video correctly.

ekr:
Consensus at meeting to merge this PR with updates and with the dummy type byte restored.

@ekr has action item to actually measure whether we can strip off the now fixed type, record.version bytes.

dkg:
I've updated this with the expected changes: dummy external ContentType octet is restored, and fragment.type is after fragment.content instead of the other way around (to make it easier to encrypt without having to shuffle buffers).

dkg:
I agree with davegarrett's suggestion about field naming here, and have supplied a change implementing it.

PR #94, opened 2014-11-13
Issue91 add update mechanism

ekr:
WIP for WG discussion. Not finished.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -312,6 +312,8 @@ draft-05
 
 - Prohibit SSL negotiation for backwards compatibility.
 
+- Add support for Update protocol
+
 draft-04
 
 - Modify key computations to include session hash.
@@ -327,7 +329,6 @@ draft-04
 - Update format of signatures with context.
 
 - Remove point format negotiation.
-
 
 draft-03
 
@@ -951,7 +952,7 @@ message MAY be fragmented across several records).
 
        enum {
            reserved(20), alert(21), handshake(22),
-           application_data(23), (255)
+           application_data(23), update(24), (255)
        } ContentType;
 
        struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -951,7 +952,7 @@ message MAY be fragmented across several records).
 
        enum {
            reserved(20), alert(21), handshake(22),
-           application_data(23), (255)
+           application_data(23), update(24), (255)
```
martinthomson: I'm going to argue for this being a handshake message later, but this is fine for now.

draft-ietf-tls-tls13.md
```diff
@@ -1096,8 +1097,10 @@ require a client write IV and a server write IV.
 
 When keys are generated, the current master secret (MS) is used
 as an entropy source. For handshake records, this means the
-hs_master_secret. For application data records, this means the
-regular master_secret.
+hs_master_secret.
+For application data, records, this means either
```
martinthomson: Extra comma

draft-ietf-tls-tls13.md
```diff
@@ -2950,6 +2956,120 @@ complete picture is that ECDH is employed with a non-trivial KDF
...
+  
+             /* This structure may be extended */
+           }
+        } Update;
+
+Whenever a site sends an Update message, it updates its own master secret
+as follows:
+
+      X_master_secret_{i+1} = PRF(X_master_secret_i, "master_secret_update",
+                                  Update_message)
```
martinthomson: This could use "client master secret" or "server master secret" as above.  You should also probably be more consistent in the naming of these things.

I've argued before for a table somewhere listing both the PRF and the signature discriminator strings.  Might be valuable as an appendix.  Once this settles, perhaps I can generate you a PR.

draft-ietf-tls-tls13.md
```diff
@@ -2950,6 +2956,120 @@ complete picture is that ECDH is employed with a non-trivial KDF
...
+
+This new master secret is used to generate traffic keys for subsequent
+records that it sends. Similarly, when a site receives an Update
+message, it updates its view of the peer's master secret and uses that
+for new traffic that it receives from the peer. The previous
+master secret MUST be discarded. [[TODO: DTLS?]]
```
martinthomson: I think that we need to talk about traffic keys and master secret separately here.  That is, you hold the master secret that will be used to generate the *next* master secret and traffic keys, plus the current set of traffic keys.  The current traffic keys are derived from the *last* master secret (which has since been discarded).

PR #105, opened 2014-12-23
remove SSL 2 backwards compatibility section & prohibit SSL negotiation

davegarrett:
A straightforward cleanup of part of the backwards compatibility section. (handles an easy part of issue #54) SSL 2.0 backwards compatibility was previously deprecated in TLS 1.2 (2008) and was warned to be "phased out with all due haste". RFC 6176 (2011) prohibits it entirely. It's a section of obsolete complexity still in the draft. Remove it and replace it with a simple "MUST NOT" send or accept.

-----
Edit to correct attribution of the "phased out with all due haste" language. Whilst in the TLS 1.2 (2008) spec, it was first introduced in SSL 3 (1996). It has been around for 19 years, not 5.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3592,7 +3592,7 @@ document.
 ## Compatibility with SSL
 
 The security of SSL 2.0 {{SSL2}} is considered insufficient for the reasons enumerated
-in [RFC6176].
+in [RFC6176], and MUST NOT be negotiated for any reason.
 
 Implementations MUST NOT send an SSL version 2.0 compatible CLIENT-HELLO.
 Implementations MUST NOT negotiate TLS 1.3 or later using an SSL version 2.0 compatible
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3592,104 +3589,23 @@ ClientHello.client_version. No single value will guarantee interoperability
...
+in [I-D.ietf-tls-sslv3-diediedie].
 
-When a TLS-capable server negotiates SSL 2.0 it SHOULD, after decrypting the
-ENCRYPTED-KEY-DATA field, check that these 8 padding bytes are 0x03. If they
-are not, the server SHOULD generate a random value for SECRET-KEY-DATA, and
-continue the handshake (which will eventually fail since the keys will not
-match). Note that reporting the error situation to the client could make the
-server vulnerable to attacks described in {{BLEI}}.
+Implementations MUST NOT send a ClientHello or ServerHello with the protocol version set
+to { 3, 0 } or less. Any endpoint receiving a Hello message with the protocol version set
```
ekr: I think this would be clearer if it read ClientHello.version or ServerHello.version, since then it would be double clear it's not the record layer.

PR #107, opened 2014-12-24
revise backwards compatibility & fix record layer version

davegarrett:
This does the following:
1) Reorganize the backwards compatibility section into subsections for easier readability
2) Freezes the ClientHello record layer version to TLS 1.0 (its effective compatible version)
3) Adds explicit requirement that all other record layer versions match the negotiated version

This plus PR #105 should be sufficient to close issue #54.

The record layer requirements were proposed by Brian Smith on the list.
http://www.ietf.org/mail-archive/web/tls/current/msg14870.html
(also mentioned by him in other discussions)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3418,19 +3418,19 @@ Cryptographic details:
 
 ## Compatibility with prior versions {#compatibility}
 
-[[TODO: Revise backward compatibility section for TLS 1.3.
-https://github.com/tlswg/tls13-spec/issues/54]]
 Since there are various versions of TLS (1.0, 1.1, 1.2, 1.3, and any future
 versions) and SSL (2.0 and 3.0), means are needed to negotiate the specific
 protocol version to use. The TLS protocol provides a built-in mechanism for
 version negotiation so as not to bother other protocol components with the
 complexities of version selection.
 
-TLS versions 1.0, 1.1, and 1.2, and SSL 3.0 are very similar, and use
-compatible ClientHello messages; thus, supporting all of them is relatively
-easy. Similarly, servers can easily handle clients trying to use future
-versions of TLS as long as the ClientHello format remains compatible, and the
-client supports the highest protocol version available in the server.
+TLS 1.x and SSL 3.0 use compatible ClientHello messages, so supporting all
+of them is relatively easy. Similarly, servers can easily handle clients trying
+to use future versions of TLS as long as the ClientHello format remains
+compatible, and the client supports the highest protocol version available
+in the server.
+
+### Negotiating with an older server
 
 A TLS 1.3 client who wishes to negotiate with such older servers will send a
 normal TLS 1.3 ClientHello, containing { 3, 4 } (TLS 1.3) in
@@ -3447,20 +3447,30 @@ If a TLS server receives a ClientHello containing a version number greater than
 the highest version supported by the server, it MUST reply according to the
 highest version supported by the server.
 
+### Negotiating with an older client
+
 A TLS server can also receive a ClientHello containing a version number smaller
 than the highest supported version. If the server wishes to negotiate with old
 clients, it will proceed as appropriate for the highest version supported by
 the server that is not greater than ClientHello.client_version. For example, if
 the server supports TLS 1.0, 1.1, and 1.2, and client_version is TLS 1.0, the
-server will proceed with a TLS 1.0 ServerHello. If server supports (or is
-willing to use) only versions greater than client_version, it MUST send a
-"protocol_version" alert message and close the connection.
+server will proceed with a TLS 1.0 ServerHello. If the server only supports
+versions greater than client_version, it MUST send a "protocol_version"
+alert message and close the connection.
 
 Whenever a client already knows the highest protocol version known to a server
 (for example, when resuming a session), it SHOULD initiate the connection in
 that native protocol.
 
-Note: some server implementations are known to implement version negotiation
+Earlier versions of the TLS specification were not fully clear on what the
+record layer version number (TLSPlaintext.version) should contain when sending
+ClientHello (i.e., before it is known which version of the protocol will be
+employed). Thus, TLS servers compliant with this specification MUST accept any
+value { 03, XX } as the record layer version number for ClientHello.
+
+### Negotiating with buggy servers
+
+Some server implementations are known to implement version negotiation
 incorrectly. For example, there are buggy TLS 1.0 servers that simply close the
 connection when the client offers a version newer than TLS 1.0. Also, it is
 known that some servers will refuse the connection if any TLS extensions are
@@ -3468,14 +3478,8 @@ included in ClientHello. Interoperability with such buggy servers is a complex
 topic beyond the scope of this document, and may require multiple connection
 attempts by the client.
 
-Earlier versions of the TLS specification were not fully clear on what the
-record layer version number (TLSPlaintext.version) should contain when sending
-ClientHello (i.e., before it is known which version of the protocol will be
-employed). Thus, TLS servers compliant with this specification MUST accept any
-value {03,XX} as the record layer version number for ClientHello.
-
 TLS clients that wish to negotiate with older servers MAY send any value
-{03,XX} as the record layer version number. Typical values would be {03,00},
+{ 03, XX } as the record layer version number. Typical values would be { 3, 0 },
 the lowest version number supported by the client, and the value of
 ClientHello.client_version. No single value will guarantee interoperability
 with all old servers, but this is a complex topic beyond the scope of this
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -996,8 +996,8 @@ TLS servers MUST accept any value { 03, XX } for this version in a ClientHello.
 See {{backward-compatibility}} for more information regarding
 negotiations with endpoints supporting other versions.
 
-Implementations MUST NOT send zero-length fragments of Handshake, Alert, or
-ChangeCipherSpec content types. Zero-length fragments of Application data MAY
+Implementations MUST NOT send zero-length fragments of Handshake or Alert
+types. Zero-length fragments of Application data MAY
 be sent as they are potentially useful as a traffic analysis countermeasure.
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -930,14 +933,10 @@ type
...
-  version value 3.4 is historical, deriving from the use of {3, 1}
-  for TLS 1.0.  (See {{record-layer-1}}.)  Note that a client that
-  supports multiple versions of TLS may not know what version will
-  be employed before it receives the ServerHello.  See
-  {{backward-compatibility}} for discussion about what record layer
-  version number should be employed for ClientHello.
+: The version of the protocol being employed in the current record.
+  For the initial ClientHello this MUST be { 3, 1 }.
+  For the ServerHello and all other TLS records this MUST be equal to the negotiated version.
+  For TLS 1.3 connections, this version is { 3, 4 }.
```
jsalowey: I think we should add the bit about the server being permissive about versions in he client hello from the appendix to here.
Earlier versions of the TLS specification were not fully clear on what the
 record layer version number (TLSPlaintext.version) should contain when sending
 ClientHello (i.e., before it is known which version of the protocol will be
 employed). Thus, TLS servers compliant with this specification MUST accept any value { 03, XX } as the record layer version number for ClientHello.

draft-ietf-tls-tls13.md
```diff
@@ -930,14 +933,10 @@ type
...
-: The version of the protocol being employed.  This document
-  describes TLS Version 1.3, which uses the version { 3, 4 }.  The
-  version value 3.4 is historical, deriving from the use of {3, 1}
-  for TLS 1.0.  (See {{record-layer-1}}.)  Note that a client that
-  supports multiple versions of TLS may not know what version will
-  be employed before it receives the ServerHello.  See
-  {{backward-compatibility}} for discussion about what record layer
-  version number should be employed for ClientHello.
+: The version of the protocol being employed in the current record.
+  For the initial ClientHello this MUST be { 3, 1 }.
```
jsalowey: Perhaps say 
"The client MUST set the version to {3, 1} for the initial ClientHello."

draft-ietf-tls-tls13.md
```diff
@@ -3558,39 +3570,48 @@ If a TLS server receives a ClientHello containing a version number greater than
...
+If an implementation negotiates usage of TLS 1.2, then negotiation of cipher
+suites also supported by TLS 1.3 SHOULD be preferred, if available.
+
+Implementations MUST NOT offer or negotiate RC4 cipher suites for any version.
+[I-D.ietf-tls-prohibiting-rc4]
+
+Old versions of TLS permitted the usage of very low strength ciphers.
+Ciphers with a strength less than 100 bits MUST NOT be offered or
+negotiated for any version.
+
```
jsalowey: draft-ietf-uta-tls-bcp-08 recommends 112.

davegarrett:
RE: "This document describes...for SSL 3.0."
I disagree that it is unnecessary. Whilst redundant, I think the reason why TLS uses weird version numbers needs to be stated somewhere in this document. Moving it to the backwards compatibility section is an option if you think it's in the way up here.

RE: second section of record layer version text
I have that set aside due to formatting concerns. The "version" section is indented in a list and putting lots of text in there risks cluttering things up a bit. I'd prefer to have the minimum definition and requirements under that heading and additional issues separately, in order to maximize readability.

davegarrett:
Well crud. I rebased this and GitHub dropped the comments from @briansmith. They haven't vanished entirely into the aether, just not shown above in the threading here. For reference, imagine there are proper "commented on an outdated diff" lines above and that I didn't confuse GitHub and am not talking to myself. I shall endeavor to be less of a GitHub noob in the future.

https://github.com/davegarrett/tls13-spec/commit/959dda3d7e7cf52ee78e767fafbbc66c9fd129f0#commitcomment-9100761
https://github.com/davegarrett/tls13-spec/commit/959dda3d7e7cf52ee78e767fafbbc66c9fd129f0#commitcomment-9100772
https://github.com/davegarrett/tls13-spec/commit/51d2686e791ff22e990ed0348fcb99df35314146#commitcomment-9100802
https://github.com/davegarrett/tls13-spec/commit/51d2686e791ff22e990ed0348fcb99df35314146#commitcomment-9100813
https://github.com/davegarrett/tls13-spec/commit/15fdc807377976e97a8471aa850bae2439ff67a2#commitcomment-9444688
https://github.com/davegarrett/tls13-spec/commit/15fdc807377976e97a8471aa850bae2439ff67a2#commitcomment-9444713

draft-ietf-tls-tls13.md
```diff
@@ -965,14 +968,10 @@ type
...
-  describes TLS Version 1.3, which uses the version { 3, 4 }.  The
-  version value 3.4 is historical, deriving from the use of {3, 1}
-  for TLS 1.0.  (See {{record-layer-1}}.)  Note that a client that
-  supports multiple versions of TLS may not know what version will
-  be employed before it receives the ServerHello.  See
-  {{backward-compatibility}} for discussion about what record layer
-  version number should be employed for ClientHello.
+: The version of the protocol being employed in the current record.
+  TLS clients MUST set this version to { 3, 1 } for the initial ClientHello.
+  For the ServerHello and all other TLS records this MUST be equal to the negotiated version.
```
davidben: This rule isn't sufficient to hammer down the record-layer version. There's also alerts sent:
* By the server in response to a protocol error in trying to read ClientHello.
* By the client in response to a protocol error in trying to read ServerHello.
* By the server in response to ClientHello with a bad version
* By the client in response to ServerHello with a bad version
(And conversely what versions the peer will accept.)

The first two are unambiguously unnegotiated, so I think either we shouldn't bother sending anything and drop the connection or the version should be {3, 1}.

Making the third {3, 1} is nice and consistent with the other unnegotiated cases. The client should accept any version (it hasn't pinned a version yet anyway). Also most server implementations will resolve the version before the cipher suite, and it would be unnecessary complexity on the server's end to require that an alert on no valid cipher suite still use {3, 1} rather than the version it negotiated.

The fourth is tricky. If the client were to send protocol_version because the version is too low and we wish for the server to actually read it, at this point the server has already switched versions. So it won't accept any record with something other than the version the server negotiated. (And recommending new servers accept that is unnecessary complication in the state machine.) Thus it should be the version the server negotiated and client implementations should take care that, when programming their record layer to echo that version, that it be treated entirely opaque and never condition on it. (Lest it, say, accidentally act on being "in SSLv3" when SSLv3 is disabled or unimplemented.)

Hammering these down would avoid the silliness where FALLBACK_SCSV's spec has its own language to specify the record-layer version.

davegarrett:
It might be simpler to just change to:

+  TLS clients MUST set this version to { 3, 1 } for the initial ClientHello and all alert messages.
+  For the ServerHello and all other TLS records this MUST be equal to the negotiated version.

Additionally, dropping the version field from TLSCipherText entirely might be worth considering, but that's outside the scope of this here.

davidben:
(I assume you mean s/clients/implementations/? Otherwise there's still an unspecified hole.)

All alerts period, or just those before version negotiation? OpenSSL, after version negotiation, will currently reject any record (of whatever type) whose version does not match the negotiated version. Go's crypto/tls does the same thing. I haven't done a survey of other implementations yet. It can certainly change for TLS 1.3. But, as an implementer, I would prefer that field be hammered down in a way that can advise older versions, whether or not 1.3 additionally clamps it post-negotiation.

If just those before version negotiation, I think a cleaner statement of that rule is:
* Prior to version negotiation, this version MUST by {3, 1}. This includes the initial ClientHello and any alerts sent during version negotiation.
* After version negotiation, this MUST be equal to the negotiated version. This includes the ServerHello and all following records.

This is fairly straightforward and clean. The record-layer initializes in "proto TLS" which uses {3, 1}. Then it switches one to a concrete version. The one ambiguity left is a client's `protocol_version` alert. That one is sent when one side has done negotiation, but not the other. Which is where the trickiness I mentioned comes up with existing servers[*]. Whether the complexity is worth bothering with is another matter.

(For context, I'm coming at this from having recently rewritten BoringSSL's version negotiation logic. 1.2's failure to clearly specify version negotiation was a huge pain point.)

[*] For completeness, OpenSSL client has its own quirks. If the client has a minimum version > TLS 1.0 AND the server sends some alert with record-layer {3, 1}, the client will fail to read it. OpenSSL client negotiates based on record-layer version, not ServerHello. But that implementation is sufficiently weird that I don't think it's worth bothering about. See links in #136.

davegarrett:
I was proposing {3,1} for all alerts, yes. (even old versions should be able to parse new; an unknown error could just be generically acted on) However, if there are implementations that would find this inconsistent with current behavior, then I agree it's not worth it.

I like the language you suggested. It deals with the error cases and is quite clear. (modulo the typo ;)

To deal with the protocol_version case, this might cover everything:

+  Prior to version negotiation, this value MUST be { 3, 1 }. This includes the ClientHello, records sent during negotiation, and errors aborting the negotiation.
+  After version negotiation, this MUST be equal to the negotiated version. This includes the ServerHello and all following records.

davegarrett:
"records sent during negotiation" would count EarlyDataExtension records as well. Arguably this should be { 3, 4 }, but I don't think it's worth the complexity.

grubba:
Version negotiation is completed by the first packet that the server sends (ie typically ServerHello), and the client typically only sends one packet if it cares about compatibility with versions prior to TLS 1.3, so forcing the packet version to {3, 1} for the ClientHello should be sufficient.

draft-ietf-tls-tls13.md
ekr: Proposed resolution at interim:
1. MUST always send {3, 1} for every message.
2. MUST always ignore.

davegarrett: It's generally been my preferred route, but was argued against by an implementer in some prior discussion.

I am very much in favor of just freezing for all records and not considering it useful information at all. This is the ideal solution, IMO.

I've also proposed dropping the field entirely from encrypted records.

davegarrett:
Per proposed resolution at interim, we can now simplify things down to one frozen and deprecated value. This is the ideal solution, IMO. In addition to updates for this, I've also renamed the field to "record_version" instead of just "version" as people easily mix up the various version fields. It now has a different expected behavior, so a different name is appropriate.

davidben:
I'm fine with freezing it going forward (it means that someday in the far far future perhaps we can forget about this craziness altogether). But note that implementations supporting TLS < 1.3 will still need to do a swap when negotiating older versions. Existing clients and servers (including anything derived from OpenSSL) will reject post-version-negotiation records that don't have the right record-layer version. I don't know if advice on that front belongs in this draft or elsewhere or nowhere, but it is a concern implementations will have for a while.

davegarrett:
Good point. I've updated backwards compatibility section for that.

davegarrett:
More fixes and rebasing.

davegarrett:
Note that I expect a new version negotiation mechanism implemented via an extension will be desired in order to deal with TLS 1.3 intolerance, and thus parts of this will be changed again. ClientHello.client_version would most likely be frozen to {3,3} in such a mechanism. That, however, is outside the scope of this PR which is to revise the existing backwards compatibility and freeze the record layer version.

PR #122, opened 2015-01-06
First cut at auto-generating appendix

ekr:
The appendix kept getting out of sync, so autogenerate it

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ next := $(draft)-$(next_ver)
 COMMIT=origin/master
 
 .PHONY: all latest submit clean
+.INTERMEDIATE: %.md2
 
 all latest: $(draft).txt $(draft).html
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -48,7 +48,11 @@ diff-commit: $(draft).txt $(draft)-$(COMMIT).txt
 $(next).md: $(draft).md
 	sed -e"s/$(basename $<)-latest/$(basename $@)/" $< > $@
 
-%.xml: %.md
+%.md2: %.md
+	- rm $@
```
martinthomson: You shouldn't have to use rm here.  Try adding
.INTERMEDIATE: %.md2

And add *.md2 to your clean target.

mk-appendix.py
```diff
@@ -0,0 +1,59 @@
...
+                print_syntax(APPENDICES[m.group(2)])
+                del APPENDICES[m.group(2)]
+                print
+            else:
+                print l,
+    else:
+        print l,
+        m = re.match("\S", l)
+        if m is None:
+            CURRENT += l
```
martinthomson: So a figure can't include a blank line if it is going to be included in the appendix?

martinthomson:
Not how I would have done it, but if it works, then it's fine.

mk-appendix.py
ekr: No, it can. It can't include a line with a leading non-space character.

PR #124, opened 2015-01-08
Issue103: Suggestion of how to clarify master secret computations.

grubba:
Also fixes some minor related issues.


davegarrett:
@ekr: I think this can be closed. The current version is a bit more clear and I don't think this would help. It adds quite a bit more to the flow diagrams, so the extra complexity actually could be unnecessarily confusing. It's also quite obsolete now and needs a new PR to consider, anyway.

ekr:
Closing. Will try to clean up this text in the big rewrite

PR #129, opened 2015-01-22
Improve alerts enums with error type as comments

beurdouche:
I have seen multiple implementations with messages marked as warning instead of fatal..
It might be more explicit this way ?

B.


seanturner:
I like this because some folks never read the text they just look at the code.

PR #134, opened 2015-01-22
Fix alert names when referred to

beurdouche:
Sometimes the alert name is in between quotes like : "protocol_version" alert ..
and sometimes it is not ... like : close_notify alert


beurdouche:
There is something else that should be fixed ...
Sometimes the alert name is in between quotes like "protocol_version" alert ..
and sometimes it is not like : close_notify 

I could make all that uniform with or without quotes.
my preference goes to adding quotes around all ...

beurdouche:
I made all alert names citations consistent across the text by adding quotes when missing
B.

beurdouche:
First commit : 
fix typos

Second commit :
I made all alert names citations consistent across the text by adding quotes when missing

B.

seanturner:
BTW - I noted the same inconsistencies.

PR #142, opened 2015-03-07
Specify a Padding Mechanism to 1.3

tomrittervg:


grubba:
From the patch:
> handshake_padded(24),

This assignment conflicts with RFC 6520 (heartbeat).

grubba:
Another issue with handshake_padded is that it adds unnecessary complexity to the handshake handling, as it is no longer possible to just add the decoded handshake packets to the handshake hash, as the padding and padding length will interleave with the handshake messages. The alternative would be to force handshake packet breaks between handshake messages where the handshake hash is needed, but that would be contra-productive and reveal more of the contents than before.

I suggest replacing the handshake_padded packet with a handshake_padding message, as this would have clear semantics, have essentially equivalent properties after encryption but reveal less of the message breaks and be trivial to implement.

grubba:
"Clear semantics"; there would still be an issue with handling of padding between flights.

I suggest that any padding after the last message of a flight MUST be in the same handshake packet as the last message. Any padding in later handshake packets would be considered part of the next flight.

tomrittervg:
Adding a separate handshake_padding ContentType with no data would not be helpful, as the ContentType is in the clear. You'd see exactly how much padding there is, and ignore it in analysis.

I had originally written it as NOT including the handshake padding in the handshake messages hash, and we could go back to that, but it seems like a clear timing attack.  I think that would solve the interleaving issues as well as the question of "How do I know the handshake is done, or if someone is going to send me empty padding records?"

I didn't think it would be too much difficult to deal with the interleaving: simply add the decrypted record payload (which includes the padding_length and padding) to the hash - I assumed that was what most implementations were doing presently (adding the data before fully parsing it). Even if you did parse it, you should have that buffer around?  But maybe not.  

I agree that implementations MUST NOT send any trailing, padding-only handshake messages, to resolve that ambiguity.

davegarrett:
This would be a good reason to encrypt the content type. (PR # 51)

martinthomson:
Can we have this discussion on the [mailing list](mailto:tls@ietf.org) please?

grubba:
Tom: It seems you misunderstood my suggestion; I suggested a handshake message type of only padding (ie an entry in the HandshakeType enum), and not a packet type (ie an entry in the ContentType enum). I assume that you are aware that multiple handshake messages may be in the same handshake packet?

draft-ietf-tls-tls13.md
```diff
@@ -982,12 +1008,58 @@ fragment
...
+containing only padding. Implementations MAY send application_data_padded
+records containing only padding and no underlying data. Padded records
+MAY contain no padding at all.
+
+Choosing a padding algorithm that suggests when and how much to pad is
+outside the scope of the specification. Later documents may define padding
+selection algorithms, or define a padding policy request mechanism through
+TLS extensions or some other means.
+
+Throughout the document, whenthe handshake and application_data record layers
```
martinthomson: s/whenthe/when the/

draft-ietf-tls-tls13.md
```diff
@@ -982,12 +1008,58 @@ fragment
...
+padding present, followed by the application data, and finally the
+padding itself.
+
+[[[Options:
+ - PKCS5 style padding, where it's application data followed by two-byte
+   repeating octets that specify the length.
+ - Padding could be required to be all zeros but implementations
+   MAY NOT/SHOULD NOT verify this
+ - Padding could be all zeros and implementations MUST verify this
+
```
ekr: Conclusion from Interim: MUST be zero but SHOULD NOT verify.

draft-ietf-tls-tls13.md
```diff
@@ -951,14 +962,29 @@ message MAY be fragmented across several records).
 
        enum {
            reserved(20), alert(21), handshake(22),
-           application_data(23), (255)
+           application_data(23), handshake_padded(25),
+           application_data_padded(26), (255)
```
ekr: Conclusion from interim: encrypted handshake messages are always padded.

draft-ietf-tls-tls13.md
```diff
@@ -951,14 +962,29 @@ message MAY be fragmented across several records).
...
        } ContentType;
 
        struct {
+           uint16 padding_length;
+           opaque fragment[TLSPlaintext.length - padding_length];
+           opaque padding[padding_length];
+       } PaddedFragment;
```
ekr: Conclusion from interim: put the padding at the beginning, so you can just have
opaque padding<0..2^16-1>

grubba: This will still add needless complexity if the packet fragments don't align with handshake message borders (ie the padding is in the middle of a handshake message), and the handshake_hash needs to be calculated (as the handshake layer may not know anything about the packet layer).

tomrittervg:
See #147

PR #147, opened 2015-03-11
Padding

dkg:
revision of padding following the interim discussion (should replace PR #142)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1044,8 +1044,11 @@ to decrease the maximum size of fragment from 2^14 to 2^14 - 2 (because
 of the additional two bytes used to specify the padding length.)
 
 Implementations MUST accept handshake_padded and application_data_padded
-records, and MAY send them.  Padded records MAY contain no underlying data
-and only padding. Similarly, padded records MAY contain no padding at all.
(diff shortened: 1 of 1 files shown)


ekr:
@dkg We need a new PR that addresses the consensus from the meeting

draft-ietf-tls-tls13.md
```diff
@@ -1008,7 +1020,16 @@ of {{RFC5116}}. The key is either the client_write_key or the server_write_key.
...
-              opaque content[TLSPlaintext.length];
+               select(type) {
+                   case alert:
+                   case application_data:
+                       opaque content[TLSPlaintext.length];
+                   case handshake:
+                   case application_data_padded:
+                       opaque padding<0..2^16-1>;
+                       opaque content[TLSPlaintext.length -
+                                      (2 + length(padding))];
```
davegarrett: If I'm not misunderstanding things, this should be TLSCiphertext.length - 2 - length(padding), which is equal to TLSPlaintext.length (not plain, as it currently is on line 1030). I think it would be more clear to just have the padding element under a conditional and have only the one content[TLSPlaintext.length], with a note on how TLSPlaintext.length is determined.

ekr:
Done

PR #150, opened 2015-03-13
Referring to peers as parties.

seanturner:
If we go with PR #118, then we might consider switching to refer to two peers as opposed to parties.


davegarrett:
The term "endpoint" is a better drop-in replacement for "party". The definition in #118 has "peer" as equivalent to "endpoint" except when specifically discussing remote endpoint.

Some parts could ideally be using both terms, e.g.:
"the detecting party sends a message to the other party"
\/
"the detecting endpoint sends a message to its peer"

ekr:
I think I agree with dave about endpoint vs. peer. element would be ok too

PR #152, opened 2015-03-14
Some nits and some consistency fixes.

seanturner:
e.g.,:
r/handshake protocol/Handshake Protocol throughout
r/Server Key Exchange/ServerKeyShare
r/HelloRetry/HelloRetryRequest

PR #153, opened 2015-03-14
Alerts

seanturner:


ekr:
@seanturner is this still needed?

seanturner:
There�s some nits that aren�t in the editor�s working copy.  Since the travis CI build is failing do you want to reject this one and I can resubmit the ones that aren�t incorroprated?

spt

draft-ietf-tls-tls13.md
```diff
@@ -3566,7 +3563,7 @@ challenge
   not trailing) zero bytes to make it 32 bytes long.
 {:br }
 
-Note: Requests to resume a TLS session MUST use a TLS ClientHello.
+Note: Requests to resume a TLS session MUST use a TLS client hello.
```
davegarrett: I think it should stay as ClientHello/ServerHello where applicable, and only switch to "hello message" when it covers multiple. I think it's easier to keep track of things if the struct names are used explicitly when reference. Also makes searching the document for stuff about them a little easier.

davegarrett:
Being explicit about which alert to send when one is called for is very helpful. (I wouldn't call it a nit)

The "This message is always fatal" to "This alert is always fatal" conversion makes sense, but I added more of those lines so in addition to rebasing you'll need to change those too.

I guess it might be best to submit a new PR with just the set of those changes, tack some other little stuff onto PR #152, and then drop this one. Making Travis CI happy appears to be a PITA sometimes.

seanturner:
You can reject this PR all of the changes are included in PR#152.

PR #155, opened 2015-03-16
Remove explicit IV for AEAD algorithms

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -843,8 +842,8 @@ record protection algorithm
   do not provide any confidentiality and
   {{record-payload-protection}} defines a special NULL_NULL AEAD
   algorithm for use in the initial handshake). This specification
-  includes the key size of this algorithm and the lengths of explicit
-  and implicit initialization vectors (or nonces).
+  includes the key size of this algorithm and of nonce for
```
martinthomson: *the* nonce for the  AEAD.

draft-ietf-tls-tls13.md
```diff
@@ -1026,14 +1022,12 @@ fragment
...
-{{RFC5116}}; with record_iv_length being the length of the explicit part. In
-this case, the implicit part SHOULD be derived from key_block as
-client_write_iv and server_write_iv (as described in {{key-calculation}}), and
-the explicit part is included in GenericAEAEDCipher.nonce_explicit.
+The nonce for the AEAD construction is formed by taking the 64-bit
+sequence number (or the 64-bit epoch/sequence number combination
+in the case of DTLS) and padding it on the left with zeroes
+in order to obtain a string of size N_MIN (see {{RFC5116}} Section 4).
+Note: This is a different construction from that in TLS 1.2, which
+specified a partially explicit nonce.
```
martinthomson: Note also that an AEAD cipher that has N_MAX less than 64 cannot be used with TLS. ?

draft-ietf-tls-tls13.md
```diff
@@ -898,8 +897,6 @@ items (some of which are not required by all ciphers, and are thus empty):
```
davegarrett: "following four items" should be "following two items" now

Alternatively, just change line 895-900 to:
"The record layer will use the security parameters to generate the client and server write keys."
(with updated/removed parenthetical)

ekr:
@martinthomson PTAL

draft-ietf-tls-tls13.md
```diff
@@ -1136,11 +1139,10 @@ which is then partitioned as follows:
...
        client_write_key[SecurityParameters.enc_key_length]
        server_write_key[SecurityParameters.enc_key_length]
-       client_write_IV[SecurityParameters.fixed_iv_length]
-       server_write_IV[SecurityParameters.fixed_iv_length]
+       client_write_IV[SecurityParameters.iv_length]
+       server_write_IV[SecurityParameters.iv_length]
 
-Currently, the client_write_IV and server_write_IV are only generated for
-implicit nonce techniques as described in Section 3.2.1 of {{RFC5116}}.
+iv_length is set to the greater of N_MAX
```
ekr: This should actually just go away. It's a typo.

ekr:
maybe "max(64 bits, N_MIN)" would be better?

davegarrett:
That sounds fine, if that notation can be used. (I don't see it elsewhere, but I don't see why not)

Instead of 64 bits, it should be 8 bytes (or octets). N_MIN is measured in bytes, not bits.
https://www.ietf.org/rfc/rfc5116.txt

draft-ietf-tls-tls13.md
```diff
@@ -919,12 +916,12 @@ These parameters are defined in the presentation language as:
...
 
        client write key
        server write key
-       client write IV
-       server write IV
+       client write iv
+       server write iv 
 
 The client write parameters are used by the server when receiving and
 processing records and vice versa. The algorithm used for generating these
```
davegarrett: the line after this one needs a period to end this sentence

draft-ietf-tls-tls13.md
```diff
@@ -1051,14 +1047,22 @@ fragment
...
+
+The nonce for the AEAD construction is formed as follows:
+
+  1. iv_length is set to max(8 bytes, N_MIN) for the AEAD algorithm
+     (see {{RFC5116}} Section 4).
+
+  2. The 64-bit sequence number is padded to the left with zeroes
+     to iv_length and XORed with the fixed iv, which is of length
+     iv_length.
+
```
davegarrett: "the fixed iv" is unclear here. I assume you mean the "the static client or server iv" (or otherwise worded).

draft-ietf-tls-tls13.md
```diff
@@ -1051,14 +1047,22 @@ fragment
...
+The resulting quantity (of length iv_length) is used as the per-record
+nonce. An AEAD algorithm where N_MAX is less than 64 bits MUST not be
+used with TLS.
```
davegarrett: I think it needs to be written more clearly that this is an action done for each record. Every time, the iv needs XORing with the seq num, as that's what's incremented. It's stated earlier that the seq num may not rollover at 0xffffffffffffffff, and incrementing and rolling over the per-record nonce would be something that might get screwed up, not to mention the added check to make sure you don't go past the starting point (iv).

I'd separate out determination of iv_length from point 1 above and move the last sentence above here to the end of it. (also N_MAX is in bytes, sot the 64-bit requirement should probably be 8 bytes like N_MIN) Then:
"
The nonce for each record's AEAD construction is formed as follows:
1. The 64-bit sequence number is padded to the left with zeros up to iv_length.
2. The padded sequence number is XORed with the client or server write iv, depending on role.
"

I shall use the possibility that I might've misunderstood something here as potential evidence that it needs to be written out more explicitly so as to avoid misunderstanding. ;)

draft-ietf-tls-tls13.md
```diff
@@ -3184,24 +3183,10 @@ end-entity certificate remain.
...
-                    Key      Implicit IV   Explicit IV
-    Cipher         Material  Size          Size
-    ------------   --------  ----------    -----------
-    NULL               0          0             0
-    AES_128_GCM       16          4             8
-    AES_256_GCM       32          4             8
 
 Key Material
 : The number of bytes from the key_block that are used for
   generating the write keys.
```
davegarrett: This definition can be cut too. It's just for the table you're cutting above. The only other place in the document that says "key material" is in Key Calculation, and they're not really referring to the same thing.

draft-ietf-tls-tls13.md
```diff
@@ -919,12 +916,12 @@ These parameters are defined in the presentation language as:
...
 The record layer will use the security parameters to generate the following four
-items (some of which are not required by all ciphers, and are thus empty):
+items:
 
        client write key
        server write key
-       client write IV
-       server write IV
+       client write iv
+       server write iv
```
davegarrett: The nittiest nit there ever was: there is a trailing space here.

ekr:
Merged but with confusing git history due to pilot error.

PR #157, opened 2015-03-26
a big pile of minor fixes

davegarrett:
Capitalize the first letter of the sentence like the rest of the alert descriptions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,9 +2,9 @@
 
 Before submitting feedback, please familiarize yourself with our current issues
 list and review the [working
-group home page](http://trac.tools.ietf.org/wg/tls/trac/wiki). If you're
+group home page](https://trac.tools.ietf.org/wg/tls/trac/wiki). If you're
 new to this, you may also want to read the [Tao of the
-IETF](http://www.ietf.org/tao.html).
+IETF](https://www.ietf.org/tao.html).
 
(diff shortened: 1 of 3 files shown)


davegarrett:
I just noticed that @seanturner noticed this too and included it in PR #153 with quite a few more significant editorial improvements around alerts.

davegarrett:
Rather than creating new PRs, I've just added a couple other minor things to this one.

davegarrett:
I've just added to this two almost identical looking commits to upgrade the HTTP links to HTTPS in README.md and CONTRIBUTING.md. Can CONTRIBUTING.md just be deleted, or is it still needed for some reason?

martinthomson:
CONTRIBUTING.md is needed to generate the banner that you see.

I see no harm in including a link to CONTRIBUTING.md from README.md if you want to do that.

davegarrett:
Transclusion would be ideal, if that capability exists. I don't think changing it to just a link is needed. As long as they're both being used and both updated as-needed, I have nothing against just duplicating the content. Just checking to make sure.

davegarrett:
Huh. Looks like I've found the most pedantic errata ever: it says "see page 99" when it should've said "see page 100" in the TLS 1.2 spec. The most recent trivial fix in this PR changes that to an actual link to the section in question. I would guess it's not worth actually filing an errata ticket for that. ;)

PR #163, opened 2015-04-16
Fix typo

ametheus:
Does this make me a contibutor?

PR #165, opened 2015-04-25
Spellcheck and trim some end line

beurdouche:
(includes #163)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1000,7 +1000,7 @@ be sent as they are potentially useful as a traffic analysis countermeasure.
 
 The record protection functions translate a TLSPlaintext structure into a
 TLSCiphertext. The deprotection functions reverse the process. In TLS 1.3
-as opposed to previous versions of TLS, all ciphers are modelled as
+as opposed to previous versions of TLS, all ciphers are modeled as
 "Authenticated Encryption with Additional Data" (AEAD) {{RFC5116}}.
 AEAD functions provide a unified encryption and authentication
 operation which turns plaintext into authenticated ciphertext and
@@ -2303,7 +2303,7 @@ EarlyData, and aside from hashing them differently, MUST treat them as
 if they had been sent immediately after the ClientHello.
 
 Servers MUST NOT send the EarlyData extension. Negotiating TLS 1.3
-serves as acknowledgement that it was processed as described above.
+serves as acknowledgment that it was processed as described above.
 
 [[OPEN ISSUE: This is a fairly general mechanism which is possibly
 overkill in the 1-RTT case, where it would potentially be more
@@ -2734,7 +2734,7 @@ Meaning of this message:
 > This message conveys the client's certificate chain to the server; the server
 will use it when verifying the CertificateVerify message (when the client
 authentication is based on signing) or calculating the premaster secret (for
-non-ephemeral Diffie- Hellman). The certificate MUST be appropriate for the
+non-ephemeral Diffie-Hellman). The certificate MUST be appropriate for the
 negotiated cipher suite's key exchange algorithm, and any negotiated extensions.
 
 In particular:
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2369,7 +2369,7 @@ If the same extension appears in both locations, the client
 MUST rely only on the value in the EncryptedExtensions block.
 [[OPEN ISSUE: Should we just produce a canonical list of what
 goes where and have it be an error to have it in the wrong
-place? That seems simpler. Perhaps have a whitelist of which
+place? That seems simpler. Perhaps have a white list of which
```
davegarrett: This is correct as-is. "whitelist" is a compound word.
https://en.wikipedia.org/wiki/Whitelist

draft-ietf-tls-tls13.md
```diff
@@ -3340,14 +3340,14 @@ provides several recommendations to assist implementors.
 
 ## Random Number Generation and Seeding
 
-TLS requires a cryptographically secure pseudorandom number generator (PRNG).
+TLS requires a cryptographically secure pseudo-random number generator (PRNG).
```
davegarrett: This doesn't need a hyphen either.
https://en.wikipedia.org/wiki/Pseudorandomness

draft-ietf-tls-tls13.md
```diff
@@ -3340,14 +3340,14 @@ provides several recommendations to assist implementors.
...
 Care must be taken in designing and seeding PRNGs. PRNGs based on secure hash
 operations, most notably SHA-1, are acceptable, but cannot provide more
 security than the size of the random number generator state.
 
 To estimate the amount of seed material being produced, add the number of bits
 of unpredictable information in each seed byte. For example, keystroke timing
-values taken from a PC compatible's 18.2 Hz timer provide 1 or 2 secure bits
+values taken from a PC's compatible 18.2 Hz timer provide 1 or 2 secure bits
```
davegarrett: This one is a bit different.

"PC compatible's 18.2 Hz timer" is wrong, though it might be fixable if a hyphen were stuck in the middle of "PC compatible".

"PC's compatible 18.2 Hz timer" is wrong, as it changes the meaning. "(IBM) PC compatible" is a specific term.
https://en.wikipedia.org/wiki/IBM_PC_compatible

I think the correct solution is to simply drop the "'s":
"values taken from a PC compatible 18.2 Hz timer provide 1 or 2 secure bits"

The word "timer" is the noun here, and all preceding are adjectives. The timer is "18.2 Hz" and is "PC compatible". We don't need the possessive form here.

draft-ietf-tls-tls13.md
beurdouche: Fixed, thx

beurdouche:
I'll let native speakers fix the typos next time xD

davegarrett:
Your intuition on how to treat that last one was not wrong, but you need to know the specifics of the terminology to handle it properly. If that weren't a specific term, you'd have been correct, though usage of the word "compatible" there would've been strange.

As to hyphenation, tech terminology turns hyphenated words into compound words way faster than natural language evolution. In other words, we tend to make up new words frequently and write them shorter each time. ;)

richsalz:
In my experience, native speakers are usually far from the best :)

davegarrett:
Native speakers can muddle through free-form usage easily, so they don't notice how bad things might be to others. Non-native speakers can't intuitively piece together words as easily, and thus follow the rules more, which can be problematic because no verbal language actually follows its own rules fully, especially ones with lots of borrowed words. We all screw up language in different and interesting ways. :p

I think a rebase is needed, as the whitespace changes were merged via your other PR.
A squash of these commits into one to drop the reverted bits is probably a good idea, while you're at it.

beurdouche:
Done, I'll look back at it tonight but seems good ... =)

PR #167, opened 2015-04-28
drop obsolete recommendation of 1024-bit DHE as secure

davegarrett:
A newer recommendation could be put in as a replacement, however I don't think the sentence adds anything to the explanation, nor do I expect whatever could be recommended to hold up forever.

Also fix some punctuation in the same paragraph.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3336,10 +3336,10 @@ Users should be able to view information about the certificate and root CA.
 TLS supports a range of key sizes and security levels, including some that
 provide no or minimal security. A proper implementation will probably not
 support many cipher suites. For instance, anonymous Diffie-Hellman is strongly
-discouraged because it cannot prevent man- in-the-middle attacks. Applications
+discouraged because it cannot prevent man-in-the-middle attacks. Applications
 should also enforce minimum and maximum key sizes. For example, certificate
-chains containing 512- bit RSA keys or signatures are not appropriate for
-high-security applications.
+chains containing keys or signatures weaker than 2048-bit RSA are not appropriate
+for secure applications.
 
 ## Implementation Pitfalls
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1437,15 +1437,12 @@ Note that higher layers should not be overly reliant on whether TLS always
...
+been designed to minimize this risk, but there are still attacks available. For
+example, an attacker could block access to the port a secure service runs on
 or attempt to get the peers to negotiate an unauthenticated connection. The
 fundamental rule is that higher levels must be cognizant of what their security
 requirements are and never transmit information over a channel less secure than
-what they require. The TLS protocol is secure in that any cipher suite offers
-its promised level of security: if you negotiate AES-GCM {{GCM}} with
-a 1024-bit DHE key exchange with a host whose certificate you have
-verified, you can expect to be that secure.
+what they require.
```
ekr: I think it's good to keep an example here. Suggest you supply one with a different key size.

davegarrett:
Ok, I've added a new security recommendation. I'm listing a TODO with it, though, as Curve25519 is probably what we're going with as MTI for a reasonable expectation of security and there's nothing to cite yet. I don't think we'd want P256 cited in there for the baseline recommendation. This is fine until the MTIs are updated, I think. People would probably say 2048 DHE is fine, which for now it is, but I went with 3072 to match the same security level for the curve. Might need to discuss on list.

ekr:
I actually expect P256 to be the MTI based on the discussion in person and on-list.

Given that RSA 2048 seems to be what browsers are pushing, I suspect that that's
what's needed here.

davegarrett:
I essentially have no problem with that currently, but this spec will be around for quite a while. I'd rather have the new general purpose expectations in the new spec so that this will age better. (you could argue I'm overthinking this, though) I could also just drop the DHE and list only ECDHE.

ekr:
I'm fine with changing this to 256-bit EC keys, but note that you'll want to refer to that for signatures as well.

davegarrett:
Ok. I've switched it to just 255 bit ECDHE and added 224 bit ECDSA as the minimum signature. (arguably the latter could be higher, but we went with 112 bit symmetric in the backwards compatibility restrictions) I'm specifically going with 255 instead of 256 so Curve25519 isn't ruled out by a bit.

PR #168, opened 2015-04-29
drop obsolete references

davegarrett:
They're not used anywhere in the document anymore.


davegarrett:
In addition to the DES & 3DES references dropped in the first commit, I checked the others and found more obsolete references to drop. The usage of these was cut out in various prior changes. Requiring ephemeral key exchanges made the 4 RFC informative references dropable as they're no longer needed in the "Diffie-Hellman Key Exchange with Authentication" section. "ENCAUTH" is obsolete and no longer cited because AEAD is now mandatory. "CCM" just got cut out, and if we need a citation for it again it will probably be RFC 6655 instead.

PR #169, opened 2015-05-08
relax certificate_list ordering requirements to match current practice

davegarrett:
Current client implementations allow more flexible certificate ordering than the spec allows and some server implementations send certificates in orders that the spec says "MUST" not be done. Relaxing the language here to a "SHOULD" is appropriate so implementers following this spec know that such situations need to be handled. Enforcing this more strictly would cause interop problems for no practical gain.

The modified language here is based on the mailing list discussion.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2665,10 +2665,13 @@ Note: Prior to TLS 1.3, "certificate_list" ordering was required to be strict,
 however some implementations already allowed for some flexibility. For maximum
 compatibility, all implementations SHOULD be prepared to handle potentially
 extraneous certificates and arbitrary orderings from any TLS version (with
-the exception of the sender's certificate, which is expected to be first). Some
-servers are configured to send both a current and deprecated intermediate for
-transitional purposes, and others were simply configured incorrectly, but these
-cases can nonetheless can be validated properly by clients capable of doing so.
+the exception of the sender's certificate). Some servers are configured to send
+both a current and deprecated intermediate for transitional purposes, and others
+are simply configured incorrectly, but these cases can nonetheless be validated
+properly by clients capable of doing so. Whilst the chain MAY be ordered in a
+variety of ways, the peer's end-entity certificate MUST be the first element in
+the vector, and implementations MUST abort with a fatal "bad_certificate" alert
+in the event that this is not the case.
 
 The same message type and structure will be used for the client's response to a
 certificate request message. Note that a client MAY send no certificates if it
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2649,15 +2649,24 @@ Structure of this message:
        } Certificate;
 
 certificate_list
-: This is a sequence (chain) of certificates.  The server's
-  certificate MUST come first in the list.  Each following
-  certificate MUST directly certify the one preceding it.  Because
-  certificate validation requires that root keys be distributed
-  independently, the self-signed certificate that specifies the root
-  certification authority (CA) MAY be omitted from the chain, under the
-  assumption that the client must already possess it in order to
-  validate it in any case.
+: This is a sequence (chain) of certificates. The sender's
+  certificate MUST come first in the list. Each following
+  certificate SHOULD directly certify one preceding it. Because
+  certificate validation requires that trust anchors be distributed
+  independently, a self-signed certificate that specifies a
+  trust anchor MAY be omitted from the chain, provided that
+  supported peers are known to possess any omitted certificates
+  they may require.
 {:br }
+
+Note: Prior to TLS 1.3, "certificate_list" ordering was required to be strict,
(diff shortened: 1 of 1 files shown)


grubba:
This seems like a really bad idea, since it will reduce interoperability even further.

davegarrett:
If you'd like to explain how formally documenting existing behavior (that we are not otherwise proposing be changed) will somehow reduce interoperability further, then please do so on the mailing list.

This proposal is only to properly document existing capabilities such that all readers of the spec actually know what to expect here. *Not* doing this requires either intentionally having a part of the spec we know to be wrong in practice, or changing this to require more strict error enforcement that would break interop.

davegarrett:
Ryan Sleevi posted to the list an in-depth argument as to why the prior requirement can be considered harmful. I'll cite it here for reference, as the threads on this topic got long and winding:
https://www.ietf.org/mail-archive/web/tls/current/msg16238.html

davegarrett:
Rebased.

@seanturner: Last discussion of this on-list ended with lots of agreement and one strong dissenter. Are you able to declare a rough consensus to merge this, or is more discussion needed? Note that this is already in "running code", as this merely documents existing practice in some implementations.

ekr:
@seanturner can this be merged?

draft-ietf-tls-tls13.md
```diff
@@ -2649,16 +2651,25 @@ Structure of this message:
...
+  trust anchor MAY be omitted from the chain, provided that
+  supported peers are known to possess any omitted certificates
+  they may require.
 {:br }
 
+Note: Prior to TLS 1.3, "certificate_list" ordering was required to be strict,
+however some implementations already allowed for some flexibility. For maximum
+compatibility, all implementations SHOULD be prepared to handle potentially
+extraneous certificates and arbitrary orderings from any TLS version (with
+the exception of the sender's certificate, which is expected to be first). Some
```
ekr: s/expected/MUST/

davegarrett: My rationale for using the word "expected" here is that there is already a "MUST" above for this and from the client's perspective, whilst the spec says it MUST be first, it cannot safely assume so and should at least be prepared to abort in the event that it is not.

What might be a better route is to drop the "expected" part in the parenthetical and add a requirement that a client "MUST" abort if it's not first. Probably always better to be more explicit, even if a little more wordy. I'll push another commit to this PR with a change in a bit.

draft-ietf-tls-tls13.md
```diff
@@ -2649,16 +2651,25 @@ Structure of this message:
...
+servers are configured to send both a current and deprecated intermediate for
+transitional purposes, and others were simply configured incorrectly, but these
+cases can nonetheless can be validated properly by clients capable of doing so.
```
davegarrett: Oh look, a typo. I'll fix that too.

davegarrett:
@ekr: Ok, revised and rebased. Please take a look.

davegarrett:
Rebased.

FTR in this PR, question from @jsalowey answered on-list:
https://github.com/davegarrett/tls13-spec/commit/6fbdc44877dcd4c73f5381821474e8db4c913ace#commitcomment-12907817
https://www.ietf.org/mail-archive/web/tls/current/msg17448.html
(GitHub's logic is that if you comment on a commit directly, as apposed to the changeset in the PR, then rebasing ditches your comment from the PR; old comments on the PR get collapsed above, but are still here)

@ekr: I don't know the WG policy here. Is @jsalowey approval sufficient sign-off for you to merge, or do we need @seanturner? Quoting from the list:

On Wed, Aug 26, 2015 at 02:11:01PM -0700, Joseph Salowey wrote:
> It looks like we have good consensus on PR 169 to relax certificate list
> ordering requirements.  I had one question on the revised text.
[question asked and satisfactorily answered by Viktor Dukhovni; link above]
On Wed, Aug 26, 2015 at 10:25:07PM -0700, Joseph Salowey wrote:
> Thanks Viktor,  I missed this part of the discussion.  The text looks fine to me as is.

So, is this finally ready to land now?

draft-ietf-tls-tls13.md
```diff
@@ -2649,16 +2651,28 @@ Structure of this message:
...
+however some implementations already allowed for some flexibility. For maximum
+compatibility, all implementations SHOULD be prepared to handle potentially
+extraneous certificates and arbitrary orderings from any TLS version (with
+the exception of the sender's certificate). Some servers are configured to send
+both a current and deprecated intermediate for transitional purposes, and others
+are simply configured incorrectly, but these cases can nonetheless be validated
+properly by clients capable of doing so. Whilst the chain MAY be ordered in a
+variety of ways, the peer's end-entity certificate MUST be the first element in
+the vector, and implementations MUST abort with a fatal "bad_certificate" alert
+in the event that this is not the case.
```
ekr: It's not clear that this is actually actionable just because of how implementations behave.

PR #170, opened 2015-05-16
When will this message be sent ?

beurdouche:
Shouldn't that be a question ?


ekr:
I think I actually prefer the current version....

davegarrett:
I agree with ekr. The current heading is better. It doesn't need to be a question.

beurdouche:
ok closing

PR #172, opened 2015-05-16
Remove glossary

ekr:


ekr:
Added because git became sad somehow

PR #175, opened 2015-05-17
remove errant "--- back" line

davegarrett:
There's a line at the end of the "IANA Considerations" section that just has the content "--- back". It's seemingly not valid markup for anything. It looks like it was introduced in the initial conversion to markdown, but has been left in there ever since. I'm pretty sure it's some accidental cruft added from the conversion, but I'm not 100% sure. (copy/paste from an editor?)

It doesn't seem to have a purpose, though of course, please correct me if I'm wrong. Here's a PR to just delete the line.


ekr:
I believe that this makes the remainder of the document into appendices.

davegarrett:
This stuff needs better documentation...

After searching around for a bit, yeah, you appear to be correct. This looks like it's for kramdown-rfc2629. It seems this variation of the syntax separates things into sections named abstract, middle, and back, labeling them on HR lines with "---".

Ok, nvm. Sorry for the noise. Closing.

Issue #53, opened 2014-07-07
Add non-GCM AEAD ciphers


ekr:
I only added the GCM AEAD ciphers. We should add all the standards track AEAD ciphers.

davegarrett:
@ekr: This was dealt with in PR #180 and can be closed.

PR #180, opened 2015-05-25
revise cipher suites section (fixes issue #53)

davegarrett:
This revises "The Cipher Suites" section to simplify redundant language, actually explain the de facto naming convention, and add all currently available AEAD cipher suites. It also drops the appendix table that has already had its section whittled down quite a bit, in favor of the explanation and new table in the main section.

Note that this is all of the AEAD ciphers available in the current IANA registry [0], plus ChaChaPoly (which has to be a normative reference [1], so must reach RFC status before this does). Issue #53 just said "standards track AEAD ciphers", but as the registry lists them all, adding them wasn't difficult. If we'd rather not cite them here then they can of course be cut out (ARIA, Camellia, & ECDHE AES-CCM are informational; the rest are standards track, including DHE AES-CCM). Alternatively, the table could be annotated to note which are on what track or split in two.

This has no changes to any requirements.

[0] https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4
[1] https://www.ietf.org/iesg/statement/normative-informative.html

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -38,6 +38,11 @@ normative:
   RFC5288:
   RFC5289:
   RFC5869:
+  RFC6209:
+  RFC6367:
+  RFC6655:
+  RFC7251:
+  I-D.ietf-tls-chacha20-poly1305:
   AES:
        title: Specification for the Advanced Encryption Standard (AES)
        date: 2001-11-26
@@ -311,6 +316,8 @@ draft-08
 - Remove support for weak and lesser used named curves.
 
 - Remove support for MD5 and SHA-224 hashes with signatures.
+
+- Revise list of currently available AEAD cipher suites.
 
 
 draft-07
@@ -1862,7 +1869,7 @@ session_id
 cipher_suites
 : This is a list of the cryptographic options supported by the
   client, with the client's first preference first.
-  Values are defined in {{the-cipher-suite}}.
+  Values are defined in {{cipher-suites}}.
 
 compression_methods
 : Versions of TLS before 1.3 supported compression and the list of
@@ -3309,7 +3316,7 @@ other than for computing other secrets.)
 #  Mandatory Cipher Suites
 
 In the absence of an application profile standard specifying otherwise, a
-TLS-compliant application MUST implement the cipher suite [TODO:Needs to be selected](https://github.com/tlswg/tls13-spec/issues/32). (See {{the-cipher-suite}} for the definition.)
+TLS-compliant application MUST implement the cipher suite [TODO:Needs to be selected](https://github.com/tlswg/tls13-spec/issues/32). (See {{cipher-suites}} for the definition.)
 
 
 #  Application Data Protocol
@@ -3321,8 +3328,8 @@ are treated as transparent data to the record layer.
 
 #  Security Considerations
 
-Security issues are discussed throughout this memo, especially in Appendices C,
-D, and E.
+Security issues are discussed throughout this memo, especially in Appendices B,
+C, and D.
 
 
 #  IANA Considerations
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3056,9 +3063,29 @@ This section describes protocol types and constants.
...
 ## The Cipher Suite
 
-The following values define the cipher suite codes used in the ClientHello and
-ServerHello messages.
-A cipher suite defines a cipher specification supported in TLS.
+A cipher suite defines a cipher specification supported in TLS and negotiated
+via hello messages in the TLS handshake.
+Cipher suites follow a general naming convention:
+
+       CipherSuite TLS_KEA_SIGN_WITH_CIPHER_HASH = VALUE;
```
richsalz: perhaps use uppercase for TLS WITH and lowercase for kea sign cipher hash?

draft-ietf-tls-tls13.md
```diff
@@ -3066,52 +3093,69 @@ negotiated, as it provides no more protection than an unsecured connection.
...
+implementations unless the application layer has specifically requested to allow
+anonymous key exchange. Anonymous key exchange may sometimes be acceptable, for
+example, to support opportunistic encryption when no set-up for authentication is
+in place, or when TLS is used as part of more complex security protocols that have
+other means to ensure authentication. The following specifications provide "DH_anon"
+key exchange cipher suites:
+AES-GCM [RFC5288], ARIA-GCM [RFC6209], and CAMELLIA-GCM [RFC6367].
+
+All cipher suites in this section are specified for usage with TLS 1.3 as well
+as with TLS 1.2 in implementations acting in a backwards compatible mode.
```
richsalz: TLS 1.2 in backward compat mode?  This sentence is confusing.

draft-ietf-tls-tls13.md
davegarrett: I could explicitly mention that the "TLS" & "WITH" parts are just static parts of the naming. That could make it more clear, as well.

davegarrett:
@richsalz I addressed your comments in this commit. I switched to a table to make things easier to read and explicitly clarified the constant string bits so as to be more clear. I also dropped the confusing sentence fragment.

richsalz:
looks good.

davegarrett:
AES-OCB will also be addable to this list, but its codepoints are all completely TBD still.
https://tools.ietf.org/html/draft-zauner-tls-aes-ocb-02

davegarrett:
Rebased, squashed, & updated for recent changes to draft.

draft-ietf-tls-tls13.md
```diff
@@ -3413,52 +3450,73 @@ negotiated, as it provides no more protection than an unsecured connection.
...
+    TLS_ECDHE_ECDSA_WITH_CAMELLIA_128_GCM_SHA256  {0xC0,0x86}    [RFC6367]
+    TLS_ECDHE_ECDSA_WITH_CAMELLIA_256_GCM_SHA384  {0xC0,0x87}    [RFC6367]
+    TLS_ECDHE_RSA_WITH_CAMELLIA_128_GCM_SHA256    {0xC0,0x8A}    [RFC6367]
+    TLS_ECDHE_RSA_WITH_CAMELLIA_256_GCM_SHA384    {0xC0,0x8B}    [RFC6367]
+    TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305          {0xCC,0xA1}  [I-D.ietf-tls-chacha20-poly1305]
+    TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305        {0xCC,0xA2}  [I-D.ietf-tls-chacha20-poly1305]
+    TLS_DHE_RSA_WITH_CHACHA20_POLY1305            {0xCC,0xA3}  [I-D.ietf-tls-chacha20-poly1305]
+
+[[TODO: CHACHA20_POLY1305 cipher suite IDs are TBD. Must update above if
+final versions change.]]
```
ekr: My understanding is that they were definitely changing the numbers. If that's so, let's just put TBD here.

draft-ietf-tls-tls13.md
```diff
@@ -3413,52 +3450,73 @@ negotiated, as it provides no more protection than an unsecured connection.
...
+
+Note: In the case of the CCM mode of AES, two variations exist: "CCM_8" which
+uses an 8-bit authentication tag and "CCM" which uses a 16-bit authentication
+tag. Both use the default hash, SHA-256.
+
+DHE AES GCM & CCM are standards track ([RFC5288] & [RFC6655]),
+as is ChaChaPoly ([I-D.ietf-tls-chacha20-poly1305]).  The ECDHE variants of AES
+and other ciphers above are currently informational, however ECDHE AES GCM
+is widely deployed.
```
ekr: Let's put this in a table instead.

draft-ietf-tls-tls13.md
```diff
@@ -3413,52 +3450,73 @@ negotiated, as it provides no more protection than an unsecured connection.
...
+anonymous key exchange. Anonymous key exchange may sometimes be acceptable, for
+example, to support opportunistic encryption when no set-up for authentication is
+in place, or when TLS is used as part of more complex security protocols that have
+other means to ensure authentication. The following specifications provide "DH_anon"
+key exchange cipher suites:
+AES-GCM [RFC5288], ARIA-GCM [RFC6209], and CAMELLIA-GCM [RFC6367].
+
+All cipher suites in this section are specified for usage with both TLS 1.2
+and TLS 1.3. (see {{backward-compatibility}})
+
```
ekr: And DTLS.

draft-ietf-tls-tls13.md
davegarrett: On Wednesday, July 15, 2015 02:22:12 pm ekr wrote:
> Let's put this in a table instead.

Which would you prefer:
a) add an asterisk next to the spec links to indicate informational
b) split the table into two for standard and informational

I think (a) is better, because that keeps ECDHE AES GCM next to DHE AES GCM. Either works.

Is it possible to promote cipher suite codepoints to standard status straight in this RFC?


Dave

ekr: Let's split the table.

I think we can in fact promote them, but 4492bis will be making them standards track anyway.

davegarrett:
Ok, all comments addressed with that last commit.

I've split the table into standards and non-standards and simplified the note to just mention ECDHE AES GCM being widely deployed.

davegarrett:
Rebased.

davegarrett:
Ok, reworded that bit too.

Issue #186, opened 2015-06-17
Prune EC curves and remove md5


ekr:

davegarrett:
Here's a PR for this: #188 

Question: Are we dropping all curves <224 bits or <255 bits?
(was consensus ever called on this specific number?)

seanturner:
There was not a specific WG consensus call on this point.  I hope we don't need one and that we should keep the curves in that are used and trim the ones that aren't.

davegarrett:
Good. In that case, I'll slash everything other than secp256r1 (prime256v1 / NIST P-256), secp384r1 (NIST P-384), secp521r1 (NIST P-521), and sect571r1 (NIST B-571). (plus, of course, new CFRG curves yet to be added) PR updated.

https://tools.ietf.org/html/rfc4492#appendix-A
https://securitypitfalls.wordpress.com/2015/05/23/april-2015-scan-results/

davegarrett:
Based on the fact that apparently precisely zero servers actually use SHA-224 for signatures, can we drop that too while we're at it? (not an exaggeration; the scan cited in the last comment shows none)

seanturner:
I would drop it.

spt

davegarrett:
OK. PR updated.

davegarrett:
@ekr: This can be closed now that PR #188 has been merged.

PR #188, opened 2015-06-17
Deprecate old curves, MD5, & SHA-224 (for issue #186)

davegarrett:
Not sure if the deprecation cutoff will be at 224 bits (~112 bit security level) or 255 bits (~128 bit security level) yet, so I've just done the former for now.

This drops the old curves under 224 bits, updates the extension example bytes, and deprecates MD5.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -33,7 +33,6 @@ normative:
   RFC2104:
   RFC2119:
   RFC2434:
-  RFC1321:
   RFC3447:
   RFC5280:
   RFC5288:
@@ -92,6 +91,7 @@ normative:
 
 informative:
   RFC0793:
+  RFC1321:
   RFC1948:
   RFC2246:
   RFC3268:
@@ -309,7 +309,15 @@ server: The endpoint which did not initiate the TLS connection.
 ##  Major Differences from TLS 1.2
 
 
+draft-08
+
+- Remove support for weak and lesser used named curves.
+
+- Remove support for MD5 and SHA-224 hashes with signatures.
+
+
 draft-07
+
 - Integration of semi-ephemeral DH proposal.
 
 - Add initial 0-RTT support
@@ -331,17 +339,12 @@ draft-06
 
 - Remove explicit IV.
 
-- Remove support for weak and lesser used named curves.
-
-- Remove support for MD5 and SHA-224 hashes with signatures.
-
 
 draft-05
 
 - Prohibit SSL negotiation for backwards compatibility.
 
 - Fix which MS is used for exporters.
-
 
 
 draft-04
@@ -357,6 +360,7 @@ draft-04
 - Remove renegotiation.
 
 - Remove point format negotiation.
+
 
 draft-03
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -331,6 +331,10 @@ draft-06
 
 - Remove explicit IV.
 
+- Remove support for weak and lesser used named curves.
+
+- Remove support for MD5 and SHA-224 hashes with signatures.
+
 
 draft-05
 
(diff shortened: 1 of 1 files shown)


davegarrett:
Based on ongoing list discussion, I've added SHA1 to the chopping block here.

davegarrett:
As noted on-list, SHA1 can't be as easily dealt with here, so that'll be a topic for a separate changeset.

draft-ietf-tls-tls13.md
```diff
@@ -2116,6 +2128,9 @@ hash
   SHA-224, SHA-256, SHA-384, and SHA-512 {{SHS}}, respectively.  The
   "none" value is provided for future extensibility, in case of a
   signature algorithm which does not require hashing before signing.
+  The usage of MD5 and SHA-224 are deprecated. The md5_RESERVED and
+  sha224_RESERVED values MUST NOT be offered or negotiated by any
+  implementation.
```
ekr: I think this should be stronger. TLS 1.3 agents MUST reject if any TLS 1.3-advertising agent offers them. wdyt?

draft-ietf-tls-tls13.md
```diff
@@ -2173,15 +2188,10 @@ The "extension_data" field of this extension SHALL contain a
...
-           sect193r1 (4), sect193r2 (5), sect233k1 (6),
-           sect233r1 (7), sect239k1 (8), sect283k1 (9),
-           sect283r1 (10), sect409k1 (11), sect409r1 (12),
-           sect571k1 (13), sect571r1 (14), secp160k1 (15),
-           secp160r1 (16), secp160r2 (17), secp192k1 (18),
-           secp192r1 (19), secp224k1 (20), secp224r1 (21),
-           secp256k1 (22), secp256r1 (23), secp384r1 (24),
-           secp521r1 (25),
+           obsolete_RESERVED (1..13),
+           sect571r1 (14),
```
ekr: This (1..13) probably isn't legal syntax. I would suggest that we just add a comment here instad

draft-ietf-tls-tls13.md
davegarrett: I have a line doing that in my WIP branch at the moment. We can get this bit in for now, though.

draft-ietf-tls-tls13.md
davegarrett: We should probably add documentation of this syntax up top.

ekr: I would get rid og sect571r1 too

draft-ietf-tls-tls13.md
```diff
@@ -2199,31 +2208,31 @@ The "extension_data" field of this extension SHALL contain a
...
 
-sect163k1, etc
-: Indicates support of the corresponding named curve
-  The named curves defined here are those specified in SEC 2 [13].
-  Note that many of these curves are also recommended in ANSI
-  X9.62 {{X962}} and FIPS 186-2 {{DSS}}.  Values 0xFE00 through 0xFEFF are
-  reserved for private use.  Values 0xFF01 and 0xFF02 were used in
-  previous versions of TLS but MUST NOT be offered by TLS 1.3
-  implementations.
-  [[OPEN ISSUE: Triage curve list.]]
```
ekr: I think this comment about MUST NOT be offfered should stick around.

draft-ietf-tls-tls13.md
davegarrett: I'm not strongly against that, but I think it might warrant discussion on the list. It would effectively reduce the maximum bits allowed here. That of course does not tell the whole story, but some people might argue against it, and it does get used a little bit (see stats link in main comments).

ekr: Do you want to raise it on-list?

richsalz: The syntactic sugar is nice, but it means that obsolete_RESERVED is ambiguous, allowing any value in the range 11-13 which sort of defeats the purpose of an enumeration unless you really want to special-case words ending in RESERVED.

davegarrett: The "RESERVED" naming convention has been used rather frequently throughout this document. We could point to an older spec for them, if you'd like.

richsalz: I don’t have a problem with xxx_RESERVED as a convention (and all uses also seem to say MUST NOT use in prose), it’s the range “(1..13)” that I dislike.

ekr: Me too, but I noted that it's already in the document so I think we should try to fix it later.

davegarrett: @ekr: It looks like you introduced that notation here:
https://github.com/tlswg/tls13-spec/commit/990192023ea39865029ae37d605dd1a46729e853#diff-9d84740dcc569a0a5a359d0fba461a05R2285

I think it's a notable improvement over having to list them all individually. We just need to formally define the syntax up top.

ekr: Actually I pulled it in from RFC 4492

richsalz: Okay, I updated my parser (https://github.com/richsalz/tlsparser) to handle this.

Issue #185, opened 2015-06-17
Generate a new ClientRandom for HelloRetryRequest


ekr:
Suggested by Karthik

davegarrett:
I created a short PR for this: PR #189

davegarrett:
I amended the PR to make sure to be clear that the new client random is used and the old is discarded, and that it still ends up in the session hash which still starts from the beginning of the connection with the initial ClientHello. (see also issue #104, which this might be sufficient to resolve)

davegarrett:
@ekr: Please close this too. (this is the issue for PR 189)

Issue #104, opened 2014-12-22
Re-start session hash and Finished hash in mismatched case


ekr:
See Figure 2,

"[[OPEN ISSUE: Do we restart the handshake hash?]]
[[OPEN ISSUE: We need to make sure that this flow doesn't introduce
downgrade issues. Potential options include continuing the handshake
hashes (as long as clients don't change their opinion of the server's
capabilities with aborted handshakes) and requiring the client to send
the same ClientHello (as is currently done) and then checking you get
the same negotiated parameters.]]"

Re-starting the hashes is conceptually cleaner, but needs security
analysis.

enygren:
For the repeated ClientHello case (in the face of HelloRetryRequest), what about having HelloRetryRequest always contain a cookie (similar to in HelloVerifyRequest) and require repeated ClientHellos to replay that cookie.  If that cookie is an HMAC by the server across an key held by the server, any server extra data (client ip?), and the invariant part of the ClientHello, then the server can revalidate the ClientHello without needing to keep any state about it.  The handshake hash would contain the cookie (which includes an HMAC of the initial ClientHello that the server should revalidate against the new ClientHello).  This aligns well with DTLS HelloVerifyRequest and also allows the first ClientHello to be excluded from the handshake hash, and then makes things like DTLS and any future client-puzzles extension work fit in fairly cleanly.

ekr:
I certainly think we want a cookie here because I want to merge HelloVerifyRequest and HelloRetryRequest. I'm less sure that the "verify that it was sent without changes" plan is practical, but it's certainly something we can explore.

PR #189, opened 2015-06-20
generate new ClientHello.random for HelloRetryRequest (for issue #185)

davegarrett:
PR to add a simple requirement that retried ClientHello messages must not reuse random values and servers must check for and reject them if they do. (suggested by Karthik)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1886,10 +1886,9 @@ ClientHello as its first message. The client will also send a
 ClientHello when the server has responded to its ClientHello with a
 ServerHello that selects cryptographic parameters that don't match the
 client's KeyShare extension. In that case, the client MUST send the same
-ClientHello (without modification) except including a new KeyShareEntry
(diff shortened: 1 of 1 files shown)


ekr:
I think it's an open issue whether we should require a new ClientRandom, but requiring that the server check for it is harmful because it makes it impossible to do a stateless HelloRetryRequest.

davegarrett:
Ok, then should I downgrade it to a SHOULD (or MAY) or drop the check requirement completely? I don't have a strong objection to dropping it.

davegarrett:
Also, would that also require the session hash to completely restart on the new ClientHello and discard the previous messages? (see also issue #104) I added in clarification for that before noticing your comment, so that may need changing as well.

martinthomson:
Weeeelll, if we include a cookie, it is possible to check statelessly.  But the cookie might get a little big :)

davegarrett:
I've revised this to drop the required check. (also now based on current master, as it was rather bitrotted)

ekr:
I don't think we want this.

PR #196, opened 2015-07-08
Default to sha256 for Supported Signatures Extension

tomato42:
Since sha1 is not considered secure at this time and sha224 is
otherwise unused in this document, default to the next secure
widely supported hash algorithm.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2177,19 +2177,24 @@ signature_algorithms extension, listing the algorithms it is willing to accept.
 If the client does not send the signature_algorithms extension, the server MUST
 do the following:
 
-- If the negotiated key exchange algorithm is one of (DHE_RSA, ECDHE_RSA), behave as if client had sent the value
-{sha1,rsa}.
+- If the negotiated key exchange algorithm is one of (DHE_RSA, ECDHE_RSA),
+behave as if client had sent the value {sha256,rsa}.
 
 - If the negotiated key exchange algorithm is DHE_DSS, behave
-as if the client had sent the value {sha1,dsa}.
+as if the client had sent the value {sha256,dsa}.
 
 - If the negotiated key exchange algorithm is ECDHE_ECDSA,
-behave as if the client had sent value {sha1,ecdsa}.
+behave as if the client had sent value {sha256,ecdsa}.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2177,20 +2177,26 @@ signature_algorithms extension, listing the algorithms it is willing to accept.
...
 Note: This extension is not meaningful for TLS versions prior to 1.2. Clients
 MUST NOT offer it if they are offering prior versions. However, even if clients
 do offer it, the rules specified in {{RFC6066}} require servers to ignore
 extensions they do not understand.
 
+Note: The default values for this extension were changed in TLS 1.3.
+Previously the server was expected to behave as if the client had sent a value
+with sha1 hash algorithm. In case the client is willing to negotiate lower
+protocol version than described in this document it MUST include this
+extension.
```
davegarrett: You can replace this last sentence with just:
"Clients willing to negotiate TLS 1.2 MUST include this extension."

The note above says there's no meaning for this extension prior to TLS 1.2, so 1.2 is the only "lower protocol version" here. Simpler to just say that.

tomato42: fixed

draft-ietf-tls-tls13.md
```diff
@@ -2177,20 +2177,25 @@ signature_algorithms extension, listing the algorithms it is willing to accept.
...
 Note: This extension is not meaningful for TLS versions prior to 1.2. Clients
 MUST NOT offer it if they are offering prior versions. However, even if clients
 do offer it, the rules specified in {{RFC6066}} require servers to ignore
 extensions they do not understand.
 
+Note: The default values for this extension were changed in TLS 1.3.
+Previously the server was expected to behave as if the client had sent a value
+with sha1 hash algorithm. In case the client is willing to negotiate TLS 1.2
+it MUST include this extension.
+
```
ekr: I like the sentiment, but I think this is going in the wrong direction. I suggest we simply require sending this all the time, and then we won't need to worry about defaults.

davegarrett:
PR #201 removes these defaults and mandates use of the "signature_algorithms" extension for all certificate authenticated cipher suites.

ekr:
This is now OBE.

PR #200, opened 2015-07-15
drop support for sect571r1

davegarrett:
It appears to be the consensus that sect571r1 should be dropped.

This drops it and adds placeholders for the CFRG curves. There are likely to be more than these two values in the final version to indicate ECDHE & certificate usage separately.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2199,9 +2199,7 @@ The "extension_data" field of this extension SHALL contain a
 %%% Named Group Extension
        enum {
            // Elliptic Curve Groups.
-           obsolete_RESERVED (1..13),
-           sect571r1 (14),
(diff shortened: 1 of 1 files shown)


ekr:
@seanturner: shall we remove sect571r1

Let's let the CFRG draft get a little further along before we add the placeholders.

davegarrett:
On Wednesday, July 15, 2015 06:23:48 pm ekr wrote:
> Let's let the CFRG draft get a little further along before we add the placeholders.

Ok. I dropped that bit and pushed a new commit here.

davegarrett:
Rob Stradling just posted to the list that he suspects that most users of this curve are just due to implementations blindly picking the biggest number, not because it was actually chosen. OpenSSL seems to do this by default and doesn't really provide an obvious way to change that.

Issue #116, opened 2014-12-30
Require signature_algorithms extension.


ekr:
We should just require it.

beurdouche:
I agree with that, it goes with potentially deleting the SignatureAlgorithm from the new ciphersuites.
By the way, shouldn't we consider also removing MD5 from HashAlgorithms even if it is used a lot 
(I don't know the numbers on that ...) ?

davegarrett:
PR #201 mandates usage of this extension and also adds an explicit error expectation for MD5 (it's already listed as deprecated in the current draft).

davegarrett:
@ekr: This should be closable now.

PR #201, opened 2015-07-16
MTI Extensions (incl. issue #116)

davegarrett:
This is a set of changes covering mailing list discussions about error alerts, certificate acceptance, SHA-1 deprecation, as well as including some recommendations about curves. There's also quite a few other little fixes mixed in there, including some cleanups in RFC2119 language. The alerts list has been fully updated to include all standardized alerts (PSK alert is needed for new PSK extension, in particular). This also adds RFC 5746 ("renegotiation_info" info extension) to the list of newly obsoleted RFCs, as renegotiation has already been removed completely thus making it no longer applicable as of this new TLS version.

PR #169 has the discussed change to certificate ordering requirements, however it is waiting on @seanturner to assess consensus so I'm not including it here.

(Edited to note: final commits from this PR have changed and parts have been broken off into other branches for other PRs)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -116,6 +116,7 @@ informative:
   RFC6066:
   RFC6176:
   RFC7465:
+  RFC7507:
   RFC7568:
   I-D.ietf-tls-negotiated-ff-dhe:
   I-D.ietf-tls-session-hash:
@@ -318,7 +319,7 @@ draft-08
 
 - Remove support for MD5 and SHA-224 hashes with signatures.
 
-- Revise list of currently available AEAD cipher suites.
+- Update lists of available AEAD cipher suites and error alerts.
  
 - Reduce maximum permitted record expansion for AEAD from 2048 to 256 octets.
 
@@ -1152,31 +1153,37 @@ as specified by the current connection state.
 
        enum {
            close_notify(0),
-           unexpected_message(10),              /* fatal */
-           bad_record_mac(20),                  /* fatal */
-           decryption_failed_RESERVED(21),      /* fatal */
-           record_overflow(22),                 /* fatal */
-           decompression_failure_RESERVED(30),  /* fatal */
-           handshake_failure(40),               /* fatal */
-           no_certificate_RESERVED(41),         /* fatal */
+           unexpected_message(10),               /* fatal */
+           bad_record_mac(20),                   /* fatal */
+           decryption_failed_RESERVED(21),       /* fatal */
+           record_overflow(22),                  /* fatal */
+           decompression_failure_RESERVED(30),   /* fatal */
+           handshake_failure(40),                /* fatal */
+           no_certificate_RESERVED(41),          /* fatal */
            bad_certificate(42),
            unsupported_certificate(43),
            certificate_revoked(44),
            certificate_expired(45),
            certificate_unknown(46),
-           illegal_parameter(47),               /* fatal */
-           unknown_ca(48),                      /* fatal */
-           access_denied(49),                   /* fatal */
-           decode_error(50),                    /* fatal */
-           decrypt_error(51),                   /* fatal */
-           export_restriction_RESERVED(60),     /* fatal */
-           protocol_version(70),                /* fatal */
-           insufficient_security(71),           /* fatal */
-           internal_error(80),                  /* fatal */
+           illegal_parameter(47),                /* fatal */
+           unknown_ca(48),                       /* fatal */
+           access_denied(49),                    /* fatal */
+           decode_error(50),                     /* fatal */
+           decrypt_error(51),                    /* fatal */
+           export_restriction_RESERVED(60),      /* fatal */
+           protocol_version(70),                 /* fatal */
+           insufficient_security(71),            /* fatal */
+           internal_error(80),                   /* fatal */
+           inappropriate_fallback(86),           /* fatal */
            user_canceled(90),
-           no_renegotiation_RESERVED(100),      /* fatal */
-           missing_extension(109),              /* fatal */
-           unsupported_extension(110),          /* fatal */
+           no_renegotiation_RESERVED(100),       /* fatal */
+           missing_extension(109),               /* fatal */
+           unsupported_extension(110),           /* fatal */
+           certificate_unobtainable(111),
+           unrecognized_name(112),
+           bad_certificate_status_response(113), /* fatal */
+           bad_certificate_hash_value(114),      /* fatal */
+           unknown_psk_identity(115),
            (255)
        } AlertDescription;
 
@@ -1366,6 +1373,10 @@ internal_error
   protocol (such as a memory allocation failure) makes it impossible
   to continue.  This alert is always fatal.
 
+inappropriate_fallback
+: Sent by a server in response to an invalid connection retry attempt
+  from a client. (see [RFC7507]) This alert is always fatal.
+
 no_renegotiation_RESERVED
 : This alert was used in previous versions of TLS. TLS 1.3 does not
   include renegotiation and TLS 1.3 implementations MUST NOT send this
@@ -1381,6 +1392,31 @@ unsupported_extension
 : Sent by clients that receive an extended ServerHello containing
   an extension that they did not put in the corresponding ClientHello.
   This alert is always fatal.
+
+certificate_unobtainable
+: Sent by servers when unable to obtain a certificate from a URL
+  provided by the client via the "client_certificate_url" extension
+  [RFC6066].
+
+unrecognized_name
+: Sent by servers when no server exists identified by the name
+  provided by the client via the "server_name" extension
+  [RFC6066].
+
+bad_certificate_status_response
+: Sent by clients when an invalid or unacceptable OCSP response is
+  provided by the server via the "status_request" extension
+  [RFC6066]. This alert is always fatal.
+
+bad_certificate_hash_value
+: Sent by servers when a retreived object does not have the correct hash
+  provided by the client via the "client_certificate_url" extension
+  [RFC6066]. This alert is always fatal.
+
+unknown_psk_identity
+: Sent by servers when an unknown PSK identity is provided by the client.
+  Sending this alert is OPTIONAL; servers MAY instead choose to send a
+  "decrypt_error" alert to merely indicate an invalid PSK identity.
 {:br }
 
 New Alert values are assigned by IANA as described in {{iana-considerations}}.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -110,6 +110,7 @@ informative:
   RFC5077:
   RFC5081:
   RFC5116:
+  RFC5246:
   RFC5705:
   RFC5763:
   RFC6066:
@@ -1174,6 +1175,7 @@ as specified by the current connection state.
            internal_error(80),                  /* fatal */
            user_canceled(90),
            no_renegotiation_RESERVED(100),      /* fatal */
+           missing_extension(109),              /* fatal */
            unsupported_extension(110),          /* fatal */
            (255)
        } AlertDescription;
@@ -1368,6 +1370,12 @@ no_renegotiation_RESERVED
 : This alert was used in previous versions of TLS. TLS 1.3 does not
   include renegotiation and TLS 1.3 implementations MUST NOT send this
   alert when in TLS 1.3 mode. This alert is always fatal.
+
+missing_extension
+: Sent by endpoints that receive a hello message not containing an
+  extension that is mandatory to send for the offered TLS version.
+  This message is always fatal.
+[[TODO: IANA Considerations.]]
 
 unsupported_extension
 : Sent by clients that receive an extended ServerHello containing
@@ -2146,32 +2154,18 @@ suite indicates permissible signature algorithms but not hash algorithms.
 {{server-certificate}} and {{server-key-share}} describe the
 appropriate rules.
 
-If the client supports only the default hash and signature algorithms (listed
-in this section), it MAY omit the "signature_algorithms" extension. If the client
-does not support the default algorithms, or supports other hash and signature
-algorithms (and it is willing to use them for verifying messages sent by the
-server, i.e., server certificates and server key share), it MUST send the
-"signature_algorithms" extension, listing the algorithms it is willing to accept.
-
-If the client does not send the "signature_algorithms" extension, the server MUST
-do the following:
-
-- If the negotiated key exchange algorithm is one of (DHE_RSA, ECDHE_RSA),
-  behave as if client had sent the value {sha1,rsa}.
-
-- If the negotiated key exchange algorithm is DHE_DSS,
-  behave as if the client had sent the value {sha1,dsa}.
-
-- If the negotiated key exchange algorithm is ECDHE_ECDSA,
-  behave as if the client had sent value {sha1,ecdsa}.
-
-Note: This extension is not meaningful for TLS versions prior to 1.2. Clients
-MUST NOT offer it if they are offering prior versions. However, even if clients
-do offer it, the rules specified in {{RFC6066}} require servers to ignore
-extensions they do not understand.
+All clients MUST send a valid "signature_algorithms" extension in their
+ClientHello when offering certificate authenticated cipher suites.
+Servers receiving a TLS 1.3 ClientHello offering certificate authenticated
+cipher suites without this extension MUST send a "missing_extension" alert
+message and close the connection.
 
 Servers MUST NOT send this extension. TLS servers MUST support receiving this
 extension.
+
+Note: TLS 1.3 servers MAY receive TLS 1.2 ClientHellos which do not contain
+this extension. If those servers are willing to negotiate TLS 1.2, they MUST
+behave in accordance with the requirements of {{RFC5246}}.
 
 #### Negotiated Groups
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2139,9 +2203,9 @@ hash
   SHA-224, SHA-256, SHA-384, and SHA-512 {{SHS}}, respectively.  The
   "none" value is provided for future extensibility, in case of a
   signature algorithm which does not require hashing before signing.
-  The usage of MD5 and SHA-224 are deprecated. The md5_RESERVED and
-  sha224_RESERVED values MUST NOT be offered or negotiated by any
-  implementation.
+  The use of MD5, SHA-1, and SHA-224 are deprecated. The md5_RESERVED,
+  sha1_RESERVED, and sha224_RESERVED values MUST NOT be offered by
+  any implementation.
```
tomato42: what if the implementation is willing to negotiate TLSv1.2 connection?

draft-ietf-tls-tls13.md
```diff
@@ -3311,10 +3413,45 @@ because TLS does not directly use this secret for anything
...
+  * Server Name Indication ("server_name"; Section 3 of {{RFC6066}})
+
+All implementations MUST use the "signature_algorithms" extension when
+offering and negotiating certificate authenticated cipher suites.
+All implementations MUST use the "supported_groups" and "client_key_share"
+extensions when offering and negotiating DHE or ECDHE cipher suites.
+All implementations MUST use the "pre_shared_key" extension when
+offering and negotiating PSK cipher suites.
+All implementations MUST support use of the "server_name" extension
+with applications capable of using it.
```
tomato42: what is the server supposed to do upon receiving a TLS1.3 Client Hello without either of those extensions?

draft-ietf-tls-tls13.md
davegarrett: You're right; there's a potential issue here. TLS 1.3 clients with this provision would offer SHA-256 and TLS 1.2 servers would either send a SHA-256 chain if they have one, send a chain with SHA-1 instead, or possibly abort. Based on recent discussion, yes, we would need to check what various implementations would do here. If they abort, I'd have to change this to allow sending SHA-1 support but not negotiating it for TLS 1.3.

draft-ietf-tls-tls13.md
davegarrett: The "signature_algorithms" and "supported_groups" sections have requirements to fatal alert without them. The "server_name" extension error is specified just below here. "client_key_share" and "pre_shared_key" don't have explicit error expectations other than guaranteed failure without them.

davegarrett:
@tomato42: I pushed a new commit here to address your comments. I've changed SHA1 to a SHOULD NOT except for backwards compatibility and added explicit error cases for if the other two extensions are missing (better explicit than implicit).

draft-ietf-tls-tls13.md
```diff
@@ -2434,10 +2524,13 @@ identifier
...
-such extension is present, the server MUST NOT negotiate
-a PSK cipher suite. If no suitable identity is present, the server
-MUST NOT negotiate a PSK cipher suite.
+PreSharedKeyExtension to indicate the pre-shared key(s) to be used.
+Servers negotiating TLS 1.3 for a ClientHello without this extension yet
+offering any PSK cipher suite MUST send a "missing_extension"
+alert message and close the connection.
+If no suitable identity is present, the server MUST NOT negotiate
+a PSK cipher suite and MAY respond with an "unknown_psk_identity"
+alert message.
```
tomato42: Didn't TLS1.2 allow for masking if the given identity is correct or not? I think many people may want that, even if it simply can't be made timing resistant...

tomato42:
Yes, I was looking for the explicit description to send alerts in that case. I like those changes.

:+1:

draft-ietf-tls-tls13.md
davegarrett: If you mean this:
https://tools.ietf.org/html/rfc4279#section-2
>   If the server does not recognize the PSK identity, it MAY respond
>   with an "unknown_psk_identity" alert message.  Alternatively, if the
>   server wishes to hide the fact that the PSK identity was not known,
>   it MAY continue the protocol as if the PSK identity existed but the
>   key was incorrect: that is, respond with a "decrypt_error" alert.

That's still here. This section is why I have "MAY respond with an "unknown_psk_identity"", rather than a "SHOULD" or "MUST".

davegarrett:
Rebased and squashed a little bit.

draft-ietf-tls-tls13.md
```diff
@@ -2306,9 +2387,11 @@ values, each representing a single set of key agreement parameters;
 for instance a client might offer shares for several elliptic curves
 or multiple integer DH groups. The shares for each ClientKeyShareOffer
 MUST by generated independently. Clients MUST NOT offer multiple
-ClientKeyShareOffers for the same parameters. It is explicitly
-permitted to send an empty client_key_share extension as this is used
-to elicit the server's parameters if the client has no useful
+ClientKeyShareOffers for the same parameters. Servers receiving duplicate
+ClientKeyShareOffers SHOULD respond with an "illegal_parameter" alert
```
tomato42: everywhere else we're saying that the endpoint MUST send an alert and end a connection, I don't see why this should be an exception

draft-ietf-tls-tls13.md
```diff
@@ -2306,9 +2387,11 @@ values, each representing a single set of key agreement parameters;
...
+and close the connection.
+It is explicitly permitted to send an empty "client_key_share" extension as
+this is used to elicit the server's parameters if the client has no useful
```
tomato42: missing: client MUST send an `unsupported_extension` alert in case the server advertises this extension

draft-ietf-tls-tls13.md
```diff
@@ -2705,18 +2806,39 @@ The following rules apply to the certificates sent by the server:
...
+the client a certificate chain of its choice that may include algorithms
+that are not known to be supported by the client. This fallback chain MAY use
+the deprecated SHA hash algorithms, SHA-1 or SHA-224, but only if necessary.
+If the client cannot construct an acceptable chain using the provided
+certificates and decides to abort the handshake, then it MUST send an
+"unsupported_certificate" alert message and close the connection.
+
+Any endpoint receiving any certificate signed using any signature algorithm
+using an MD5 hash MUST send a "bad_certificate" alert message and close
+the connection.
```
tomato42: what about self signed trust anchors? there the signature hash is irrelevant

tomato42:
Line 2011 in e4f5ae5: it's missing the information the client MUST abort the handshake with alert `unsupported_extension`  if the server sends a extension not advertised by client (that does duplicate paragraph from line 2101)

Line 2065 in e4f5ae5: the client MUST abort the handshake with `handshake_failure` in case the selected values don't match HelloRetryRequest.

Line 2293 in  e4f5ae5: as the client may advertise the obsolete curves for backwards compatibility (there are servers which will accept only those curves), the client MUST check if the server hasn't selected one the backwards-compatible curves in TLS1.3 and abort the connection

Line 3905 in e4f5ae5: it's missing that the server receiving SSLv2 type client hello with TLS1.3 version, MUST abort connection with `protocol_version` alert

I think we should be explicit in saying that the server MUST accept Client Hello message with version field higher than 3, 4 but process it as if it was set to 3, 4.

draft-ietf-tls-tls13.md
davegarrett: An exact duplicate doesn't hurt; it's just waste. There's really no point in forcing implementations to check past the point where they find what they need. I'm not strongly against a MUST here, though.

draft-ietf-tls-tls13.md
davegarrett: If you self-sign, you can upgrade to a better hash yourself. I see no reason to add any exception for MD5 now that collisions cost under a dollar, even if collisions might not be relevant in certain use-cases. Any use of MD5 comes with the risk that someone can screw up and use MD5.

Self-signed SHA1 can be used in a fallback just the same as others.

draft-ietf-tls-tls13.md
davegarrett: I'll add one in here and for the other client-only extensions defined in this document that don't already have it.

davegarrett:
@tomato42: Comments addressed in a new commit pushed here.

The only two comments I didn't change something for are:
* I think "SHOULD" is acceptable for duplicate ClientKeyShareOffers checking, as that actually requires a little bit of effort for not much gain. Not strongly against a "MUST" though.
* I'm leaving MD5 at a total ban and expectation of fatal error, because it's garbage. I know the hash is meaningless in the context of trust anchors, but any exception to the rule of "if you see MD5, kill it" is too much of an exception at this point. (note that this does not apply to SHA1+MD5 in the context of old TLS versions, which is ugly but acceptable for backwards comparability for now)

davegarrett:
Latest commit adds a couple new error alerts so that the current catchall of "handshake_failure" can be better broken up into individual specialized errors to make debugging less painful.

davegarrett:
I've split out all the purely editorial stuff into PR #208 to reduce the size of this notably. This is based on top of that (the first two commits are that PR).

davegarrett:
Rebased on top of master now that the editorial bits are merged.

davegarrett:
@ekr: Ok, this has now been pruned down to the first two commits, which covers just the MTI extensions and updating of the alerts registry (just updating the documentation with the current standards). The other stuff is in another branch which I'll submit as a new PR after this part is dealt with.

GitHub forgot @ekr's comments above, as they were on the old commit itself rather than technically on the PR (an annoying habbit of GitHub, apparently). Still all here:
https://github.com/davegarrett/tls13-spec/commit/387499c265bd92f7af6eeec4865cb51e91a5e2f2

I've cut the allowed assumption of the TLS 1.2 SHA1 defaults when the extension isn't sent. The rest of the issues center on whether the TLC BCP is supposed to be the current standard set of expectations or not, as that's pretty much all in line with that RFC.

ekr:
There's certainly not WG consensus that the UTA BCP applies to TLS as a whole, so this is not the time to merge those requirements in. Feel free to start a message on the mailing list, but I need to get -08 out so it's going to have to wait

draft-ietf-tls-tls13.md
```diff
@@ -3324,11 +3353,33 @@ because TLS does not directly use this secret for anything
...
+applications capable of using it. All implementations MUST use the
+"signature_algorithms" extension when offering and negotiating certificate
+authenticated cipher suites. All implementations MUST use the
+"supported_groups" extension when offering and negotiating DHE or ECDHE
+cipher suites.
+
+Additionally, any implementations that allow handshake renegotiation in prior
+versions of TLS MUST implement the "renegotiation_info" extension. When
+negotiating TLS 1.2 or prior, endpoints MUST support and use the
+"renegotiation_info" extension or refuse all renegotiation attempts.
```
ekr: As noted previously, please remove the requirement to do RI and send SNI. Feel free to start a separate mailing list thread with this topic.

draft-ietf-tls-tls13.md
```diff
@@ -3771,6 +3822,9 @@ set to { 3, 0 } or less. Any endpoint receiving a Hello message with
 ClientHello.client_version or ServerHello.server_version set to { 3, 0 } MUST respond
 with a "protocol_version" alert message and close the connection.
 
+Implementations MUST NOT use the Truncated HMAC extension, defined in
+Section 7 of [RFC6066].
+
```
ekr: This text should explain why, namely that it's not relevant to AEAD.

davegarrett:
Understood. I'll split them off, then. Shouldn't take long.

draft-ietf-tls-tls13.md
```diff
@@ -2242,12 +2267,15 @@ extension type (Supported Group Extension):
...
 DHE and ECDHE) for which it offers a cipher suite.
-If the client does not supply a "supported_groups" extension with a
-compatible group, the server MUST NOT negotiate a cipher suite of the
-relevant type.  For instance, if a client supplies only ECDHE groups,
-the server MUST NOT negotiate finite field Diffie-Hellman.  If no
-acceptable group can be selected across all cipher suites, then the
-server MUST generate a fatal "handshake_failure" alert.
+Servers receiving a TLS 1.3 ClientHello without this extension yet
+offering either DHE or ECDHE cipher suites MUST send a
+"missing_extension" alert message and close the connection.
```
ekr: This seems over-aggressive. I approve of requiring the extension but I don't think there was agreement that servers MUST choke if you don't send it.

draft-ietf-tls-tls13.md
```diff
@@ -2706,11 +2734,12 @@ The following rules apply to the certificates sent by the server:
...
 
-If the client provided a "signature_algorithms" extension, then all
-certificates provided by the server MUST be signed by a hash/signature
-algorithm pair that appears in that extension. Note that this implies that a
-certificate containing a key for one signature algorithm MAY be signed using a
-different signature algorithm (for instance, an RSA key signed with a DSA key).
+All certificates provided by the server MUST be signed by a hash/signature
+algorithm pair that appears in the "signature_algorithms" extension provided
+by the client (see {{signature-algorithms}}).
+Note that this implies that a certificate containing a key for one signature
```
ekr: This needs an open issue note, since there has been debate on this point.

draft-ietf-tls-tls13.md
```diff
@@ -2706,11 +2734,12 @@ The following rules apply to the certificates sent by the server:
...
+algorithm MAY be signed using a different signature algorithm (for instance,
+an RSA key signed with a DSA key).
```
ekr: s/DSA/ECDSA/ since we are contemplating removing it

ekr:
"Understood. I'll split them off, then. Shouldn't take long."

Thanks

draft-ietf-tls-tls13.md
davegarrett: I disagree with that. If the extension is mandatory to send, then a hello without it is essentially malformed. Not requiring it to check if it's not negotiating (EC)DHE at the moment might be ok, if that's what you mean.

ekr: It's actually quite common to have requirements that something be a certain way and that the other side not check. To give an extreme example, consider the following:
http://rtcweb-wg.github.io/jsep/#sec.profile-names, where you are required
to send UDP/TLS/RTP/SAVP[F] and there is an extensive list of other values you must treat as equivalent to that.

I don't have a problem with a requirement that you MUST NOT negotiate (EC)DHE without a there being a supported_groups extension that contains the group in question and that if there is no valid configuration then you must send "handshake_failure", but there simply wasn't WG consensus for requiring anything stronger.

davegarrett:
Ok, commit updated on this PR:
* cut out SNI bit
* changed TLS 1.2 back-compat language to your suggestion
* lowered threshold for sending missing_extension in response to no supported_groups to only the case where it already causes a handshake_failure (e.g. just use the new more precise error code when the server knows that the extension is missing)
* added "open issue" notes
* changed the DSA->ECDSA in the example
* expanded on truncated HMAC bit, though I didn't go as far as citing the specific paper on its security (BCP does that); as noted, it's N/A for AEAD, anyway

Took a little bit extra time to double-check that I got everything you and Travis wanted changed. (it wanted an informative ref listing for the TLS 1.2 spec, which we apparently didn't have in the current version) GitHub's comment system is beginning to annoy me, btw. :|

I haven't added in the MTI bits for the client_key_share & pre_shared_key extensions that are in the larger branch, as you didn't review those yet, so just a note for now.

This is also rebased onto current master and I made sure the second commit is documentation only (the bulk of the changed lines are just shifting whitespace so they all fit).

draft-ietf-tls-tls13.md
martinthomson: I think that the jsep analogy here is bad. In that case, you have additional compatibility requirements.

Here, I think that the rule you have is perfect. The server should pick a cipher suite, then check named_groups. If there isn't a usable group, barf.

ekr: On Thu, Aug 27, 2015 at 10:07 AM, Martin Thomson <notifications@github.com>
wrote:

> In draft-ietf-tls-tls13.md
> <https://github.com/tlswg/tls13-spec/pull/201#discussion_r38119931>:
>
> > @@ -2242,12 +2267,15 @@ extension type (Supported Group Extension):
> >  The client MUST supply a "supported_groups" extension containing at
> >  least one group for each key exchange algorithm (currently
> >  DHE and ECDHE) for which it offers a cipher suite.
> > -If the client does not supply a "supported_groups" extension with a
> > -compatible group, the server MUST NOT negotiate a cipher suite of the
> > -relevant type.  For instance, if a client supplies only ECDHE groups,
> > -the server MUST NOT negotiate finite field Diffie-Hellman.  If no
> > -acceptable group can be selected across all cipher suites, then the
> > -server MUST generate a fatal "handshake_failure" alert.
> > +Servers receiving a TLS 1.3 ClientHello without this extension yet
> > +offering either DHE or ECDHE cipher suites MUST send a
> > +"missing_extension" alert message and close the connection.
>
> I think that the jsep analogy here is bad. In that case, you have
> additional compatibility requirements.
>
> Here, I think that the rule you have is perfect. The server should pick a
> cipher suite, then check named_groups. If there isn't a usable group, barf.
>

If you want to  make this change, please take it to the list.

davegarrett: GitHub can't even find its own comments. The email links to nowhere, though if you manually find the right collapsed outdated diff in the closed PR it's in there. I'm normally all for discussion in the bug, but not in closed PRs with this system. New issue/PR and/or list for this sort of thing, please, if just so I can keep track of it. ;)

Issue #191, opened 2015-06-30
master secret is master of very little


martinthomson:
The key derivation process uses the "master secret" as the basis for application traffic keys and both the resumption and exporter secrets.  There are lots of other secrets though.

I don't have a better name right now, so close this if none better can be found.

Also s/resumption master secret/resumption secret/ throughout, it's a little inconsistent.

ekr:
Now master of more

PR #205, opened 2015-07-20
fix spacing & minor tweaks for secrets

davegarrett:
Just a few things I noticed that I think make it a little bit easier to follow.

* Name resumption_secret consistently when used
* Show full legend for each handshake flow
* Simplify/clarify secret sources table a bit and make text cases consistent

The resumption secret naming was also apparently noticed by @martinthomson in issue #191.

(Edited to note: this PR has been pruned down to a smaller list of changes; see commit message)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1552,7 +1552,7 @@ ClientKeyShare, as shown in Figure 2:
          + ClientKeyShare        -------->
                                                        ServerHello
                                                     ServerKeyShare
-                                            {EncryptedExtensions}
+                                             {EncryptedExtensions}
                                             {ServerConfiguration*}
                                                     {Certificate*}
                                              {CertificateRequest*}
@@ -1606,7 +1606,7 @@ the first-flight data.
                                               + KnownConfiguration
                                              + EarlyDataIndication
                                                     ServerKeyShare
-                                            {EncryptedExtensions}
+                                             {EncryptedExtensions}
                                             {ServerConfiguration*}
                                                     {Certificate*}
                                              {CertificateRequest*}
@@ -1677,12 +1677,11 @@ a PSK and the second uses it:
 
 
 Initial Handshake:
-
        ClientHello
          + ClientKeyShare       -------->
                                                        ServerHello
                                                     ServerKeyShare
-                                            {EncryptedExtensions}
+                                             {EncryptedExtensions}
                                             {ServerConfiguration*}
                                                     {Certificate*}
                                              {CertificateRequest*}
@@ -1701,7 +1700,7 @@ Subsequent Handshake:
            PreSharedKeyExtension -------->
                                                        ServerHello
                                             +PreSharedKeyExtension
-                                            {EncryptedExtensions}
+                                             {EncryptedExtensions}
                                  <--------              {Finished}
        {Certificate*}
        {Finished}                -------->
@@ -3177,7 +3176,9 @@ These shared secret values are used to generate cryptographic keys as
 shown below.
 
 The derivation process is as follows, where L denotes the length of
-the underlying hash function for HKDF {{RFC5869}}.
+the underlying hash function for HKDF {{RFC5869}}. SS and ES denote
+the sources from the table above. Whilst SS and ES may be the same
+in some cases, the extracted xSS and xES will not.
 
 ~~~
   HKDF-Expand-Label(Secret, Label, HashValue, Length) =
@@ -3187,7 +3188,7 @@ the underlying hash function for HKDF {{RFC5869}}.
 
   2. xES = HKDF(0, ES, "extractedES", L)
 
-  3. master_secret= HKDF(xSS, xES, "master secret", L)
+  3. master_secret = HKDF(xSS, xES, "master secret", L)
 
   4. finished_secret = HKDF-Expand-Label(xSS,
                                          "finished secret",
@@ -3298,7 +3299,7 @@ CertificateVerify.
 ###  Diffie-Hellman
 
 A conventional Diffie-Hellman computation is performed. The negotiated key (Z)
-is used as the shared_secret, and is used in the key schedule as
+is used as the shared secret, and is used in the key schedule as
 specified above. Leading bytes of Z that contain all zero bits are stripped
 before it is used as the input to HKDF.
 


draft-ietf-tls-tls13.md
```diff
@@ -1567,6 +1567,15 @@ ClientKeyShare, as shown in Figure 2:
...
        [Application Data]        <------->     [Application Data]
+
+            *  Indicates optional or situation-dependent
+               messages that are not always sent.
+
+            {} Indicates messages protected using keys
+               derived from the ephemeral secret.
+
+            [] Indicates messages protected using keys
+               derived from the master secret.
```
ekr: I don't think repeating this is helpful

draft-ietf-tls-tls13.md
```diff
@@ -3152,15 +3176,15 @@ mode (DHE, ECDHE, PSK, etc.) and is summarized in the table below:
...
-    (w/ known_configuration)      w/ Known Key      w/ server ephemeral
+    (EC)DHE                   client ephemeral         client ephemeral
+    w/ known_configuration        w/ known key      w/ server ephemeral
 
-    PSK                         Pre-Shared Key           Pre-shared key
+    PSK                         pre-shared key           none (ES = SS)
 
-    PSK + (EC)DHE               Pre-Shared Key         Client ephemeral
+    PSK + (EC)DHE               pre-shared key         client ephemeral
                                                     w/ server ephemeral
```
ekr: I actually think that the caps makes this easier to read.

davegarrett:
Ok, to respond to comments I've pushed a new commit.

RE: capitalization: Upper or lower is fine, as long as it's consistent. I've switched it to all upper (title case) instead of all lower.

RE: legend repetition: If you're just reading one section, e.g. got to it via TOC, you might not have seen the legend (recently). I've made it less wordy now and also added a line to indicate "+" is for extensions, as I realized that wasn't necessarily clear. (and added a "+" where the notation was inconsistent) We do have an extension that was previously a message, now. The basic point is consistency; make each handshake flow make sense on its own.

I think repeating the legend is helpful, but if you really don't want to repeat it that I can cut this down to just the little stuff and table. Showing "N/A" instead of repeating things in the table is the most helpful part here, I think. It states that SS=ES in some circumstances elsewhere, but the table isn't clear that these are the exact same values, especially because the extracted ones are _not_.

davegarrett:
Edited again and repushed for some more naming consistency.

davegarrett:
Revised & rebased. Of note, I think these changes to the table make it far easier to see what's going on here.

draft-ietf-tls-tls13.md
```diff
@@ -1562,6 +1558,11 @@ ClientKeyShare, as shown in Figure 2:
        {CertificateVerify*}
        {Finished}                -------->
        [Application Data]        <------->     [Application Data]
+
+          +  extension sent in hello message
+          *  optional/situational and not always sent
+          {} protected using keys derived from ephemeral secret
+          [] protected using keys derived from master secret
```
ekr: I still don't think it helps to duplicate these.

draft-ietf-tls-tls13.md
```diff
@@ -3159,37 +3165,41 @@ The exact source of each of these secrets depends on the operational
...
 
 ~~~
-    Key Exchange            Static Secret (SS)    Ephemeral Secret (ES)
-    ------------            ------------------    ---------------------
-    (EC)DHE                   Client ephemeral         Client ephemeral
-    (full handshake)       w/ server ephemeral      w/ server ephemeral
+    Key Exchange     Static Secret (SS)    Ephemeral Secret (ES)
+    ------------     ------------------    ---------------------
+    1-RTT (EC)DHE       N/A (SS = ES)        Client Ephemeral
+                                           + Server Ephemeral
```
ekr: NA isn't correct. You use them both.

draft-ietf-tls-tls13.md
```diff
@@ -3159,37 +3165,41 @@ The exact source of each of these secrets depends on the operational
...
 
 ~~~
   HKDF-Expand-Label(Secret, Label, HashValue, Length) =
        HKDF-Expand(Secret, Label + '\0' + HashValue, Length)
 
-  1. xSS = HKDF(0, SS, "extractedSS", L)
+  1. static_secret = HKDF(0, SS, "extracted SS", L)
 
-  2. xES = HKDF(0, ES, "extractedES", L)
+  2. ephemeral_secret = HKDF(0, ES, "extracted ES", L)
```
ekr: This isn't correct. the static secret is the DH ZZ.

Also, the seeds are a technical change, so remove.

draft-ietf-tls-tls13.md
```diff
@@ -3251,17 +3261,17 @@ The following table describes the inputs to the key calculation for
...
+                                                           ServerKeyShare
 
-  Application  master "application data key expansion"    All handshake
-               secret                                      messages but
-                                                               Finished
-                                                         (session_hash)
+  Application  Master     "application data key expansion"  All handshake
+                                                             messages but
+                                                                 Finished
+                                                           (session_hash)
```
ekr: See above.

draft-ietf-tls-tls13.md
davegarrett: It's not applicable in that there is no separate value; it's equal to the other. I guess this would come across better with just "SS = ES". The issue I have with the current version is that it's verbose and not clear that these are generated from the same values. (though, the secrets themselves are different)

The "+" could be a "&" instead.

ekr: SS = ES. xSS != xES.

davegarrett: Yes, that's what I said in the previous comment. They're generated from the same values, but the extracted versions are different. This table here is specifically about the sources of these secrets.

My point here is that the notation & terminology should be more clear and spelled out, as evident by our miscommunications here. ;)

The very short abbreviations may be doing more harm than good. Maybe name SS & ES more explicitly; something like raw_ephemeral_secret?

ekr: I'd like to leave this as-is for now, but maybe I'll clean it up in the rewrite.

davegarrett:
@ekr: If you're going to be rewriting this, I'll just set those parts aside. The commit here is now just the little stuff: fix the spacing, add a line stating what SS/ES are in the list of derivations, and fix "shared_secret" (there's no such value named that; just drop the underscore).

draft-ietf-tls-tls13.md
```diff
@@ -3177,7 +3176,10 @@ These shared secret values are used to generate cryptographic keys as
 shown below.
 
 The derivation process is as follows, where L denotes the length of
-the underlying hash function for HKDF {{RFC5869}}.
+the underlying hash function for HKDF {{RFC5869}} and SS/ES denotes
+the sources from the table above. Whilst SS and ES may be the same
```
ekr: SS and ES denote

draft-ietf-tls-tls13.md
```diff
@@ -3177,7 +3176,10 @@ These shared secret values are used to generate cryptographic keys as
...
+in some cases, the extracted static_secret and ephemeral_secret
+will not.
```
ekr: xSS and xES

davegarrett:
@ekr: commit updated with fixes for your comments above

PR #208, opened 2015-07-24
editorial fixes

davegarrett:
This is all the purely editorial fixes pulled out of PR #201 to make it not as large and easier to review. Nothing in here changes specified behavior/policy.

Of note, some typos, fixes to RFC 2119 language, another cleanup of changes in TLS 1.1 stuff, some simplified wording, list "supported_groups" with a value of 10 (as it's already defined, not TBD), use "supported_groups" name consistently instead of sometimes calling it "named_groups", add RFC 5746 to the "obsoletes" field (renegotiated extension), and correct some grammar reported by Viktor Dukhovni. One commit was easily cherry-picked; the other is a squash-together of little stuff pulled out (didn't take that long to do). See the full commit message for itemized changes.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1467,7 +1467,7 @@ ServerKeyShare
   Secret (in this mode they are the same).  [{{server-key-share}}]
 
 ServerConfiguration
-: supplies a configuration for a future handshake (see {{cached-server-configuration}}).
+: supplies a configuration for zero-rtt handshakes (see {{zero-rtt-exchange}}).
 [{{server-configuration}}]
 
 EncryptedExtensions
@@ -1493,7 +1493,7 @@ CertificateVerify
 Finished
 : a MAC over the entire handshake computed using the Static Secret.
   This message provides key confirmation and
-  In some modes (see {{cached-server-configuration}}) it also authenticates the handshake using the
+  In some modes (see {{zero-rtt-exchange}}) it also authenticates the handshake using the
   the Static Secret. [{{server-finished}}]
 {:br }
 
@@ -1586,31 +1586,15 @@ will send a "handshake_failure" or "insufficient_security" fatal alert
 TLS also allows several optimized variants of the basic handshake, as
 described below.
 
-
-### Cached Server Configuration
-
-During an initial handshake, the server can provide a ServerConfiguration
-message containing a long-term (EC)DH share. On future
-connections, the client can indicate to the server that it knows the
-server's configuration and if that configuration is valid the server
-can omit both the Certificate or CertificateVerify message (provided
-that a new configuration is not supplied in this handshake).
-
-When a known configuration is used, the server's long-term DHE key is
-combined with the client's ClientKeyShare to produce SS. ES is
-computed as above.  This optimization allows the server to amortize
-the transmission of these messages and the server's signature over
-multiple handshakes, thus reducing the server's computational cost for
-cipher suites where signatures are slower than key agreement,
-principally RSA signatures paired with ECDHE.
-
-
 ### Zero-RTT Exchange
 
-When a cached ServerConfiguration is used, the client can also send
-application data as well as its Certificate and CertificateVerify
-(if client authentication is requested) on its first flight, thus
-reducing handshake latency, as shown below.
+TLS 1.3 supports a "0-RTT" mode in which the client can send
+application data as well as its Certificate and CertificateVerify (if
+client authentication is requested) on its first flight, thus reducing
+handshake latency. In order to enable this functionality, the server
+provides a ServerConfiguration message containing a long-term (EC)DH
+share. On future connections, the client uses that share to encrypt
+the first-flight data.
 
 ~~~
        Client                                               Server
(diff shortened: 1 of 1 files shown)


davegarrett:
To clarify the one alert change in here:
* Negotiated Groups currently says:
>  If no acceptable group can be selected across all cipher suites, then the server MUST generate a fatal “handshake_failure” alert.
https://tlswg.github.io/tls13-spec/#negotiated-groups
* Server Hello currently says:
> If the client proposed groups are not acceptable by the server, it will respond with an "insufficient_security” fatal alert.
https://tlswg.github.io/tls13-spec/#server-hello

So, one of them is wrong. The "insufficient_security" definition is for suites only, so I changed the latter to "handshake_failure". Stuff like this is why I'm proposing splitting these up elsewhere.

draft-ietf-tls-tls13.md
```diff
@@ -1199,8 +1204,8 @@ received after a closure alert is ignored. If a transport-level close is
 received prior to a "close_notify", the receiver cannot know that all the
 data that was sent has been received.
 
-Unless some other fatal alert has been transmitted, each party is required to
-send a "close_notify" alert before closing the write side of the connection. The
+Each party is REQUIRED to send a "close_notify" alert before closing the write
+side of the connection, unless some other fatal alert has been transmitted. The
```
ekr: s/is REQUIRED/MUST/

draft-ietf-tls-tls13.md
```diff
@@ -2236,9 +2232,12 @@ ffdhe2048, etc.
 
 Values within "obsolete_RESERVED" ranges were used in previous versions
 of TLS and MUST NOT be offered or negotiated by TLS 1.3 implementations.
+The obsolete curves have various known or theoretical weaknesses and are
+no longer considered safe for general use. The supported values are more
+than sufficient for interoperability with existing and future servers.
```
ekr: This seems over-strong. Some of the ones we are deprecating are just disfavored.

draft-ietf-tls-tls13.md
```diff
@@ -2236,9 +2232,12 @@ ffdhe2048, etc.
...
 
 Items in named_curve_list are ordered according to the client's
-preferences (favorite choice first).
+preferences (preferred choice first).
```
ekr: s/preferred/most preferred/

draft-ietf-tls-tls13.md
```diff
@@ -2368,17 +2368,20 @@ X9.62.]]
...
+  an RSA certificate, as well as on the client if the server has
+  an ECDSA certificate.
 
-- Allowing the client and server to do a 0-RTT exchange (See {{zero-rtt-exchange}})
+- Allowing the client and server to do a 0-RTT exchange
+  (see {{zero-rtt-exchange}}).
 
-The extension is defined as:
+The "extension_data" field of this extension contains a
+"KnownConfigurationExtension" value:
```
ekr: This has been OBE, so you'll need to rebase on top of PR 211

draft-ietf-tls-tls13.md
```diff
@@ -2462,19 +2468,23 @@ opts to do so, it MUST supply an Early Data Indication
...
 the "known_configuration" extension.
 
+The "extension_data" field of this extension contains an
+"EarlyDataIndication" value:
+
 %%% Hello Messages
-          enum { early_handshake(1), early_data(2),
-                 early_handshake_and_data(3), (255) } EarlyDataType;
+       enum { early_handshake(1), early_data(2),
+              early_handshake_and_data(3), (255) } EarlyDataType;
```
ekr: Also needs rebasing...

draft-ietf-tls-tls13.md
davegarrett: I think "known or theoretical weaknesses" covers this rather well, though it's obviously a bit hand-wavey. They're not necessarily broken, just weak enough to be written out of the spec. All curves in the main obsolete set are either less than 255-bit (~128-bit security level), and thus weaker than the par we're setting for the spec, or a binary/Koblitz curve, which has at least theoretical weaknesses as was mentioned on the mailing list. The two in the second set are for arbitrary curves from RFC 4492, and can be configured to use whatever, including weak curves.

The "known or theoretical" wording is imprecise but not inaccurate. More information could be added here, but I don't think it'd be that useful.

davegarrett:
Ok, this is now rebased on top of PR #211. The PR dependency tree is now:
PR #211 -> PR #208 -> PR #201 
(always_sign -> editorialfixes -> alertsandcerts)

My old WIP branches are on top of that. GitHub's graph:
https://github.com/ekr/tls13-spec/network

The latest commit here contains the couple other fixes for review comments above.

@ekr: This should address everything except for your note about the "known or theoretical" language, which I've explained above. I'm fine with changing it, if need be, but I think some sort of very brief reasoning as to why all these curves are obsolete should be in here. A "MUST NOT" with no explanation is more likely to get overlooked. The particular wording isn't all that important so long as it's clear that obsolete curves should be assumed to be unsafe for general use, if just as a precaution.

As mentioned in the PR #201, each commit there can be more or less reviewed independently, though they're all touching the same area so basing them all in one branch made things easier.

draft-ietf-tls-tls13.md
ekr: I don't think it's helpful, especially when we're talking about 571t.

I think it's fine to just say "not widely used".

davegarrett:
Ok, I revised the last commit to reword the bit about the curves. It now says "various known/theoretical weaknesses or have had very little usage" instead. The rewording should be a little more clear and is more generalized to not label everything with the same level of safety concern.

ekr:
merged separately.

PR #209, opened 2015-07-24
TLS Client Auth

Andrei-Popov:
Based on the TLS1.3 client authentication discussion at the TLS WG meeting, this PR:
1. Allows the server to send CertificateRequest at any time after the handshake is complete.
2. Allows the server to narrow down the list of acceptable client certificates by optionally specifying a list of certificate extension OIDs with their allowed values.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2743,7 +2743,8 @@ When this message will be sent:
 
 > A non-anonymous server can optionally request a certificate from the client,
 if appropriate for the selected cipher suite. This message, if sent, will
-immediately follow the server's Certificate message.
+immediately follow the server's Certificate message. The server may also send
+this message at any time after the handshake is complete.
 
 Structure of this message:
 
@@ -2757,10 +2758,16 @@ Structure of this message:
        opaque DistinguishedName<1..2^16-1>;
 
        struct {
+           opaque certificate_extension_oid<0..2^8-1>;
+           opaque certificate_extension_values<0..2^8-1>;
+       } CertificateExtension
+
+       struct {
            ClientCertificateType certificate_types<1..2^8-1>;
            SignatureAndHashAlgorithm
              supported_signature_algorithms<2..2^16-2>;
            DistinguishedName certificate_authorities<0..2^16-1>;
+           CertificateExtension certificate_extensions<0..2^16-1>;
        } CertificateRequest;
 
 certificate_types
@@ -2785,6 +2792,22 @@ certificate_authorities
   the certificate_authorities list is empty, then the client MAY
   send any certificate of the appropriate ClientCertificateType,
   unless there is some external arrangement to the contrary.
+
+certificate_extensions
+: A list of certificate extension OIDs [RFC5280] with their allowed
+  values, represented in DER-encoded format. Some certificate
+  extension OIDs allow multiple values (e.g. Extended Key Usage).
+  If the server has included a non-empty certificate_extensions list,
+  the client certificate MUST contain all of the specified extension
+  OIDs that the client recognizes. For each extension OID recognized
+  by the client, all of the specified values MUST be present in the
+  client certificate. However, the client MUST ignore and skip any
+  unrecognized certificate extension OIDs. If the client has ignored
+  some of the required certificate extension OIDs, and supplied a
+  certificate that does not satisfy the request, the server MAY at
+  its discretion either continue the handshake without client
+  authentication, or respond with a fatal unsupported_certificate
+  alert.
 {:br }
 
 The interaction of the certificate_types and
@@ -2996,9 +3019,11 @@ Certificate and CertificateVerify message.
 When this message will be sent:
 
 > This message is the first handshake message the client can send
-after receiving the server's Finished. This message is only sent if the server requests a
-certificate. If no suitable certificate is available, the client MUST send a
-certificate message containing no certificates. That is, the certificate_list
+after receiving the server's Finished. This message is also sent when the
+client receives a CertificateRequest after the handshake is complete. This
+message is only sent if the server requests a certificate. If no suitable
+certificate is available, the client MUST send a certificate message
+containing no certificates. That is, the certificate_list
 structure has a length of zero. If the client does not send any certificates,
 the server MAY at its discretion either continue the handshake without client
 authentication, or respond with a fatal "handshake_failure" alert. Also, if some
@@ -3060,6 +3085,13 @@ In particular:
   that this relaxes the constraints on certificate-signing
   algorithms found in prior versions of TLS.
 
+- If the certificate_extensions list in the certificate request message
+  was non-empty, the end-entity certificate MUST contain all of the
+  specified extension OIDs that the client recognizes. For each extension
+  OID recognized by the client, all of the specified values MUST be present
+  in the client certificate. However, the client MUST ignore and skip any
+  unrecognized certificate extension OIDs.
+
 Note that, as with the server certificate, there are certificates that use
 algorithms/algorithm combinations that cannot be currently used with TLS.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3103,6 +3135,11 @@ it in the "pre_shared_key" extension in its ClientHello
 ({{pre-shared-key-extension}}) and supplying a suitable PSK cipher
 suite.
 
+The server MAY also send a NewSessionTicket message after the handshake is
+complete, following the client CertificateVerify message. This allows the
```
martinthomson: NewSessionTicket needs to follow the client's Finished.  For one, the client's CertificateVerify won't appear in every exchange.

draft-ietf-tls-tls13.md
```diff
@@ -2743,7 +2743,8 @@ When this message will be sent:
 
 > A non-anonymous server can optionally request a certificate from the client,
 if appropriate for the selected cipher suite. This message, if sent, will
-immediately follow the server's Certificate message.
+immediately follow the server's Certificate message. The server may also send
+this message at any time after the handshake is complete.
```
martinthomson: Didn't we agree that the client would be permitted to authentication either zero or one times?  This needs to call that out.

draft-ietf-tls-tls13.md
```diff
@@ -2757,10 +2758,16 @@ Structure of this message:
        opaque DistinguishedName<1..2^16-1>;
 
        struct {
+           opaque certificate_extension_oid<0..2^8-1>;
+           opaque certificate_extension_values<0..2^8-1>;
+       } CertificateExtension
```
davegarrett: missing semicolon

draft-ietf-tls-tls13.md
```diff
@@ -2757,10 +2758,16 @@ Structure of this message:
        opaque DistinguishedName<1..2^16-1>;
 
        struct {
+           opaque certificate_extension_oid<0..2^8-1>;
+           opaque certificate_extension_values<0..2^8-1>;
```
ekr: oid seems like it should be non-empty.

draft-ietf-tls-tls13.md
```diff
@@ -2757,10 +2758,16 @@ Structure of this message:
...
+       } CertificateExtension
+
+       struct {
            ClientCertificateType certificate_types<1..2^8-1>;
            SignatureAndHashAlgorithm
              supported_signature_algorithms<2..2^16-2>;
            DistinguishedName certificate_authorities<0..2^16-1>;
```
ekr: What would you think about pushing this up into the ServerHello.?That would support client-initiated client auth.

draft-ietf-tls-tls13.md
```diff
@@ -2785,6 +2792,22 @@ certificate_authorities
   the certificate_authorities list is empty, then the client MAY
   send any certificate of the appropriate ClientCertificateType,
   unless there is some external arrangement to the contrary.
+
+certificate_extensions
+: A list of certificate extension OIDs [RFC5280] with their allowed
+  values, represented in DER-encoded format. Some certificate
+  extension OIDs allow multiple values (e.g. Extended Key Usage).
```
ekr: Does that mean that I can supply two instances of an OID or an OID with two values?

draft-ietf-tls-tls13.md
```diff
@@ -2785,6 +2792,22 @@ certificate_authorities
...
+  the client certificate MUST contain all of the specified extension
+  OIDs that the client recognizes. For each extension OID recognized
+  by the client, all of the specified values MUST be present in the
+  client certificate. However, the client MUST ignore and skip any
+  unrecognized certificate extension OIDs. If the client has ignored
+  some of the required certificate extension OIDs, and supplied a
+  certificate that does not satisfy the request, the server MAY at
+  its discretion either continue the handshake without client
+  authentication, or respond with a fatal unsupported_certificate
+  alert.
```
ekr: I feel like this text in general needs to be clearer about the separation of authority between TLS and the PKI. Specifically, as I understand it, the way I evaluate a field in this message against a given cert is by:

1. Looking to see if the extension is present in the certificate.
2. If so, following some extension-specific rule to determine if there is a match.

I.e., the data in this field need not be bitwise-equal to the extension.

Is this all correct?

draft-ietf-tls-tls13.md
```diff
@@ -2785,6 +2792,22 @@ certificate_authorities
...
+  If the server has included a non-empty certificate_extensions list,
+  the client certificate MUST contain all of the specified extension
+  OIDs that the client recognizes. For each extension OID recognized
+  by the client, all of the specified values MUST be present in the
+  client certificate. However, the client MUST ignore and skip any
+  unrecognized certificate extension OIDs. If the client has ignored
+  some of the required certificate extension OIDs, and supplied a
+  certificate that does not satisfy the request, the server MAY at
+  its discretion either continue the handshake without client
+  authentication, or respond with a fatal unsupported_certificate
```
ekr: I think you are actually proceeding with unacceptable client auth because the client has already authenticatated, right?

draft-ietf-tls-tls13.md
ekr: I think we should at this point relax the rule and say you can generate a NewSessionTicket whenever.

draft-ietf-tls-tls13.md
Andrei-Popov: Yes, will fix.

draft-ietf-tls-tls13.md
Andrei-Popov: An OID with multiple values.

draft-ietf-tls-tls13.md
Andrei-Popov: Yes, this is all correct. The extension-specific rules would typically live in a PKI library, so the TLS code does not need to be cluttered with this.

draft-ietf-tls-tls13.md
Andrei-Popov: This is to say that when the client cert is not acceptable for some reason, the server can either proceed in some way (e.g. show an “access denied” page), or terminate the connection. I believe this is what servers do in TLS<1.3.

draft-ietf-tls-tls13.md
Andrei-Popov: Yes, I’d be fine with relaxing the rule.

draft-ietf-tls-tls13.md
ekr: Yeah, I think this is just an editorial issue.

ekr: Note: need to validate that either (a) certificates never contain duplicate extensions or (b) you can have multiple instances of the same extension.

richsalz: Right now certs are pretty opaque to the protocol, and many TLS implementations.  Understanding X509v3 extensions seems way out of scope for TLS.  For example, we don't even talk about validity periods!   Strike this and leave it up to the application.

Andrei-Popov: With this proposal, certs remain opaque to the protocol.

Any OID matching code remains in the PKI library where it belongs.

Cheers,

Andrei

richsalz: OID's are borderline, but value's seem like it is going way too far in mixing the layers.  But as long as this is a MAY support, not MTI, I could live with it.

Andrei-Popov: Yes, this is optional: a client only honors the OIDs the client recognizes. And the client may not recognize any OIDs.

Andrei-Popov: RFC 5280 section 4.2 says:
“A certificate MUST NOT include more
   than one instance of a particular extension.”

Cheers,

Andrei

ekr:
Closed so we can split into two.

PR #211, opened 2015-08-03
Always sign

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -1586,31 +1586,15 @@ will send a "handshake_failure" or "insufficient_security" fatal alert
...
-
 ### Zero-RTT Exchange
 
-When a cached ServerConfiguration is used, the client can also send
-application data as well as its Certificate and CertificateVerify
-(if client authentication is requested) on its first flight, thus
-reducing handshake latency, as shown below.
+TLS 1.3 supports a "0-RTT" mode in which the client can send
+application data as well as its Certificate and CertificateVerify (if
+client authentication is requested) on its first flight, thus reducing
```
martinthomson: Does it make sense to describe client authentication as being requested, or as being permitted in this context?

draft-ietf-tls-tls13.md
```diff
@@ -2455,38 +2393,59 @@ alert.
...
   means that only handshake data is being sent. "early_data"
-  means that only data is being sent. "early_handshake_and_data"
+  means that only data is being sent. "client_authentication_and_data"
   means that both are being sent.
 {:br }
 
+The client specifies the cryptographic configuration for the 0-RTT
+data using the "configuration", "cipher_suite", and "extensions"
+values. For configurations received in-band (in a previous TLS connection)
+the client MUST (1) send the same cryptographic determining parameters
```
martinthomson: "MUST (1)" is a strange construction, I'd use a list.

draft-ietf-tls-tls13.md
```diff
@@ -2455,38 +2393,59 @@ alert.
...
+(Section{{cryptographic-determining-parameters}}),
```
martinthomson: Section{{foo}} will render very strangely.

draft-ietf-tls-tls13.md
```diff
@@ -2518,11 +2476,18 @@ can behave in one of two ways:
   process the early data. It is not possible for the server
   to accept only a subset of the early data messages.
 
-The server MUST first validate that the client's "known_configuration"
-extension is valid and that the client has suppled a valid
-key share in the "client_key_shares" extension. If not, it MUST
-ignore the extension and discard the early handshake data
-and early data.
+Prior to accepting the EarlyDataIndication extension, the server
+MUST perform the following checks.
```
martinthomson: Colon:

draft-ietf-tls-tls13.md
```diff
@@ -2864,15 +2841,22 @@ The idea is just to minimize exposure.]]
...
 : The long-term DH key that is being established for this configuration.
 
-early_data_allowed
-: Whether the client may send data in its first flight (see {{early-data-indication}}).
+early_data_type
+: The type of 0-RTT handshake that this configuration is to be used
+for (see {{early-data-indication}}). If "client_authentication"
+or "client_authentication_and_data", then the client should select
+the certificate for future handshakes based on the CertificateRequest
+parameters supplied in this handshake.
```
martinthomson: I think that this needs a lot more detail.  If the server requests client authentication in the first connection and the client does not provide a certificate, is the client permitted to connect without authentication in future?  Are they expected to copy their choice, or can they make the new connection with authentication?  What happens if the server does not request a certificate in the first connection but provides a configuration that permits client authentication?

martinthomson:
This is definitely an improvement, but I think that it covers more changes than are absolutely necessary.

draft-ietf-tls-tls13.md
ekr: I think it should be required, not optional

draft-ietf-tls-tls13.md
ekr: Willdo

draft-ietf-tls-tls13.md
ekr: I think it would be best for now to be strict.

ekr:
Can you explain what you mean with your overall comment?

martinthomson:
There are a few things in here that weren't strictly necessary.  Mostly good things that we agreed to elsewhere. Like extensions in ServerConfiguration.

ekr:
Do you think I should break this up?

martinthomson:
Nah, I just wanted to flag it so that other reviewers wouldn't miss it.

draft-ietf-tls-tls13.md
martinthomson: How would that manifest?  Only permit client authentication if client authentication were both permitted/requested and used in the previous connection?

draft-ietf-tls-tls13.md
```diff
@@ -1467,7 +1467,7 @@ ServerKeyShare
   Secret (in this mode they are the same).  [{{server-key-share}}]
 
 ServerConfiguration
-: supplies a configuration for a future handshake (see {{cached-server-configuration}}).
+: supplies a configuration for zero-rtt handshakes (see {{zero-rtt-exchange}}).
```
davegarrett: For terminology consistency, I'd suggest using "0-RTT handshakes" instead of "zero-rtt handshakes" here. The former is used in most places.

draft-ietf-tls-tls13.md
```diff
@@ -1619,6 +1603,7 @@ reducing handshake latency, as shown below.
          + ClientKeyShare
          + KnownConfiguration
          + EarlyDataIndication
+       (EncryptedExtensions)
```
davegarrett: Isn't this message optional too? (missing asterisk if it is)

draft-ietf-tls-tls13.md
```diff
@@ -2455,38 +2393,59 @@ alert.
...
 
+configuration_id
+: The label for the configuration in question.
+
+cipher_suite
+: The cipher suite which the client is using to encrypt the early data.
+
+extensions
+: The extensions required to define the cryptographic configuration
+  for the clients early data (see below for details)
```
davegarrett: missing period

draft-ietf-tls-tls13.md
```diff
@@ -2538,17 +2503,27 @@ not the the 0-RTT cryptographic transforms, including:
...
 
+In order to allow the server to decrypt 0-RTT data, the client
+needs to provide enough information to allow the server to
+decrypt the traffic without negotiation. This is accomplished
+by having the client indicate the "cryptographic determining
+parameters" in its ClientHello, which are necessary to decrypt
+the client's packets. This includes the following
+values 
 
+- The cipher suite identifier.
```
davegarrett: Needs an extra newline between these two bullet points to match the style from the rest of the document.

draft-ietf-tls-tls13.md
```diff
@@ -2538,17 +2503,27 @@ not the the 0-RTT cryptographic transforms, including:
...
-]]
+##### Cryptographic Determining Parameters
 
+In order to allow the server to decrypt 0-RTT data, the client
+needs to provide enough information to allow the server to
+decrypt the traffic without negotiation. This is accomplished
+by having the client indicate the "cryptographic determining
+parameters" in its ClientHello, which are necessary to decrypt
+the client's packets. This includes the following
+values
```
davegarrett: needs a colon

draft-ietf-tls-tls13.md
```diff
@@ -2455,38 +2393,59 @@ alert.
 #### Early Data Indication
 
 In cases where TLS clients have previously interacted with the
-server and the server has supplied a known configuration, the client
+server and the server has supplied a server configuration, the client
```
davegarrett: "server configuration" -> "ServerConfiguration {{server-configuration}}"

Using the message name and adding a cross-reference link makes learning how this stuff works easier. Also easier to search the document for something specific if the naming is consistent.

draft-ietf-tls-tls13.md
martinthomson: Personally, I'd rather have EncryptedExtensions mandatory, and simply have it empty if there are none.  Fewer chances for errors that way.

draft-ietf-tls-tls13.md
```diff
@@ -2835,19 +2810,25 @@ message is sent as the last message before the CertificateVerify.
...
+              opaque extension_data<0..2^16-1>;
+          } ConfigurationExtension;
 
           struct {
               opaque configuration_id<1..2^16-1>;
               uint32 expiration_date;
               NamedGroup group;
               opaque server_key<1..2^16-1>;
-              Boolean early_data_allowed;
+              EarlyDataType early_data_type;
```
davegarrett: You should remove the "Primitive Types" section now that you're not using this boolean type anymore.

I also suggest "allowed_early_data_types", just for clarity.

PR #218, opened 2015-08-18
Reducing AEAD expansion to a maximum of 256 bytes

### LINKED ISSUE #55: Reduce the "overhead" values in TLSCiphertext, etc.
This is currently 2^{14} + 2048 (in TLSCiphertexT) and then 2^{14} + 1024 later.

This is introduced by leaving room for compression. Should we just reduce this entirely across the board. Certainly we don't need to leave room for compression.

martinthomson:
As discussed on list.

Closes #55.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1034,7 +1034,7 @@ record_version
 
 length
 : The length (in bytes) of the following TLSCiphertext.fragment.
-  The length MUST NOT exceed 2^14 + 2048.
+  The length MUST NOT exceed 2^14 + 256.
 
 fragment
 : The AEAD encrypted form of TLSPlaintext.fragment.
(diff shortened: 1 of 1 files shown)


davegarrett:
Minor suggestions:

1) Explicitly specify the expectation of a "record_overflow" alert after the two "MUST NOT"s.
2) Also drop the issue #55 "OPEN ISSUE" note that's on line 1087-1088.
3) Amend the changelog to mention this.
4) Add a sentence noting the overhead cap has changed to the "Note: In versions of TLS prior to 1.3" paragraph, or elsewhere.

martinthomson:
As @davegarrett suggests, I've updated the PR.  I have not added the Note.  I don't like those.  The change log is enough.

davegarrett:
Why was the changelog amended as for draft-09? The draft-08 hasn't been published yet. Is it in the progress of being released? (not shown on DataTracker and releases haven't been tagged on GitHub since draft-05)

ekr:
Pilot error. Please feel free to submit a PR.

PR #221, opened 2015-08-22
very trivial fixes

davegarrett:
Consistently capitalize AEAD, fix a missing period, and fix a typo in a comment (for the sole purpose of appeasing my perfectionist tendencies ;).


ekr:
I don't think changing aead-ciphered to AEAD-ciphered in the protocol description is an improvement.

The typo fix is good :)

davegarrett:
3 lines below aead-ciphered we have TLSCiphertext. Abbreviations are already kept in caps in almost every place. I don't see why it shouldn't be consistent and do so here as well.

ekr:
I'd prefer to stay consistent with TLS 1.2.

davegarrett:
Meh, ok. It looks out of place to me, but it's not a big deal. Commit revised to just fix the missing period and typo.

PR #231, opened 2015-08-27
Shift decision to abort on unusable cert chain to client and deprecate SHA-1 for it

davegarrett:
@ekr: Another chunk that was split off of the prior PR with results of list discussions. If you intend to release draft-08 soon, then this may be too late for that, though it doesn't have anything that hasn't been discussed in detail and I can make changes if needed.

The changes here are:
SHA-1 is a SHOULD NOT with clients doing so for TLS 1.2 compat required to prioritize it last
SHA-1 is a MUST NOT for servers, unless there is no other option, in which case it MAY give it a try
MD5 is a flat MUST NOT, because it's garbage (especially by itself)
the client gets the final call on server cert chains to avoid aborts that could've worked
(the language of this last part was ironed out heavily in list discussions with Viktor Dukhovni)

Allowing the client to have the final say also makes it vaguely possible to attempt to debug it, as opposed to a server returning an error that could mean pretty much anything.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2883,7 +2883,7 @@ If the server cannot produce a certificate chain that is signed only via the
 indicated supported pairs, then it SHOULD continue the handshake by sending
 the client a certificate chain of its choice that may include algorithms
 that are not known to be supported by the client. This fallback chain MAY
-use the deprecated SHA hash algorithms, SHA-1 or SHA-224.
+use the deprecated SHA-1 hash algorithm.
 If the client cannot construct an acceptable chain using the provided
 certificates and decides to abort the handshake, then it MUST send an
 "unsupported_certificate" alert message and close the connection.
@@ -2893,7 +2893,7 @@ using an MD5 hash MUST send a "bad_certificate" alert message and close
 the connection.
 
 As SHA-1 and SHA-224 are deprecated, support for them is NOT RECOMMENDED.
-Endpoints that reject chains due to use of a deprecated SHA hash MUST send
+Endpoints that reject chains due to use of a deprecated hash MUST send
 a fatal "bad_certificate" alert message before closing the connection.
 All servers are RECOMMENDED to transition to SHA-256 or better as soon
 as possible to maintain interoperability with implementations
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -320,6 +320,9 @@ draft-09
 
 - Change HKDF labeling to include protocol version and value lengths.
 
+- Shift the final decision to abort a handshake due to incompatible
+  certificates to the client rather than having servers abort early.
+
 
 draft-08
 
@@ -2864,12 +2867,19 @@ The following rules apply to the certificates sent by the server:
   guide certificate selection. As servers MAY require the presence of the server_name
   extension, clients SHOULD send this extension.
 
-All certificates provided by the server MUST be signed by a hash/signature
-algorithm pair that appears in the "signature_algorithms" extension provided
-by the client (see {{signature-algorithms}}).
-[[OPEN ISSUE: changing this is under consideration]]
-Note that this implies that a certificate containing a key for one signature
-algorithm MAY be signed using a different signature algorithm (for instance,
+All certificates provided by the server MUST be signed by a
+hash/signature algorithm pair that appears in the "signature_algorithms"
+extension provided by the client, where possible (see {{signature-algorithms}}).
+If the server cannot produce a certificate chain that is signed only via the
+indicated supported pairs, then it SHOULD continue the handshake by sending
+the client a certificate chain of its choice that may include algorithms
+that are not known to be supported by the client. If the client
+cannot construct an acceptable chain using the provided certificates
+and decides to abort the handshake, then it MUST send an
+"unsupported_certificate" alert message and close the connection.
+
+Note that a certificate containing a key for one signature algorithm
+MAY be signed using a different signature algorithm (for instance,
 an RSA key signed with a ECDSA key).
 
 If the server has multiple certificates, it chooses one of them based on the
@@ -3078,7 +3088,8 @@ implementation need only maintain one running hash per hash type for
 CertificateVerify, Finished and other messages.
 
 > The signature algorithm and hash algorithm MUST be a pair offered in the
-client's "signature_algorithms" extension (see {{signature-algorithms}}). Note that
+client's "signature_algorithms" extension unless no valid certificate chain can be
+produced without unsupported algorithms (see {{signature-algorithms}}). Note that
 there is a possibility for inconsistencies here. For instance, the client might
 offer ECDHE_ECDSA key exchange but omit any ECDSA pairs from its
 "signature_algorithms" extension. In order to negotiate correctly, the server
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2274,8 +2274,8 @@ hash
   The use of MD5, SHA-1, and SHA-224 are deprecated. The md5_RESERVED
   and sha224_RESERVED values MUST NOT be offered by any implementations.
   The sha1_RESERVED value SHOULD NOT be offered, however clients willing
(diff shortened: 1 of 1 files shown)


davegarrett:
rebased again

draft-ietf-tls-tls13.md
```diff
@@ -2195,9 +2201,11 @@ hash
...
   "none" value is provided for future extensibility, in case of a
   signature algorithm which does not require hashing before signing.
-  The use of MD5 and SHA-224 are deprecated. The md5_RESERVED and
-  sha224_RESERVED values MUST NOT be offered or negotiated by any
-  implementation.
+  The use of MD5, SHA-1, and SHA-224 are deprecated. The md5_RESERVED
+  and sha224_RESERVED values MUST NOT be offered by any implementations.
+  The sha1_RESERVED value SHOULD NOT be offered, however clients willing
+  to negotiate use of TLS 1.2 MAY offer support for SHA-1 in the near
+  term for backwards compatibility with old servers.
```
ekr: s/near term//

draft-ietf-tls-tls13.md
```diff
@@ -2212,6 +2220,12 @@ suite indicates permissible signature algorithms but not hash algorithms.
 {{server-certificate}} and {{server-key-share}} describe the
 appropriate rules.
 
+Clients offering support for SHA-1 for TLS 1.2 servers MUST do so by listing
+those hash/signature pairs as the lowest priority (listed after all other
+pairs in the supported_signature_algorithms vector). TLS 1.3 servers MUST NOT
+use SHA-1 unless no valid certificate chain can be produced without it
```
ekr: s/use SHA-1/offer a SHA-1 signed certificate/.

draft-ietf-tls-tls13.md
```diff
@@ -2792,12 +2806,30 @@ The following rules apply to the certificates sent by the server:
...
 
-All certificates provided by the server MUST be signed by a hash/signature
-algorithm pair that appears in the "signature_algorithms" extension provided
-by the client (see {{signature-algorithms}}).
-[[OPEN ISSUE: changing this is under consideration]]
-Note that this implies that a certificate containing a key for one signature
-algorithm MAY be signed using a different signature algorithm (for instance,
+All certificates provided by the server MUST be signed by a
+hash/signature algorithm pair that appears in the "signature_algorithms"
+extension provided by the client, where possible (see {{signature-algorithms}}).
```
ekr: s/where possible/if they are able to provide such a chain/.

draft-ietf-tls-tls13.md
```diff
@@ -2792,12 +2806,30 @@ The following rules apply to the certificates sent by the server:
...
+If the server cannot produce a certificate chain that is signed only via the
+indicated supported pairs, then it SHOULD continue the handshake by sending
+the client a certificate chain of its choice that may include algorithms
+that are not known to be supported by the client. This fallback chain MAY use
+the deprecated SHA hash algorithms, SHA-1 or SHA-224, but only if necessary.
+If the client cannot construct an acceptable chain using the provided
```
ekr: s/, but only if necessary// as it is redundant.

draft-ietf-tls-tls13.md
```diff
@@ -2792,12 +2806,30 @@ The following rules apply to the certificates sent by the server:
...
+certificates and decides to abort the handshake, then it MUST send an
+"unsupported_certificate" alert message and close the connection.
+
+Any endpoint receiving any certificate signed using any signature algorithm
+using an MD5 hash MUST send a "bad_certificate" alert message and close
+the connection.
```
ekr: s/an MD5/a MD5/

draft-ietf-tls-tls13.md
```diff
@@ -2792,12 +2806,30 @@ The following rules apply to the certificates sent by the server:
...
+
+Endpoints receiving certificates using SHA-1 hashes MAY send a "bad_certificate"
+alert message and close the connection. All servers are RECOMMENDED to
+transition to SHA-256 or better as soon as possible to maintain
```
ekr: This doesn't seem quite right: If you advertised SHA-1 you should accept it. If you didn't, then you SHOULD send this alert.

ekr:
@seanturner: I believe there was rough consensus to incorporate these changes. Please confirm.

@davegarrett: this needs update and rebase.

draft-ietf-tls-tls13.md
davegarrett: http://www.grammar-monster.com/lessons/an_or_a.htm

The use of "an" here is correct. The letter 'M' is pronounced "em", which starts with a vowel sound. It's the starting sound that matters, not the letter itself.

davegarrett:
@ekr: Changes made for comments.

I rewrote the reaction to SHA-1/SHA-224 as follows:
* If you offered it, you SHOULD accept it.
* It's deprecated, thus support is NOT RECOMMENDED.
* If you do reject it, you MUST respond with the proper fatal alert before closing the connection.

draft-ietf-tls-tls13.md
```diff
@@ -2791,12 +2805,34 @@ The following rules apply to the certificates sent by the server:
...
+certificates and decides to abort the handshake, then it MUST send an
+"unsupported_certificate" alert message and close the connection.
+
+Any endpoint receiving any certificate signed using any signature algorithm
+using an MD5 hash MUST send a "bad_certificate" alert message and close
+the connection.
+
+Endpoints that offer support for SHA-1 or SHA-224 via the
+"signature_algorithms" extension SHOULD accept a chain that makes use of it.
+As these algorithms are deprecated, support for them is NOT RECOMMENDED.
```
ekr: This seems simultaneously too strong and too weak. If you're going to reject
a certificate just because it has SHA-1 or SHA-224 then you shouldn't advertise
them at all (so SHOULD is too weak). However, there's no obligation not to reject the cert for other reasons (so SHOULD is too strong).

davegarrett: Ok, I've repushed this commit here with that first sentence removed (and "these algorithms" replaced with "SHA-1 and SHA-224"). It doesn't really add anything that useful, and I think other areas do address this enough. Clients will know if they want to accept it or not; the issue for this paragraph is really just what to do when they don't, and to make sure it's clear that some won't.

ekr:
This doesn't seem right either.

If you're offering SHA-1 then you shouldn't reject a cert just because it has SHA-1 (so SHOULD seems wrong if that's what you're referring to)  but you might reject it for other reasons, so SHOULD accept also seems wrong there too.

davegarrett:
The newest language ( https://github.com/davegarrett/tls13-spec/commit/a79b9393fad161ab8412afcb9a05dd0147cfe6a3 ) just says it's NOT RECOMMENDED to support, which is just a restating of what's said in the updated "Signature Algorithms". It shouldn't be offered, but may be for back compat. The second sentence just says that if it's rejected, the error must be sent. The third is unchanged as it's just a recommendation to transition ASAP. There's no longer a mention of "SHOULD accept" in this commit. I'm saying that I don't think it needs to specially mention that it should accept SHA-1 if it was offered; that's a given. I don't think we need to add a special note to say that it MUST accept it if offered. The PR overall lets the client make whatever decisions it needs here, and as you said, there are other reasons to reject.

davegarrett:
rebased

davegarrett:
Rebased and SHA-224 disallowed in fallback chains, per interim consensus.

ekr:
@davegarrett @seanturner Is there any reason not to totally deprecate SHA-1 for signatures in CertificateVerify, since we have already committed to PSS.

@davegarrett this seems to need conflict resolution,

davegarrett:
@ekr: Yeah, git is never smart enough to merge the changelog automatically. Rebased.

davegarrett:
On Wednesday, September 23, 2015 02:36:06 pm ekr wrote:
> @davegarrett @seanturner Is there any reason not to totally deprecate SHA-1 for signatures in CertificateVerify, since we have already committed to PSS.

I think that's a good idea, but I'd like to do that in a follow-up PR, if that's alright.

davegarrett:
@ekr: Looks like we both were fixing the changelog conflict at the same time. This is merged now, so closing.

PR #232, opened 2015-08-27
expand MTI Extensions and add more strict requirements

davegarrett:
Modified extension requirements based on recent list discussion. It's a little verbose with repeated boilerplate, but it's clear no matter which section you read now. I've also moved the relevant requirements up top to a consistent place for each extension's section.

The level of strictness here is now:
* If attempting to negotiate and the mandatory extension isn't available: MUST error
* If just detecting an invalid combination, but not actively negotiating: MAY error
* SNI is clearly listed as MUST support, servers MAY require, and SHOULD error if required
* The "unsupported_extension" alert is reworded more generally and put in as the consequence for violating the pre-existing "MUST NOT" send the client-only extensions from a server requirements. Also added some more, namely the fact that "client_key_share" is clearly an error to be sent from a server.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1398,9 +1398,10 @@ missing_extension
 [[TODO: IANA Considerations.]]
 
 unsupported_extension
-: Sent by clients that receive an extended ServerHello containing
-  an extension that they did not put in the corresponding ClientHello.
-  This alert is always fatal.
+: Sent by endpoints receiving any hello message containing an extension
+  known to be prohibited for inclusion in the given hello message, including
+  any extensions in a ServerHello not first offered in the corresponding
+  ClientHello. This alert is always fatal.
 
 certificate_unobtainable
 : Sent by servers when unable to obtain a certificate from a URL
@@ -2143,6 +2144,19 @@ be taken into account when designing new extensions:
 The client uses the "signature_algorithms" extension to indicate to the server
 which signature/hash algorithm pairs may be used in digital signatures.
 
+All clients MUST send a valid "signature_algorithms" extension containing
+at least one supported SignatureAndHashAlgorithm when offering any
+certificate authenticated cipher suites.
+Servers MUST NOT negotiate use of a certificate authenticated cipher suite
+unless the client supplies a supported SignatureAndHashAlgorithm.
+If the extension is not provided and no alternative cipher suite is available,
+the server MUST close the connection with a fatal "missing_extension" alert.
+(see {{mti-extensions}})
+
+Servers MUST NOT send this extension. TLS servers MUST support receiving
+this extension. Clients receiving this extension MUST respond with an
+"unsupported_extension" alert and close the connection.
+
 The "extension_data" field of this extension contains a
 "supported_signature_algorithms" value:
 
@@ -2198,15 +2212,6 @@ suite indicates permissible signature algorithms but not hash algorithms.
 {{server-certificate}} and {{server-key-share}} describe the
 appropriate rules.
 
-All clients MUST send a valid "signature_algorithms" extension in their
-ClientHello when offering certificate authenticated cipher suites.
-Servers receiving a TLS 1.3 ClientHello offering certificate authenticated
-cipher suites without this extension MUST send a "missing_extension" alert
-message and close the connection.
-
-Servers MUST NOT send this extension. TLS servers MUST support receiving this
-extension.
-
 Note: TLS 1.3 servers MAY receive TLS 1.2 ClientHellos which do not contain
 this extension. If those servers are willing to negotiate TLS 1.2, they MUST
 behave in accordance with the requirements of {{RFC5246}}.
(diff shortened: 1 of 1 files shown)


davegarrett:
@ekr & @martinthomson: Please take a look when you get a chance. Are we all on the same page with this language? Anything that needs tweaking to focus on the above better?

draft-ietf-tls-tls13.md
```diff
@@ -1398,9 +1398,10 @@ missing_extension
 [[TODO: IANA Considerations.]]
 
 unsupported_extension
-: Sent by clients that receive an extended ServerHello containing
-  an extension that they did not put in the corresponding ClientHello.
-  This alert is always fatal.
+: Sent by endpoints receiving any hello message containing an extension
+  known to be invalid when included in the given hello message, including
```
ekr: It would be helpful to have "invalid" be clearer. E.g., "forbidden".

draft-ietf-tls-tls13.md
```diff
@@ -2143,6 +2144,18 @@ be taken into account when designing new extensions:
 The client uses the "signature_algorithms" extension to indicate to the server
 which signature/hash algorithm pairs may be used in digital signatures.
 
+All clients MUST send a valid "signature_algorithms" extension containing
+at least one supported SignatureAndHashAlgorithm when offering any
+certificate authenticated cipher suites.
+Servers attempting to negotiate use of certificate authenticated cipher
+suites for TLS 1.3 or later MUST close the connection with a fatal
+"missing_extension" alert in the event that this extension is not present.
+(see {{mti-extensions}})
```
ekr: Rephrase:
"Server MUST not negotiate use of a certificate authenticated cipher suite unless the client supplies a supported SignatureAndHashAlgorithm extension. If the client provides such a cipher suite without a SignatureAndHashAlgorithm extension the server SHOULD close..."

draft-ietf-tls-tls13.md
```diff
@@ -3390,14 +3424,29 @@ TLS-compliant application MUST implement the following TLS extensions:
...
 
-All implementations MUST use the "signature_algorithms" extension when
-offering and negotiating certificate authenticated cipher suites.
-All implementations MUST use the "supported_groups" extension when
-offering and negotiating DHE or ECDHE cipher suites.
+Additionally, all implementations MUST support use of the "server_name"
+extension with applications capable of using it.
+Servers MAY require clients to send a valid "server_name" extension.
+Servers requiring this extension SHOULD respond to a ClientHello
+lacking a "server_name" extension with a fatal "missing_extension" alert.
```
martinthomson: I don't think that this is right.  If we ever want to have SNI encryption, then this would be a problem for that.

davegarrett: SNI is already a MAY require and SHOULD send:
https://tlswg.github.io/tls13-spec/#server-certificate

There's just no statement of what to do when it's required and not given.

Encrypted SNI is essentially in the current draft now. The EncryptedExtensions message is now considered mandatory to send for clients for 0-RTT, so just sticking SNI in there should do the trick. I think the consensus was that rearranging the world to attempt to get it encrypted in a normal first-flight is a mess and not worth the effort. If we can actually get the ServerConfiguration to be easily deployable through other channels, this could allow for first-connect encrypted SNI via 0-RTT, which would be great. The text here works with all this just fine as it doesn't require a new version of the extension. One consideration might be to add a flag to the server config as to whether or not SNI should be encrypted, just in case someone would rather have 1-RTT fallback for SNI instead of expecting 0-RTT to work as announced.

davegarrett:
@ekr: New commit pushed here with changes for your comments:
* "invalid" -> "prohibited"
* Text is now: "Servers MUST NOT negotiate use of ___ cipher suites unless the client supplies a supported ___. If the extension is not provided and no alternative cipher suite is available, the server MUST close the connection with a fatal "missing_extension" alert." So, trying to find a viable cipher suite is preferred and the error is only required when there's nothing else to do.

draft-ietf-tls-tls13.md
```diff
@@ -2221,6 +2226,25 @@ Note: In versions of TLS prior to TLS 1.3, this extension was named
 "elliptic_curves" and only contained elliptic curve groups. See
 {{RFC4492}} and {{I-D.ietf-tls-negotiated-ff-dhe}}.
 
+All clients MUST send a valid "supported_groups" extension containing
+at least one group for each key exchange algorithm (currently
+DHE and ECDHE) for which it offers a cipher suite.
```
ekr: This seems like it needs to be reworder for PSK

davegarrett: How so? It's replacing this, from down below in this same section:
> The client MUST supply a "supported_groups" extension containing at
> least one group for each key exchange algorithm (currently
> DHE and ECDHE) for which it offers a cipher suite.

It's almost entirely a cut/paste from your original text.

I don't see anything that PSK needs special note for.

ekr: PSK is a key exchange algorithm. I realize it was wrongish before, but since we're here, we should fix it.

davegarrett: Well, I sort of consider however the PSK got shared out-of-band as the key exchange. PSK suites just use it.

Just adding "ephemeral" would make it stand completely apart rather easily, I guess.

draft-ietf-tls-tls13.md
davegarrett: Though, the caveat of doing encrypted SNI just via 0-RTT is that you'd probably need to have all virtual servers share the same server config key, which would mean you'd probably also want to disable early data. The alternative would be to have a second key, but this is already off-topic...

In any case, for SNI encryption to really be adopted, servers would have to opt-in to it, which means changing whatever is specified here is not going to be a problem.

davegarrett:
@ekr: Ok, repushed with one-word change to clarify to "at least one group for each ephemeral key exchange algorithm". This excludes PSK, but is still fine for (EC)DHE when used with it. Seems like the simplest fix.

Issue #222, opened 2015-08-27
Remove DSA support


ekr:
It has been suggested we should remove DSA. Tracking issue

beurdouche:
+1

ekr:
@seanturner do we have consensus on this or can you call for it on-list?

seanturner:
we should do a call, but I�ll phrase it as in is there anybody who wants to speak in favor of keeping it.

spt

davegarrett:
@seanturner: Also, please clarify if it's to just be dropped from the document, but technically still permitted, or dropped from the spec and no longer considered acceptable with TLS 1.3+.

ekr:
I would prefer we either allow it or not, but not have it exist in some intermediate status.

davegarrett:
I'm not a fan of keeping it technically permitted either, but I think there was mention of this on-list at some point. I don't remember this being explicitly decided.

ekr:
This was done.

Issue #59, opened 2014-07-07
Update text to deal with being able to use DSA with SHA > SHA-1


ekr:

davegarrett:
This should be closable now that DSA has been dropped from the spec.

PR #235, opened 2015-08-29
Remove DSA signatures

beurdouche:
For when the WG decides it is ok...

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2152,7 +2152,7 @@ The "extension_data" field of this extension contains a
        } HashAlgorithm;
 
        enum { 
-           anonymous(0), 
+           anonymous_RESERVED(0),
            rsa(1), 
            dsa_RESERVED(2), 
            ecdsa(3), 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3052,9 +3052,6 @@ Hash used for the handshake.
 Note: Alerts and any other record types are not handshake messages
 and are not included in the hash computations. Also, HelloRequest
 messages and the Finished message are omitted from handshake hashes.
-The input to the client and server Finished messages may not be
-the same because the server's Finished does not include the client's
-Certificate and CertificateVerify message.
 
 
 ###  Client Certificate
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -740,7 +740,7 @@ or its successors.  Data to be signed/verified is hashed, and the
 result run directly through the ECDSA algorithm with no additional
 hashing.  The default hash function is SHA-1 {{SHS}}.  However, an
 alternative hash function, such as one of the new SHA hash functions
-specified in FIPS 180-4 {{SHS}} may be used instead if the certificate
+specified in FIPS 180-4 {{SHS}} SHOULD be used instead if the certificate
 containing the EC public key explicitly requires use of another hash
 function.  (The mechanism for specifying the required hash function
 has not been standardized, but this provision anticipates such
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -740,7 +740,7 @@ or its successors.  Data to be signed/verified is hashed, and the
 result run directly through the ECDSA algorithm with no additional
 hashing.  The default hash function is SHA-1 {{SHS}}.  However, an
 alternative hash function, such as one of the new SHA hash functions
-specified in FIPS 180-2 may be used instead if the certificate
+specified in FIPS 180-4 {{SHS}} may be used instead if the certificate
 containing the EC public key explicitly requires use of another hash
 function.  (The mechanism for specifying the required hash function
 has not been standardized, but this provision anticipates such
@@ -2263,7 +2263,7 @@ The "extension_data" field of this extension contains a
 secp256r1, etc.
 : Indicates support of the corresponding named curve.
   Note that some curves are also recommended in ANSI
-  X9.62 {{X962}} and FIPS 186-2 {{DSS}}.
+  X9.62 {{X962}} and FIPS 186-4 {{DSS}}.
   Values 0xFE00 through 0xFEFF are reserved for private use.
 
 ffdhe2048, etc.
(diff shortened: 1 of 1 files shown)


davegarrett:
I only see a couple spots left that need updating:
1) Update the version number to FIPS 186-4 in "Negotiated Groups" to match the updated DSS reference.
2) Update the version number to FIPS 180-4 in "Digital Signing" to match the current SHS reference, while you're at it.

davegarrett:
To cross-ref: this PR fixes issue #222 and obsoletes issue #59

draft-ietf-tls-tls13.md
```diff
@@ -2204,7 +2186,9 @@ signature
   The values indicate anonymous signatures, RSASSA-PKCS1-v1_5
   {{RFC3447}}, DSA {{DSS}}, and ECDSA {{ECDSA}}, respectively.  The
   "anonymous" value is meaningless in this context but used in
-  {{server-key-share}}.  It MUST NOT appear in this extension.
+  {{server-key-share}}. It MUST NOT appear in this extension.
```
davegarrett: minor tweak suggestion: merge these two sentences to drop the "it" for slightly better clarity, e.g.
> The
>  "anonymous" value is meaningless in this context, but used in
>  {{server-key-share}}, and MUST NOT appear in this extension.

draft-ietf-tls-tls13.md
```diff
@@ -2204,7 +2186,9 @@ signature
...
+  The use of DSA is deprecated. The dsa_RESERVED value MUST NOT 
+  be offered or negotiated by any implementation.
```
davegarrett: Actually, disregard my previous note on this part of the text. Don't improve the anon bit; remove it. It appears to be obsolete, and this value isn't used anymore. The Server Key Message in that section is fundamentally different now. Change the "anonymous" value in the enum to "anon_RESERVED" and change this text to something like this, instead:
> The use of DSA and the anonymous value are deprecated.
> The anon_RESERVED and dsa_RESERVED values MUST
> NOT be offered or negotiated by any implementation.


>   anonymous_RESERVED placeholder value is meaningless in this context but used in
>  {{server-key-share}}.  It MUST NOT appear in this extension.
+  {{server-key-share}}. It MUST NOT appear in this extension.
+  The use of DSA is deprecated. The dsa_RESERVED value MUST NOT 
+  be offered or negotiated by any implementation.

davegarrett: Actually, disregard my previous note on this part of the text. Don't improve the anon bit; remove it. It appears to be obsolete, and this value isn't used anymore. The Server Key Message in that section is fundamentally different now. Change the "anonymous" value in the enum to "anon_RESERVED" and change this text to something like this, instead:
> The use of DSA and the anonymous value are deprecated.
> The anon_RESERVED and dsa_RESERVED values MUST
> NOT be offered or negotiated by any implementation.

ekr:
@seanturner  do we have consensus to merge this PR?

draft-ietf-tls-tls13.md
```diff
@@ -2202,9 +2184,10 @@ hash
 signature
 : This field indicates the signature algorithm that may be used.
   The values indicate anonymous signatures, RSASSA-PKCS1-v1_5
-  {{RFC3447}}, DSA {{DSS}}, and ECDSA {{ECDSA}}, respectively.  The
-  "anonymous" value is meaningless in this context but used in
-  {{server-key-share}}.  It MUST NOT appear in this extension.
+  {{RFC3447}}, DSA {{DSS}}, and ECDSA {{ECDSA}}, respectively.
+  The use of DSA and anonymous are deprecated. The anonymous_RESERVED
+  and dsa_RESERVED values MUST NOT be offered or negotiated by any
+  implementation.
```
ekr: A little birdie tells me that we're not going to actually forbid DSA in the spec right now, so please remove this last sentence. Also "is deprecated" I believe.

draft-ietf-tls-tls13.md
```diff
@@ -3084,6 +3051,7 @@ The input to the client and server Finished messages may not be
 the same because the server's Finished does not include the client's
 Certificate and CertificateVerify message.
 
+
```
ekr: This seems superfluous.

draft-ietf-tls-tls13.md
davegarrett: The "are deprecated" is correct. It's "X and Y are" OR "X or Y is".

If we're going to continue to wait for more complaining about DSA on the list, just change the "MUST NOT" to a "SHOULD NOT" for now. This does mean two parts, though, as the anonymous value should stay a "MUST NOT". So, let's just make this bit:

> The anonymous_RESERVED value was used internally in prior versions of TLS and MUST NOT be offered or negotiated by any implementation of any TLS version. The dsa_RESERVED value is deprecated as of TLS 1.3 and SHOULD NOT be offered or negotiated by any implementation.

draft-ietf-tls-tls13.md
```diff
@@ -739,28 +732,12 @@ implementations MUST accept both without parameters and with NULL parameters.
...
-refer to the standard, and it uses "DSS" in the code point definitions for
-historical continuity.
-
 All ECDSA computations MUST be performed according to ANSI X9.62 {{X962}}
 or its successors.  Data to be signed/verified is hashed, and the
 result run directly through the ECDSA algorithm with no additional
 hashing.  The default hash function is SHA-1 {{SHS}}.  However, an
 alternative hash function, such as one of the new SHA hash functions
-specified in FIPS 180-2 may be used instead if the certificate
+specified in FIPS 180-4 {{SHS}} may be used instead if the certificate
```
davegarrett: Just noticed this. This "may" should probably be changed to a "SHOULD".

draft-ietf-tls-tls13.md
ekr: I agree that "X or Y are deprecated" but this says "The use of X and Y are deprecated" so the subject is "use" not "X and Y".

davegarrett: On Wednesday, September 16, 2015 06:25:55 pm ekr wrote:
> I agree that "X or Y are deprecated" but this says "The use of X and Y are deprecated" so the subject is "use" not "X and Y".

That makes sense, though both sound right to me... Now I don't know which is correct, but I think it's the current version. English is annoying. :p

Regardless, the anon is a MUST NOT and dsa isn't (yet), so it's getting changed either way.

beurdouche:
Updated, rebased, merged conflicts with PSS patch and force pushed.

ekr:
Merged

Issue #214, opened 2015-08-04
Extensionify ServerKeyShare


davegarrett:
ClientKeyShare was converted from a message into an extension. I suggest doing the same for ServerKeyShare. This would make the two consistent with each-other and work towards issue #184.

ekr:
Consensus in interim is "let's do it".

davegarrett:
Here's an early WIP:
https://github.com/davegarrett/tls13-spec/commit/66d2fb35a46157641d63944355e64b02e1bfd45d

The client extension and server message already use the same struct, so just merging them into one general extension is relatively simple and better than two almost identical extensions.

davegarrett:
Ok, PR #268 filed for this. Will probably need some changes, so feedback / drive-by reviews would be appreciated in that PR.

PR #236, opened 2015-08-31
less verbose/redundant mandatory extension boilerplate

davegarrett:
After reading through the new parts of the published v08 draft, I think my PR to clean up the mandatory extension handling was a little too verbose and had too many repeated sentences. This refactors it a little bit to just explicitly define two terms, MANDATORY TO SEND and CLIENT ONLY, and just use them where applicable. (all caps to be clear and quick to see on skim, similar to RFC 2119 terminology) It's basically the same language, just generalized and not repeated 3 or 4 times verbatim.

The one repeated sentence could probably be shorter, but I think it's more clear when stated explicitly for each section.

Overall, there's less text now and each extension's section is significantly less verbose.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2202,23 +2202,24 @@ be taken into account when designing new extensions:
   the possibility of version rollback should be a significant consideration in
   any major design change.
 
+Note: As of TLS 1.3, some extensions are REQUIRED to be sent in order to
+negotiate certain classes of cipher suites, as opposed to prior versions
+where extensions were generally optional. See {{mti-extensions}} for a list
+as well as details about how to respond in the event that any of these core
+extensions are mishandled.
+
 ####  Signature Algorithms
 
 The client uses the "signature_algorithms" extension to indicate to the server
 which signature/hash algorithm pairs may be used in digital signatures.
 
-All clients MUST send a valid "signature_algorithms" extension containing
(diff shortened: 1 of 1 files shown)


davegarrett:
rebased

davegarrett:
rebased again

@ekr: This is just an editorial cleanup of my recent additions. Please take a look at this at some point (and similarly PR #234) to see if it looks good or needs changes.

davegarrett:
Yes, that idea does sound better than doing it in paragraphs. I'll give redoing it like that a shot. I'll update this PR when I have something.

davegarrett:
Ok, revised rather significantly. (PR #214 is not yet updated) I tried doing the list style heading, but it just didn't seem to work all that well. The biggest reduction in repetition in this is just moving the client-only bit down into the MTI extensions section and just listing the 3 there with the applicable requirements. This version here is shorter than the current draft, but not as short as the previous version of this PR. More clear to read, overall, though.

davegarrett:
The text here assumes omitting the ClientKeyShare is allowed, by the way. Landing that PR first is probably best, after which I'll fix the conflicts here.

davegarrett:
merged here:
https://github.com/tlswg/tls13-spec/commit/bdf899333ba2ed6922183d6aaff9d4c8c5d2a18f
and tweaked more here:
https://github.com/tlswg/tls13-spec/commit/07b18ebd34241865de95db78e78cf8ff124842e8

Closing.

PR #243, opened 2015-09-16
update reference for RFC7627 & linkify RFC4492 author notes

davegarrett:


davegarrett:
Travis fails because it doesn't know about the new RFC yet. I don't know what's needed to fix that.

ekr:
@martinthomson can you update our local copy of the rfc bib?

martinthomson:
Updated.  You should poke travis again.

ekr:
I'm just going to assume this is cool and merge.

Issue #154, opened 2015-03-14
add ecdsa_sign to Client Request section


cem-:
ecdsa_sign should be an included in the ClientCertificateType enum definition as well as the certificate_types list in the client request section. It is already present in the client certificate section.

cem-:
per ekr, this section is going away:
https://www.ietf.org/mail-archive/web/tls/current/msg15504.html

PR #245, opened 2015-09-17
Minor fix of syntax for SigAlg enum

beurdouche:
Missed that when rebasing. sry.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2151,13 +2151,13 @@ The "extension_data" field of this extension contains a
            (255)
        } HashAlgorithm;
 
-       enum { 
+       enum {
            anonymous_RESERVED(0),
-           rsa(1), 
-           dsa_RESERVED(2), 
-           ecdsa(3), 
-           (255) }
-         SignatureAlgorithm;
+           rsa(1),
+           dsa_RESERVED(2),
+           ecdsa(3),
+           (255)
+       } SignatureAlgorithm;
 
        struct {
            HashAlgorithm hash;


ekr:
@beurdouche @davegarrett @martinthomson we're collecting a lot of foo_RESERVEDs. 

I'm thinking of modifying mk-appendix.py to strip them out of the syntax in the main text,
but leave them in the appendix. I think this would improve clarity. Thoughts?

beurdouche:
In general I am in favour of cleaning up …
Do you mean removing from both the enums and the text and creating a section in the annexes for “deprecated stuff” ?

ekr:
My plan was:

1. Remove the references in the text
2. Leave the references in the in-line syntax in the .md file
3. Strip the references in the in-line syntax in the rendered file
4. Leave the references in the appendix.

So, for instance, in the case of this structure, I would do:

ekr:
enum {
           rsa(1),
           ecdsa(3),
   (255) }	

In the main document, and remove the discussion of the reserved ones in the
text, but the appendix at the end would have the syntax as-is.

beurdouche:
I like it, we should make sure that people notice the deprecated stuff though.

davegarrett:
I think I'm on-board with moving all deprecations to a single appendix to get them mostly out of the main document, but I don't think doing it purely through automation is the best approach. For example, your cited example could now be written all on one line and be more clear. I also don't like the idea of the text in the rendered version and version we edit being that different. I think it might be better to have both the full and abridged versions be defined in the main text, next to each-other, but have the script cut out the full version to move to the appendix and leave the abridged version in place. This way we have two sets that get edited, but they're right next to each-other so we won't get them out of sync easily.

Each enum with any "_RESERVED" could have it's text about deprecated values moved to a dedicated "Deprecated Values" section in the appendix, but with something like the following repeated in each area that does so:

"This registry contains obsolete values that were used in previous versions of TLS that MUST NOT be used as of this version of the specification for safe interoperability. See {{newsection-relevantsubsection}} for details."

That new section would have a placeholder for each full version to paste in (e.g. a "%%!"-style reference) followed by the relevant explanations of deprecations and their use requirements.

ekr:
@davegarrett I'm really not enthused about having two copies. The whole point of the automation was to remove that.

I don't generally think we need any text about the reserved values at all. The reason to list them in the document is to avoid inadvertant reassignment (which I've already accidentally done once) but this doesn't apply to the text.

davegarrett:
@ekr: There's so many new deprecations that I think they need to be mentioned quite clearly to avoid implementations that transition from TLS 1.2 to TLS 1.3 from keeping them around accidentally.

ekr:
@davegarrett some yes, some no. I don't think that (for instance) we need to have specific descriptions of the deprecated alerts.

davegarrett:
@ekr: As to how to do this moving without 2 copies, maybe we could have the script be smart enough to reformat the enum as-needed? Maybe, some kind of annotation we could add to enums that would trigger the script stripping the reserved values to also reformat the enum to automatically linewrap as-needed. Then, we'd do what you're suggesting, auto-strip *_RESERVED, but for the short ones the script would prettify them to shorten them too. (basically just involves stripping linebreaks in between the braces and making the comma+space delimiting consistent)

I don't think the alerts descriptions are a problem, though I don't disagree that they're not that helpful. It's the cluttering up of each other section with deprecated stuff that looks like a bigger problem.

richsalz:
Please take this to the mailing list, not here.  And in doing so, please explain what "I don't disagree that they're not that helpful" means, with fewer negatives :)

davegarrett:
Sure, I used a triple-negative, but I did it correctly... ;)
Translated: I agree that full descriptions of obsolete alerts are not productive in-line with the main index of alerts.

@richsalz: This is a purely editorial discussion, not really a TLS WG issue. Personally, I generally think this is the sort of thing that does belong in Github, but I have no objection to it being moved elsewhere if needed due to WG policy.

That said, I think this discussion is basically over as I'm agreeing to @ekr's idea with the caveat that I'd like a way to specify individual enums to have the script also strip linebreaks from.

ClientCertificateType is the most galling example, by the way. It's now mostly _RESERVED, though it needs updating for ECDSA (issue #154 should probably be re-opened until this is dealt with, somehow).

ekr:
@davegarrett: I think the plan is just to remove ClientCertificateType, so that should stop being a problem.

beurdouche:
If automatisation is a pain, I am volunteering to do it by hand if you want, also this might be safer...
There is 41 mentions of keyword "reserved" and 10 of "deprecated" in the spec.

ekr:
Let me take a crack at automating it first and then we can touch base
later...

PR #249, opened 2015-09-21
Padding with opaque

dkg:
This introduces a padding message for handshakes, as well as a padding approach for application data using a TLS opaque string.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1772,7 +1772,8 @@ processed and transmitted as specified by the current active session state.
            server_key_share(7), encrypted_extensions(8),
            certificate(11), reserved(12), certificate_request(13),
            reserved(14), certificate_verify(15), reserved(16),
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -811,10 +813,15 @@ all possible attacks against it. As a practical matter, this means that the
 protocol designer must be aware of what security properties TLS does and does
 not provide and cannot safely rely on the latter.
 
-Note in particular that type and length of a record are not protected by
-encryption. If this information is itself sensitive, application designers may
-wish to take steps (e.g., padding, cover traffic) to minimize information leakage.
+The padded variant of the application data protocol is designed to
+obscure the length of the exact data transmitted. However, padding
+cannot provide perfect security, and statistical attacks on padding
+may reveal some information regardless. Without using padding, the
```
martinthomson: Nit: s/may/could

PR #264, opened 2015-09-23
Forbid SHA-1 in CertificateVerify

ekr:
This completes the removal of SHA-1 for CertificateVerify. There's no reason to allow it there at all


martinthomson:
LGTM

PR #266, opened 2015-09-24
Explaining the DoS potential from empty records

martinthomson:
I think that this is OK.  I also tweaked the language in the paragraph.


ekr:
This seems kind of unnecessary.

davegarrett:
It's probably worth dedicating a section in the appendix to covering DoS risks in detail, rather than just sticking one point inline.

PR #267, opened 2015-09-24
Add MTI cipher suites. Fixes #32

### LINKED ISSUE #32: Determine replacement MTI cipher suite
Now that we are removing static RSA cipher suites, we need a new MTI cipher suite.

ekr:
@seanturner please confirm this matches consensus

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -327,6 +327,8 @@ draft-09
 
 - Deprecate SHA-1 with signatures.
 
+- Add MTI algorithms.
+
 
 draft-08
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2451,7 +2451,7 @@ offers
 {:br }
 
 Clients may offer an arbitrary number of ClientKeyShareOffer
-values, each representing a single set of key agreement parameters;
+values, each representing a single set of key exchange parameters;
 for instance a client might offer shares for several elliptic curves
 or multiple integer DH groups. The shares for each ClientKeyShareOffer
 MUST by generated independently. Clients MUST NOT offer multiple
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3518,9 +3520,23 @@ other than for computing other secrets.)
...
-[TODO:Needs to be selected](https://github.com/tlswg/tls13-spec/issues/32).
-(See {{cipher-suites}} for the definition.)
+TLS-compliant application MUST implement the following cipher suites:
+
+~~~~
+    TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 
+    TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+~~~~
+
+These cipher suites MUST support both digital signatures and key agreement
```
davegarrett: "key agreement" -> "key exchange"

This document uses "key exchange" almost everywhere. Might even be a good idea to change the 2 instances of "agreement" to "exchange" to be consistent.

To be explicitly clear, it could be: "ECDHE key exchange"

draft-ietf-tls-tls13.md
```diff
@@ -3518,9 +3520,23 @@ other than for computing other secrets.)
...
 
 In the absence of an application profile standard specifying otherwise, a
-TLS-compliant application MUST implement the cipher suite
-[TODO:Needs to be selected](https://github.com/tlswg/tls13-spec/issues/32).
-(See {{cipher-suites}} for the definition.)
+TLS-compliant application MUST implement the following cipher suites:
+
+~~~~
+    TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 
+    TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
```
davegarrett: A rather dumb nit: These are ordered "ECDSA, RSA" but the other pair is ordered "RSA, ECDSA". Could you please swap the order of one of these pairs? (I don't care which) It doesn't really affect anything; just looks odd to me. :p

Issue #190, opened 2015-06-29
Omit ClientKeyShare if it's empty


ekr:
Now that ClientKeyShare is looking like an extension, should we omit it if it's empty.

ekr:
Consensus: omit if empty

martinthomson:
Of course, if ServerKeyShare is the same extension, then it can't be empty from the client.

davegarrett:
It could probably be omitted in a system with one generalized key sharing extension, however I think it's simpler to just spend the 6 bytes for an empty offer than to have an extension with special permission to be added to a retried ClientHello. I don't like special cases if I can avoid them. I can't say for sure until I'm done with the patch and ready to submit a PR, though.

martinthomson:
Actually, thinking on this more, you can safely omit it.  It prevents the server from generating shares, but that isn't actually bad.  It prevents the client from being tricked by a falsified `HelloRetryRequest`.  In any case, you can't use the server's share until after the `HelloRetryRequest` and subsequent `ClientHello` anyway.

ekr:
Omitted.

PR #268, opened 2015-09-26
unified key share extension (fixes issue #214)

davegarrett:
Fixes issue #214. Based on top of PR #236. In addition, this should be considered dependent on PR #267.

Merging the server & client key sharing into a single extension simplifies things nicely. The result is an extension that looks almost identical to the PSK extension (though, probably should stay separate). Making this change necessitated touching quite a few other areas in minor ways, mostly renaming. Of note, I also revised much of the text around HelloRetryRequest. The existing explanation was a little hard to follow, not the least of which was due to issues around a double negative that needed revising. I didn't change any fundamental procedure here, but I think this revision makes it easier to read. Rather than leave the new section for the new extension with a TODO at the end, I also added a paragraph for recommendations on what to send. (it notes X25519, so MTI ciphers should land first)

Note that the issue of how to do the handshake hash for a retried hello is still out-standing (issue #104), so for the moment things assume that the full transcript starts at the first ClientHello, which is the safe route. Dropping that first exchange without authenticating it in some way opens up a downgrade attack, so until a good way to deal with that is in here, I think it should expect the full transcript be used in the session hash.

Issue #190 also needs to be addressed, and at this point I think the simplest thing to do is maintain consistency and state that the KeyShareOffer extension is always required when offering any (EC)DHE cipher suites, even if it's empty to request server selection. This avoids needing to state a special case where the retried ClientHello is allowed to introduce a new extension not previously included in its first attempt. It's only 6 bytes. (I could make it 4 by having empty extension_data instead of empty client_shares, but I don't think that's worth the 2 bytes there)

The last issue I brush up against is issue #185, for which I have an older PR #189 still in the queue. It's not directly involved in this change; just related.

Feedback would be appreciated.

Again, this is based on top of PR #236. A diff without its commit included is here:
https://github.com/davegarrett/tls13-spec/compare/extensionboilerplate...davegarrett:keyshare


davegarrett:
@ekr: Ok... This now depends on 2 other PRs, which of course confuses GitHub greatly. The HelloRetryRequest are separate and the editorial changes to the boilerplate are revised. The rest that was here had to be squashed because git can't handle it properly otherwise. The only difference here from what was before, in terms of actual content, is the addition of the word "fatal" before '"handshake_failure" alert' and an added link to "(see {{incorrect-dhe-share}})".

davegarrett:
@ekr: Rebased and cleaned up. The first commit in this PR is #236. The second two are PR #275. The last is what this PR is for. I really wish GitHub handled dependencies better here. It's either basing it in a tree like this, or separate PRs that inevitably have merge conflicts. Apologies for the confusion & excess bugspam from this.

The commit should be easier to read now that the retry request part is fixed and factored out. To restate, almost everything outside of the "Key Share Offer" section and deletion of "Server Key Share" is renaming. (also a correction of the ExtensionType in the enum; the old extension type name had an 's' in that case and not in others)

davegarrett:
All the necessary git incantations have been performed again and all 3 PRs should merge onto current master cleanly, at the moment. (this time with merge commits so that GitHub's Network Graph shows it properly)

ekr:
@davegarrett I am stuck in meetings through end of today. I have a long flight tomorrow and will try to take a look then.

davegarrett:
On Friday, October 09, 2015 08:44:56 am ekr wrote:
> @davegarrett I am stuck in meetings through end of today. I have a long flight tomorrow and will try to take a look then.

Thanks for the update. Due to the way I've split things up here, I would recommend reviewing in the following order:

PR #275 (all comments addressed), PR #236 (fully revised; less drastic change, now), PR #268 (this, now one commit)

davegarrett:
To make this simpler to review, I've split the mass renaming out into a separate commit. No changes to the total changeset; just split into two commits now.

ekr:
Sorry, wasn't able to get to this. It's going to have to be next week.

draft-ietf-tls-tls13.md
```diff
@@ -2357,84 +2372,109 @@ extension type (Supported Group Extension):
...
+descending priority: X25519, secp256r1, & ffdhe3072. Note that
+clients MUST NOT offer KeyShare values for key exchange methods
+not specified in the set of cipher suites offered in its ClientHello.
+(e.g. the recommendation of ffdhe3072 is not applicable if not
+offering any DHE cipher suites) Clients with prior expectations of
+what groups are likely to be supported by the server will generally
+want to include those groups in addition to, or instead of, this
+general set. Groups stated as mandatory to implement for ciphers
+offered by the client are RECOMMENDED to be sent in most, if not
+all, offers. (see {{mti-cipher-suites}})
```
ekr: Please do not intermix multiple types of changes in PRs like this. We don't have consensus on the priority of the groups and given that P-256 is MTI and X25519 is not, we shouldn't be recommending X25519 over P-256. Replace this section with the TODO.

draft-ietf-tls-tls13.md
```diff
@@ -2068,28 +2073,41 @@ selected_group
...
+
+If the HelloRetryRequest is verified as acceptable, the client MUST send
+a new ClientHello with an updated KeyShareOffer extension to the server.
+This KeyShareOffer MUST contain the exact previous set of offered keys
+with a new KeyShare value for the server's selected_group
+appended to the end of the list (or a new extension with exactly one
+value, if none was provided initially). All other fields in this new
+ClientHello MUST match the first attempt. Note that omitting any
+information previously included in the initial ClientHello introduces
+the risk of downgrade attack, as this retry exchange is unauthenticated.
```
ekr: Please just merge the extensions in this PR, not add a bunch of other material.

draft-ietf-tls-tls13.md
```diff
@@ -2357,84 +2372,109 @@ extension type (Supported Group Extension):
...
-MUST by generated independently. Clients MUST NOT offer multiple
-ClientKeyShareOffers for the same parameters. It is explicitly
-permitted to send an empty "client_key_share" extension as this is used
-to elicit the server's parameters if the client has no useful
-information.
+server_offer
+: A single KeyShare value for the negotiated cipher suite.
+  Servers MUST NOT send a KeyShare for a group not offered
+  by the client.
+{:br }
```
ekr: It would be better to have the client and server extensions be identical.

davegarrett: And permit the server to respond to clients with multiple KeyShares to pick from?

ekr: No. It should be required to send exactly one.

davegarrett: On Monday, October 12, 2015 07:50:26 pm ekr wrote:
> No. It should be required to send exactly one.

Why introduce that inconsistency, though? If you'd like to simplify it like that, ok, but the PSK extension should be doing the same thing. It's designed the same way as this is, currently, and it only has slightly different client/server field names.
https://tools.ietf.org/html/draft-ietf-tls-tls13-09#section-6.3.2.4

Or, do you just want to assign PSK a NamedGroup and merge it into this? (in a followup) One omni-field for all the key agreements would be far simpler.

draft-ietf-tls-tls13.md
davegarrett: Well, the CFRG was requested to provide new curves because we're not satisfied with the ones we have, so putting the newer and better performing one that was requested first isn't exactly a stretch. (especially because it's only a general recommendation) I thought this was generic enough, but I'll cut it out and treat it as a different issue.

draft-ietf-tls-tls13.md
davegarrett: On Monday, October 12, 2015 07:22:04 pm ekr wrote:
> Please just merge the extensions in this PR, not add a bunch of other material.

This isn't directly part of this PR. It's PR #275, which this is based on top of. (all comments were addressed in that) If that's now on hold, I'll decouple this from it, do the renames in the existing text, and fix the merge conflicts later, then.

davegarrett:
Rebased on top of master (which has changes from PR #236 merged in and further tweaked). This is no longer based on any other PRs. (it will now conflict with PR #275) I really wish git/GitHub was smart enough to handle dependencies without so much fiddling. The mass renaming ended up getting squashed back into the main commit in the process. The second commit is for making it a generic field, as requested, however this makes it oddly different from the PSK extension now.

davegarrett:
Added a tiny followup commit to drop the vestigial "server_key_share_RESERVED(7)".

davegarrett:
@ekr: When you merged this, you checked in https://github.com/tlswg/tls13-spec/commit/6922f873e67e16dd7b499b7b0330743568591c9a but then reverted it in https://github.com/tlswg/tls13-spec/commit/74c4c7da1bf5bcea4f75cfc38c1aedb406098d3d . I also see that you did not check in the commit to merge the client/server fields as requested. I take it you decided against that?

ekr:
I left the removal of server_key_share_RESERVED but reverted the message renumbering, since that seemed unnecessary.

Yes, I was persuaded by your argument that merging was a bad idea. I think I noted that in my response to the PR.

davegarrett:
On Saturday, October 17, 2015 03:16:49 pm ekr wrote:
> I think I noted that in my response to the PR.

I don't see a response to that. In any case, yes, merging them would likely be over-generalizing.

I support the name change, by the way, as it makes it a little more clear. (new PR for a minor followup fix, as you missed as spot)

davegarrett:
On Saturday, October 17, 2015 03:16:49 pm ekr wrote:
> I left the removal of server_key_share_RESERVED but reverted the message renumbering, since that seemed unnecessary.

I saw it as the other way around: keeping the numbering is unnecessary. Both values are new to TLS 1.3, so there's no reason to leave a gap in the values.

PR #270, opened 2015-10-03
Reserved in appendix

ekr:


ekr:
Note: un_RESERVEDing SHA-1 and DSA is intentional. The point is that that things which are RESERVED are totally forbidden. Because you can offer SHA-1, we need the code point.
If we decide to make DSA MUST_NOT, then I'll mark it reserved and update the text.

draft-ietf-tls-tls13.md
```diff
@@ -1428,11 +1413,6 @@ decrypt_error
   to correctly verify a signature or validate a Finished message.
   This alert is always fatal.
 
-export_restriction_RESERVED
-: This alert was used in some earlier versions of TLS. It MUST NOT
-  be sent by compliant implementations. This alert is always fatal.
-
-protocol_version
```
davegarrett: collateral damage; "protocol_version" name shouldn't be cut here

davegarrett:
On Saturday, October 03, 2015 04:42:44 pm ekr wrote:
> Note: un_RESERVEDing SHA-1 and DSA is intentional. The point is that that things which are RESERVED are totally forbidden. Because you can offer SHA-1, we need the code point.
> If we decide to make DSA MUST_NOT, then I'll mark it reserved and update the text.

Agreed. Using the "RESERVED" notation before was a bit iffy and with this new refactoring it has to be changed now. Just need to make sure it's mentioned explicitly that it's restricted.

draft-ietf-tls-tls13.md
```diff
@@ -2234,7 +2211,7 @@ The "extension_data" field of this extension contains a
        enum {
            anonymous_RESERVED(0),
            rsa(1),
-           dsa_RESERVED(2),
+           dsa(2),
            ecdsa(3),
            rsapss(4),
```
davegarrett: This can be collapsed to one line for the 4 now.

ekr: Yeah, I thought about that, but the script doesn't handle it well. Going to leave for now.

draft-ietf-tls-tls13.md
```diff
@@ -2919,11 +2886,13 @@ Structure of this message:
...
            rsa_sign(1), 
            dss_sign_RESERVED(2), 
-           rsa_fixed_dh(3), 
+           rsa_fixed_dh_RESERVED(3), 
            dss_fixed_dh_RESERVED(4),
            rsa_ephemeral_dh_RESERVED(5), 
            dss_ephemeral_dh_RESERVED(6),
-           fortezza_dms_RESERVED(20), (255)
+           fortezza_dms_RESERVED(20),
+           ecdsa_sign(64),
```
davegarrett: What about rsa_fixed_ecdh(65) & ecdsa_fixed_ecdh(66)?

ekr: The main point of having reserved fields here is to avoid duplicate assignment and we're going to be removing this entire type, so I don't see a lot of point in adding these.

PR #273, opened 2015-10-03
Don't use ServerConfiguration till it's verified. Fixed #229

### LINKED ISSUE #229: Require that ServerConfiguration not be acceptable till handshake finished
Hubert Karlo writes:
> > > Also, the record protection used for early handshake messages should be
> > > indicated.
> >
> > Can you expand on that?
>
> How does the client know what record protection algorithms are valid
> for 0RTT transmission for that server?

And how does the client know that the algorithms came from the server. We
should have a "client MUST wait for the full handshake to finish before
recording this information" or we will have a very nice cipher downgrade. Just
having it signed is likely not a good idea, as they may depend on ciphersuites
advertised by client.

ekr:


martinthomson:
LGTM

PR #274, opened 2015-10-03
Omit empty ClientKeyShare. Fixes #190

### LINKED ISSUE #190: Omit ClientKeyShare if it's empty

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -1428,10 +1428,6 @@ decrypt_error
   to correctly verify a signature or validate a Finished message.
   This alert is always fatal.
 
-export_restriction_RESERVED
-: This alert was used in some earlier versions of TLS. It MUST NOT
-  be sent by compliant implementations. This alert is always fatal.
```
davegarrett: Unrelated change from another PR.

ekr:
Holding off on this because Garrett pointed out that having no extension to elicit server state and than "appending" is weird.

davegarrett:
My objection is not strong. After rewrites, it doesn't stand out as much in the text. It's fine so long as it's explicit that zero offers is no extension and an empty vector is not permitted. (don't want both showing up) I'll have a line to that effect in the extension merging PR, so no objection to this PR as-is.

ekr:
I'm holding off on this. We have been looking at HelloRetryRequest and it looks like it would be easier to require the empty extension.

Issue #73, opened 2014-07-20
Retry on ClientHello


martinthomson:
MUST include all the original shares, plus the one the server selected. Otherwise we create a downgrade opportunity.

PR #275, opened 2015-10-04
revise HelloRetryRequest

davegarrett:
This is the commit with the HelloRetryRequest changes from PR #268 backported to the current master. It's mostly just editorial, with a few important fixes. There was one part with a double negative that was supposed to be a triple negative and made no sense. As I was interacting with this area in merging the DH key share methods together, I ended up rewriting this somewhat to make it easier to follow and fix the various issues. Aside from being more verbose and explicit about how to handle errors, this doesn't change any of the design in the current draft spec.

It also makes note of, but does not alter the current state of, issue #104.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2066,12 +2066,11 @@ extensions
 
 When this message will be sent:
 
-> Servers send this message in response to a ClientHello
-message when it was able to find an acceptable set of algorithms and
-groups that are mutually supported, but
-the client's ClientKeyShare did not contain an acceptable
-value. If it cannot find such a match, it will respond with a fatal
-"handshake_failure" alert.
+> Servers send this message in response to a ClientHello message when it
+was able to find an acceptable set of algorithms and groups that are
+mutually supported, but the client's ClientKeyShare did not contain an
+acceptable value. If it cannot find such a match, it will respond with a
+fatal "handshake_failure" alert.
 
 Structure of this message:
 
@@ -2089,35 +2088,40 @@ selected_group
 : The group which the client MUST use for its new ClientHello.
 {:br }
 
-The server_version, cipher_suite, and extensions fields have the
-same meanings as their corresponding values in the ServerHello. The
-server SHOULD send only the extensions necessary for the client to
-generate a correct ClientHello pair. As with a ServerHello, only
-extensions first offered by the client can appear in a HelloRetryRequest.
+The server_version, cipher_suite, and extensions fields have the same
+meanings as their corresponding values in the ServerHello. The server
+SHOULD send only the extensions necessary for the client to generate a
+correct ClientHello pair. As with a ServerHello, only extensions first
+offered by the client can appear in a HelloRetryRequest.
 
-Upon receipt of a HelloRetryRequest, the client MUST first verify
-that the selected_group was offered in its "supported_groups" extension and
-that no ClientKeyShareOffer for the selected_group was offered in its "client_key_share"
-extension. If either of these conditions are detected, then the client
-MUST abort the handshake with a fatal "handshake_failure"
-alert. Clients SHOULD also abort with an "handshake_failure" alert in response to any second
-HelloRetryRequest which is sent on the same connection (i.e.,
-where the ClientHello was itself in response to a HelloRetryRequest).
+Upon receipt of a HelloRetryRequest, the client MUST first verify that
+the selected_group was offered in its "supported_groups" extension and
+that no ClientKeyShareOffer for the selected_group was offered in its
+"client_key_share" extension. If either of these conditions are detected,
+then the client MUST abort the handshake with a fatal "handshake_failure"
+alert.
 
-If the HelloRetryRequest is verified as acceptable, the client MUST send a new
-ClientHello with an updated ClientKeyShare extension to the server. This ClientKeyShare MUST contain
-the exact previous set of offered keys with a new ClientKeyShareOffer value for the server's selected_group
-appended to the end of the list (or a new extension with exactly one value, if none was provided initially).
-All other fields in this new ClientHello MUST match the first attempt.
-Note that omitting any information previously included in the initial ClientHello
-introduces the risk of downgrade attack, as this retry exchange is unauthenticated.
+Clients SHOULD also abort with an "handshake_failure" alert in response
+to any second HelloRetryRequest which is sent on the same connection
+(i.e., where the ClientHello was itself in response to a
+HelloRetryRequest).
+
+If the HelloRetryRequest is verified as acceptable, the client MUST send
+a new ClientHello with an updated ClientKeyShare extension to the server.
+This ClientKeyShare MUST contain the exact previous set of offered keys
+with a new ClientKeyShareOffer value for the server's selected_group
+appended to the end of the list (or a new extension with exactly one
+value, if none was provided initially). All other fields in this new
+ClientHello MUST match the first attempt. Note that omitting any
+information previously included in the initial ClientHello introduces
+the risk of downgrade attack, as this retry exchange is unauthenticated.
 
 After sending the new ClientHello and receiving a new ServerHello,
 the client MUST verify that the same ProtocolVersion, CipherSuite, and
 NamedGroup specified in the server's HelloRetryRequest are selected in the
-server's new ServerHello/ServerKeyShare. If any of these values
-differ, the client MUST abort the connection with a fatal
-"handshake_failure" alert.
+server's new ServerHello/ServerKeyShare. If any of these values differ,
+the client MUST abort the connection with a fatal "handshake_failure"
+alert.
 
 [[OPEN ISSUE: https://github.com/tlswg/tls13-spec/issues/104]]
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2068,8 +2068,8 @@ When this message will be sent:
 
 > Servers send this message in response to a ClientHello message when it
 was able to find an acceptable set of algorithms and groups that are
-mutually supported, but the client's ClientKeyShare did not contain an
-acceptable value. If it cannot find such a match, it will respond with a
+mutually supported, but the client's ClientKeyShare offer did not contain an
+acceptable value. If it cannot find such a match, it MUST respond with a
 fatal "handshake_failure" alert.
 
 Structure of this message:
@@ -2117,8 +2117,8 @@ information previously included in the initial ClientHello introduces
 the risk of downgrade attack, as this retry exchange is unauthenticated.
 
 After sending the new ClientHello and receiving a new ServerHello,
-the client MUST verify that the same ProtocolVersion, CipherSuite, and
-NamedGroup specified in the server's HelloRetryRequest are selected in the
+the client MUST verify that the same server_version, cipher_suite, and
+selected_group specified in the server's HelloRetryRequest are selected in the
 server's new ServerHello/ServerKeyShare. If any of these values differ,
 the client MUST abort the connection with a fatal "handshake_failure"
 alert.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2070,7 +2070,7 @@ When this message will be sent:
 message when it was able to find an acceptable set of algorithms and
 groups that are mutually supported, but
 the client's ClientKeyShare did not contain an acceptable
-offer. If it cannot find such a match, it will respond with a
+value. If it cannot find such a match, it will respond with a fatal
 "handshake_failure" alert.
 
 Structure of this message:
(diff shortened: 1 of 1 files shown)


ekr:
There's a lot of reflow here and it's making it hard to tell what you actually changed in the text.
Are you able to separate out those changes?

davegarrett:
On Saturday, October 03, 2015 08:25:11 pm ekr wrote:
> There's a lot of reflow here and it's making it hard to tell what you actually changed in the text.
> Are you able to separate out those changes?

I pushed a commit to make the linebreaks similar, but the differ on GitHub isn't really smart enough for it to help much.
Top is a little less changed here:
https://github.com/tlswg/tls13-spec/pull/275/files
The "rich diff" doesn't care and says it's all different:
https://github.com/tlswg/tls13-spec/pull/275/files?short_path=9d84740#diff-9d84740dcc569a0a5a359d0fba461a05

To break it down manually:

The "Upon receipt of a HelloRetryRequest[...]" is rewritten. It previously said:
> MUST first verify that the "selected_group" field does not identify a group which was not in the original ClientHello. If it was present, then the client MUST abort the handshake with a fatal "handshake_failure" alert.
It's saying MUST verify that the field does not (1) identify a group not (2) in first ClientHello and if it was present (?) then MUST abort. With (1)+(2) it's a double negative, and then (?) should be a third negative to make sense if we're talking about a group in the supported_groups; it's just a confusing double negative if we're talking about client_key_share. Either way, it's not clear. I changed this to note both places by extension name, and that we're retrying IFF the client had it in the supported_groups but not the client_key_share. If it's not supported or a key was offered, then error.

I then make the infinite loop check its own paragraph, and change it to use the more specific "unexpected_message" alert for the unexpected repeat retry request.

The next paragraph has a change saying the retry MUST contain exactly one share for the group the server requested and that the rest of the hello MUST remain the same (a requirement mentioned elsewhere).

The next bit fills in the step of the server replying to the new ClientHello.

The next bit reorders the wording to explicitly state what was selected. It also lists the version field for completeness and notes the error by name that would be here: illegal_parameter (again, we could just do a generic handshake_failure, but then nobody knows what to fix).

Lastly is just a note of issue #104. For safety, I'm assuming the transcript for the hash is everything. The current draft text is very ambiguous as to what the handshake hash is on a retry.

draft-ietf-tls-tls13.md
```diff
@@ -2069,8 +2069,8 @@ When this message will be sent:
 > Servers send this message in response to a ClientHello
 message when it was able to find an acceptable set of algorithms and
 groups that are mutually supported, but
-the client's ClientKeyShare did not contain an acceptable
-offer. If it cannot find such a match, it will respond with a
+the client's ClientKeyShareOffers did not contain an acceptable
```
ekr: I don't think this is an improvement. We generally refer to the extension, not to the
inner structs.

draft-ietf-tls-tls13.md
```diff
@@ -2069,8 +2069,8 @@ When this message will be sent:
...
+value. If it cannot find such a match, it will instead respond with a fatal
```
ekr: instead seems superfluous here.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2089,42 @@ selected_group
...
-that the "selected_group" field does not identify a group which
-was not in the original ClientHello. If it was present, then
-the client MUST abort the handshake with a fatal "handshake_failure"
-alert. Clients SHOULD also abort with "handshake_failure" in response to any second
-HelloRetryRequest which was sent in the same connection (i.e.,
+that the selected_group was offered in its "supported_groups" extension and
+that no ClientKeyShareOffer for the selected_group was offered in its "client_key_share"
+extension. If either of these conditions are detected, then the client
+MUST abort the handshake with a fatal "handshake_failure"
+alert. Clients SHOULD also abort with an "unexpected_message" alert in response to any second
```
ekr: I would prefer "handshake_failure" here.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2089,42 @@ selected_group
...
 
-Otherwise, the client MUST send a ClientHello with a new
-ClientKeyShare extension to the server. The ClientKeyShare MUST append
-a new ClientKeyShareOffer which is consistent with the
-"selected_group" field to the groups in the original ClientKeyShare.
+If the HelloRetryRequest is verified as acceptable, the client MUST send a new
+ClientHello with a new ClientKeyShare extension to the server. This ClientKeyShare MUST contain
+exactly one ClientKeyShareOffer value which is consistent with the
+selected_group field provided by the server in its HelloRetryRequest.
+All other fields in this new ClientHello MUST match the first attempt.
```
ekr: This changes the meaning in an important way. The client must *add* a ClientKeyShareOffer
that reflects that group, not replace all its ClientKeyShareOffers with the new one. Moroever, it must append it because it is lower priority.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2089,42 @@ selected_group
...
 
-Upon re-sending the ClientHello and receiving the
-server's ServerHello/ServerKeyShare, the client MUST verify that
-the selected CipherSuite and NamedGroup match that supplied in
-the HelloRetryRequest.
+If the new ClientHello is acceptable, the server will continue the
+handshake by replying with a new ServerHello, after which point
+the handshake follows normal procedure. (see {{incorrect-dhe-share}})
+
```
ekr: This seems redundant. The intention is that the server is to be stateless.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2089,42 @@ selected_group
...
+Upon receipt of a new ServerHello in response to a retried ClientHello,
```
ekr: I prefer the phrasing in the original.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2089,42 @@ selected_group
...
+the client MUST verify that the same ProtocolVersion, CipherSuite, and
+NamedGroup specified in the server's HelloRetryRequest are offered in the
+server's new ServerHello/ServerKeyShare. If any of these values
```
ekr: They're not offered. They're selected.

draft-ietf-tls-tls13.md
davegarrett: To borrow your phrase: pilot error.  I mixed the structs up when pulling this out, as the names are all very similar. Will fix.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2089,42 @@ selected_group
...
+differ, the client MUST abort the connection with a fatal
+"illegal_parameter" alert.
```
ekr: This doesn't seem like "illegal_parameter" to me. "handshake_failure"

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2089,42 @@ selected_group
...
+
+Note that the full session transcript starts at the initial ClientHello,
+and thus the handshake hash covers all exchanged messages, including
+both the first and second ClientHello messages. [{{the-handshake-hash}}]
+[[OPEN ISSUE: https://github.com/tlswg/tls13-spec/issues/104]]
```
ekr: I don't expect us to resolve this issue this way. Please remove this text.

draft-ietf-tls-tls13.md
```diff
@@ -3464,8 +3478,8 @@ each class of traffic keys:
                             )
 
 handshake_messages
-: All handshake messages sent or
-  received, starting at ClientHello up to the present time, with the
+: All handshake messages sent or received, starting at the initial
+  ClientHello up to the present time, with the
```
ekr: See above

draft-ietf-tls-tls13.md
davegarrett: Yes, I'm aware. There's no point in wasting bandwidth resending the rejected keys. Also, we'll be changing the key offer to allow clients to omit the extension, rather than leave it empty. Stating it this way makes it always the same: new extension with the selected key offer. It can't append if it was omitted; it would need special casing in that case.

draft-ietf-tls-tls13.md
davegarrett: Redundancy is not always bad. The previous text skipped a step; this spells it out explicitly. Stateless or not, this is what the handshake process is here. This text doesn't state that the server must keep a state here, just that logically the handshake continues (as is the case in your flow diagram). On a related note, these two separate sections should probably be merged.

Also note that the current text is incomplete with regard to how the handshake hash will deal with this, and you've noted that I should cut a note below. That text and this phrasing come from trying not to explicitly write something with an easy downgrade attack.

draft-ietf-tls-tls13.md
ekr: This supersetting design was the result of group consensus in Toronto.

See: http://www.ietf.org/proceedings/interim/2014/07/20/tls/minutes/minutes-interim-2014-tls-2
and https://github.com/tlswg/tls13-spec/issues/73

Please don't intermix technical changes with editorial ones.

davegarrett: On Saturday, October 03, 2015 10:40:36 pm ekr wrote:
> This supersetting design was the result of group consensus in Toronto.
> 
> See: http://www.ietf.org/proceedings/interim/2014/07/20/tls/minutes/minutes-interim-2014-tls-2
> and https://github.com/tlswg/tls13-spec/issues/73

Ah. The previous keys are kept so that this can be done statelessly without the easy downgrade attack. Sorry, I missed that bit and was thinking #104 was to cover the whole issue. I'll change it back. Sorry for confusing things.

draft-ietf-tls-tls13.md
ekr: I don't think redundancy is helpful here.

The text is incomplete because this is still an open issue. I'd rather not important text that resolves that until the issue is resolved, especially as I currently expect us to resolve it in the opposite direction than the way you have done here. If you have an analysis that demonstrates a downgrade attack with the present design (i.e., with appending the group rather than replacing it) when the handshake hash is not continued, that would be useful for that discussion.

ekr: I think we crossed wires here with your note above. However, if you think of any attacks on the non-hash-continuation design, that would be very helpful. We're currently trying to get a tight analysis of that, so any thinking you have here would be most welcome.

davegarrett: On Saturday, October 03, 2015 10:52:46 pm ekr wrote:
> I think we crossed wires here with your note above. However, if you think of any attacks on the non-hash-continuation design, that would be very helpful. We're currently trying to get a tight analysis of that, so any thinking you have here would be most welcome.

Essentially, I was writing from a logical flow perspective without thinking about how to implement it statelessly. I missed that the keys resend was intentional to avoid the simple attack in the stateless case (issue #73, opened & closed a year ago) and I attributed #104 to covering the whole issue. Obvious in hindsight. I just thought more was incomplete here than was actually the case.

I'll hand-wave my mistake into saying that I think writing out things explicitly here will help avoid confusions and mistakes with regard to reading & implementing this. ;)

In this case, I'd actually like to just write that out explicitly here to make sure it's clear and clients don't skimp here to save bandwidth.

Will push changes shortly...

davegarrett:
@ekr: Addressed all comments and squashed to two commits: one with attempt to minimize wrapping changes and a followup to just fix the wrapping for the whole section. (wrapped to 75 char lines) Sorry; took longer than that should've. Was also fighting with git a bit there.

draft-ietf-tls-tls13.md
```diff
@@ -2066,12 +2066,11 @@ extensions
...
-> Servers send this message in response to a ClientHello
-message when it was able to find an acceptable set of algorithms and
-groups that are mutually supported, but
-the client's ClientKeyShare did not contain an acceptable
-offer. If it cannot find such a match, it will respond with a
-"handshake_failure" alert.
+> Servers send this message in response to a ClientHello message when it
+was able to find an acceptable set of algorithms and groups that are
+mutually supported, but the client's ClientKeyShare did not contain an
+acceptable value. If it cannot find such a match, it will respond with a
```
ekr: I prefer "offer" here. I could also live with ClientKeyShare offer.

davegarrett:
@ekr: fixed

I prefer "ClientKeyShare offer" as I generally think specs are easier to use when you cite names directly, thus making it easy to search for them in the rest of the document. Also changed the will to a MUST.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2088,42 @@ selected_group
...
-Upon re-sending the ClientHello and receiving the
-server's ServerHello/ServerKeyShare, the client MUST verify that
-the selected CipherSuite and NamedGroup match that supplied in
-the HelloRetryRequest.
+The server_version, cipher_suite, and extensions fields have the same
+meanings as their corresponding values in the ServerHello. The server
+SHOULD send only the extensions necessary for the client to generate a
+correct ClientHello pair. As with a ServerHello, only extensions first
+offered by the client can appear in a HelloRetryRequest.
+
```
ekr: I think the way to deal with this is to have a list. I'm working on that in another PR.

draft-ietf-tls-tls13.md
```diff
@@ -2089,28 +2088,42 @@ selected_group
...
+ClientHello MUST match the first attempt. Note that omitting any
+information previously included in the initial ClientHello introduces
+the risk of downgrade attack, as this retry exchange is unauthenticated.
+
+After sending the new ClientHello and receiving a new ServerHello,
+the client MUST verify that the same ProtocolVersion, CipherSuite, and
+NamedGroup specified in the server's HelloRetryRequest are selected in the
+server's new ServerHello/ServerKeyShare. If any of these values differ,
+the client MUST abort the connection with a fatal "handshake_failure"
+alert.
```
ekr: This seems to be inconsistent about how to name these fields. "server_version" above and ProtocolVersion here. Yes, this is preexisting, but we should make them match

davegarrett: On Monday, October 05, 2015 02:37:42 pm ekr wrote:
> > +After sending the new ClientHello and receiving a new ServerHello,
> > +the client MUST verify that the same ProtocolVersion, CipherSuite, and
> > +NamedGroup specified in the server's HelloRetryRequest are selected in the
> > +server's new ServerHello/ServerKeyShare. If any of these values differ,
> > +the client MUST abort the connection with a fatal "handshake_failure"
> > +alert.
> 
> This seems to be inconsistent about how to name these fields. "server_version" above and ProtocolVersion here. Yes, this is preexisting, but we should make them match

Sure. I amended the last commit for changes for your comments to change this too & repushed.

davegarrett:
@ekr: All comments have been fixed and this merges with master cleanly (I just tested). Is there anything else you want me to change here, or is this ready?

ekr:
@martinthomson and I have been looking at some restructuring of HRR to make the handshake cleaner. I will be posting something this week, but a holding off on this PR till then.

davegarrett:
This was mostly dealt with in https://github.com/tlswg/tls13-spec/commit/74c4c7da1bf5bcea4f75cfc38c1aedb406098d3d#diff-9d84740dcc569a0a5a359d0fba461a05L2097 . I'm going to close this, then. I'll file a new PR with any little stuff.

davegarrett:
Followup -> PR #297

PR #284, opened 2015-10-09
Anti downgrade

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -1907,8 +1909,27 @@ random_bytes
   See {{implementation-notes}} for additional information.
 {:br }
 
+TLS 1.3 server implementations which respond to a ClientHello with a
+client_version indicating TLS 1.2 or below MUST set the first eight
+bytes of its Random value to the bytes 44 4F 57 4E 47 52 44 01.
```
davegarrett: Could you please put each of these two byte strings on their own lines with spaces around? (e.g. same formatting as end of "Negotiated Groups" section) Easier to read, that way.

seanturner:
Chairs have judged that there is consensus to merge, see:
https://mailarchive.ietf.org/arch/msg/tls/UfoOWPuJXCL1MyhJ4yAYpg3r53E

PR #288, opened 2015-10-12
Explicitly specify SSLv2/SSLv3

yuhong:
Explicitly specify SSL version 2 and SSL version 3.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3846,7 +3846,7 @@ TLS protocol issues:
 -  Do you ignore the TLS record layer version number in all TLS
   records? (see {{backward-compatibility}})
 
--  Have you ensured that all support for SSLv2 (other than SSLv2-style ClientHello processing), SSLv3, RC4, EXPORT ciphers, and
+-  Have you ensured that all support for SSL 2.0 (other than SSL 2.0-style ClientHello processing), SSL 3.0, RC4, EXPORT ciphers, and
   MD5 (via the Signature Algorithms extension) is completely removed from
   all possible configurations that support TLS 1.3 or later, and that
   attempts to use these obsolete capabilities fail correctly?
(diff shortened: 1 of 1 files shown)


davegarrett:
I don't think this is a helpful change. That list is long enough already.

"SSL" necessarily includes SSL 1-3. Dropping SSL2 compatible CLIENT-HELLO support is recommended, so it doesn't need singling out separately as permitted. Also, supporting it is not support for SSL2; it's an obsolete compatibility shim for TLS, so that's not even covered here. Implementations can support both TLS & SSL2 CLIENT-HELLO, but not SSL, just fine.

yuhong:
"SSL" is often used to refer to all SSL/TLS versions.

davegarrett:
And that practice needs to stop.

If this was anywhere other than directly in the TLS specification itself, I might agree with you with a need for clarification. However, it's clearly not telling people to make sure not to implement itself. ;)

davegarrett:
If @ekr agrees with enumerating it to "SSL 2.0, SSL 3.0", instead of simply "SSL", I don't think it's needed but I don't object.

The SSL2 compatible CLIENT-HELLO parenthetical, however, is not relevant here as that's a shim for TLS, not SSL2 support itself, and I'd like to keep mention of it as banished as possible. Implementations are permitted to drop support for it, but it's only a recommendation.

Note that we previously had a mix of "SSL3", "SSLv3", and "SSL 3.0" all scattered about, which made things a bit inconsistent and annoying to search the document for. I normalized all mentions of versions to use the "SSL 3.0" style version notation a while back, so please use this notation.

ekr:
I agree with @davegarrett, this isn't really making things clearer.

yuhong:
Also, TLS 1.0 was far less different from SSL 3.0 than SSL 3.0 was against SSL 2.0.

PR #290, opened 2015-10-14
Removing certificate_types and adding certificate_extensions

Andrei-Popov:
This PR includes CertificateRequest changes discussed at the TLS WG Interim:
1. Removes certificate_types, which are no longer needed.
2. Adds client cert selection by certificate extension values. This helps make CertificateRequest more specific and eliminate the need for the confusing "Choose a certificate" UI.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2357,10 +2357,10 @@ parameters for zero or more non-PSK key establishment methods (currently
 DHE or ECDHE).
 
 Clients which offer one or more (EC)DHE cipher suites 
-SHOULD send at least one supported ClientKeyShareOffer value and
+MUST send at least one supported ClientKeyShareOffer value and
 servers MUST NOT negotiate any of these cipher suites unless a supported
 value was provided.
-It is explicitly permitted for a client to omit this extension in order
+It is explicitly permitted for a client to send an empty extension in order
 to request a group from the server, at the cost of an additional round trip.
 (see {{hello-retry-request}})
 If this extension is not provided in a retried ClientHello, the server
@@ -2860,24 +2860,11 @@ When this message will be sent:
 
 > A non-anonymous server can optionally request a certificate from the client,
 if appropriate for the selected cipher suite. This message, if sent, will
-immediately follow the server's Certificate message. The server may also send
-this message at any time after the handshake is complete.
+immediately follow the server's Certificate message.
 
 Structure of this message:
 
 %%% Authentication Messages
-       enum {
-           rsa_sign(1), 
-           dss_sign_RESERVED(2), 
-           rsa_fixed_dh_RESERVED(3), 
-           dss_fixed_dh_RESERVED(4),
-           rsa_ephemeral_dh_RESERVED(5), 
-           dss_ephemeral_dh_RESERVED(6),
-           fortezza_dms_RESERVED(20),
-           ecdsa_sign(64),
-           (255)
-       } ClientCertificateType;
-
        opaque DistinguishedName<1..2^16-1>;
 
        struct {
@@ -2886,23 +2873,18 @@ Structure of this message:
        } CertificateExtension;
 
        struct {
-           ClientCertificateType certificate_types<1..2^8-1>;
            SignatureAndHashAlgorithm
              supported_signature_algorithms<2..2^16-2>;
            DistinguishedName certificate_authorities<0..2^16-1>;
            CertificateExtension certificate_extensions<0..2^16-1>;
        } CertificateRequest;
 
-certificate_types
-: A list of the types of certificate types that the client may
-  offer.
-
-          rsa_sign        a certificate containing an RSA key
-          rsa_fixed_dh    a certificate containing a static DH key.
-
 supported_signature_algorithms
 : A list of the hash/signature algorithm pairs that the server is
-  able to verify, listed in descending order of preference.
+  able to verify, listed in descending order of preference. Any 
+  certificates provided by the client MUST be signed using a
+  hash/signature algorithm pair found in 
+  supported_signature_algorithms.
 
 certificate_authorities
 : A list of the distinguished names {{X501}} of acceptable
@@ -2911,8 +2893,9 @@ certificate_authorities
   root CA or for a subordinate CA; thus, this message can be used to
   describe known roots as well as a desired authorization space.  If
   the certificate_authorities list is empty, then the client MAY
-  send any certificate of the appropriate ClientCertificateType,
-  unless there is some external arrangement to the contrary.
+  send any certificate that meets the rest of the selection criteria 
+  in the CertificateRequest, unless there is some external arrangement 
+  to the contrary.
 
 certificate_extensions
 : A list of certificate extension OIDs [RFC5280] with their allowed
@@ -2925,7 +2908,7 @@ certificate_extensions
   client certificate (but the certificate MAY have other values as 
   well). However, the client MUST ignore and skip any unrecognized 
   certificate extension OIDs. If the client has ignored some of the 
-  required certificate extension OIDs, and supplied a  certificate 
+  required certificate extension OIDs, and supplied a certificate 
   that does not satisfy the request, the server MAY at its discretion 
   either continue the session without client authentication, or 
   terminate the session with a fatal unsupported_certificate alert.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2878,37 +2878,26 @@ immediately follow the server's Certificate message.
...
-          rsa_sign        a certificate containing an RSA key
-          rsa_fixed_dh    a certificate containing a static DH key.
-
 supported_signature_algorithms
 : A list of the hash/signature algorithm pairs that the server is
-  able to verify, listed in descending order of preference.
+  able to verify, listed in descending order of preference. Any 
+  certificates provided by the client MUST be signed using a
+  hash/signature algorithm pair found in 
+  supported_signature_algorithms.
```
ekr: We have relaxed this restriction for the server and made it a strong SHOULD. Should we do the same here?

draft-ietf-tls-tls13.md
```diff
@@ -2917,28 +2906,49 @@ certificate_authorities
...
+: A list of certificate extension OIDs [RFC5280] with their allowed
+  values, represented in DER-encoded format. Some certificate
+  extension OIDs allow multiple values (e.g. Extended Key Usage).
+  If the server has included a non-empty certificate_extensions list,
+  the client certificate MUST contain all of the specified extension
+  OIDs that the client recognizes. For each extension OID recognized
+  by the client, all of the specified values MUST be present in the
+  client certificate (but the certificate MAY have other values as 
+  well). However, the client MUST ignore and skip any unrecognized 
+  certificate extension OIDs. If the client has ignored some of the
```
ekr: Perhaps "ignore the contents of"?

draft-ietf-tls-tls13.md
```diff
@@ -2917,28 +2906,49 @@ certificate_authorities
...
+  corresponding value types. Depending on the type, matching 
+  certificate extension values are not necessarily bitwise-equal. It 
+  is expected that TLS implementations will rely on their PKI 
+  libraries to perform certificate selection using certificate 
+  extension OIDs.
+
+  This document defines matching rules for two standard certificate 
+  extensions defined in [RFC5280]:
+
+  - Key Usage extension in a certificate matches the request when all
```
ekr: Nit: I would put a "The" in front of this sentence and the one below.

draft-ietf-tls-tls13.md
```diff
@@ -3186,6 +3196,13 @@ In particular:
   that this relaxes the constraints on certificate-signing
   algorithms found in prior versions of TLS.
 
+- If the certificate_extensions list in the certificate request message
```
ekr: You need to remove the certificate types definitions above.

draft-ietf-tls-tls13.md
```diff
@@ -3186,6 +3196,13 @@ In particular:
...
+  was non-empty, the end-entity certificate MUST contain all of the
+  specified extension OIDs that the client recognizes. For each extension
+  OID recognized by the client, all of the specified values MUST be present
+  in the client certificate. However, the client MUST ignore and skip any
+  unrecognized certificate extension OIDs.
+
```
ekr: This text seems duplicative of the text describing the extensions. I would suggest just poingint to that section.

draft-ietf-tls-tls13.md
Andrei-Popov: I think algorithm negotiation should be definitive, both for the server and the client. What was the reason for relaxing this requirement for the server?

ekr: The reasoning was that you have some limited number of certificates and that it's better to provide something than fail the handshake.

I don't care much either way, but I wanted to raise it.

draft-ietf-tls-tls13.md
Andrei-Popov: Sure, if this is clearer. The idea is that the client MUST ignore OIDs in the CertificateRequest that the client does not recognize, MUST not try to match them, and MUST skip to the next OID.

draft-ietf-tls-tls13.md
Andrei-Popov: Totally.

draft-ietf-tls-tls13.md
ekr: Yeah, that's what I got from it.  I don't think it's a big deal either way

draft-ietf-tls-tls13.md
davegarrett: On Wednesday, October 14, 2015 05:13:52 pm Andrei-Popov wrote:
> I think algorithm negotiation should be definitive, both for the server and the client. What was the reason for relaxing this requirement for the server?

On Wednesday, October 14, 2015 05:16:36 pm ekr wrote:
> The reasoning was that you have some limited number of certificates and that it's better to provide something than fail the handshake.

The goal was to not fail so the client could make the final decision to abort or not, rather than the server, as the server doesn't always have enough information to make that decision properly.

Doing this in the opposite direction as well could make sense, but giving the client the option to abort rather then send its certificates might be useful in other cases.

draft-ietf-tls-tls13.md
```diff
@@ -3554,13 +3571,6 @@ This document uses several registries that were originally created in
 {{RFC4346}}. IANA has updated these to reference this document. The registries
 and their allocation policies (unchanged from {{RFC4346}}) are listed below.
 
--  TLS ClientCertificateType Identifiers Registry: Future values in
-  the range 0-63 (decimal) inclusive are assigned via Standards
-  Action {{RFC2434}}.  Values in the range 64-223 (decimal) inclusive
-  are assigned via Specification Required {{RFC2434}}.  Values from
-  224-255 (decimal) inclusive are reserved for Private Use
-  {{RFC2434}}.
-
```
davegarrett: The IANA Considerations section should probably declare retirement of the ClientCertificateType registry.

Andrei-Popov: Can we retire a registry that is still in use? It’s just the TLS 1.3 spec that no longer uses it (assuming the PR gets integrated). I’m inclined to let the editor make this change if he thinks it’s needed.

draft-ietf-tls-tls13.md
Andrei-Popov: I think it was not a good change for the server. The client explicitly indicated lack of support for the algorithms in the server’s certificate(s), and sending unsupported certificate(s) is an attack on the client. Hopefully, the client will prevent this attack, but the protocol should not put a compliant server in the position of an attacker☺.

For client auth, the argument of “let’s not break the handshake” makes even less sense: a client including an empty certificate does not necessarily fail the handshake. The client simply says “I can’t satisfy this CertificateRequest”.

draft-ietf-tls-tls13.md
ekr: I tend to agree with Andrei. One might imagine having people register new values for TLS 1.2

draft-ietf-tls-tls13.md
ekr: I can live with Andrei's proposed text.

draft-ietf-tls-tls13.md
Andrei-Popov: Done.

draft-ietf-tls-tls13.md
davegarrett: On Wednesday, October 14, 2015 05:51:07 pm Andrei-Popov wrote:
> I think it was not a good change for the server. The client explicitly indicated lack of support for the algorithms in the server’s certificate(s), and sending unsupported certificate(s) is an attack on the client.

It's not an attack if it knows it's a possibility because it's part of the spec. One of the prime reasons to make the change was to handle things like DANE & opportunistic encryption better, as there's no way for the client to sufficiently inform the server of what it supports (if it cares at all). See the mailing list discussions on this topic, as we're getting off-topic now.

draft-ietf-tls-tls13.md
Andrei-Popov: Yes, that's how I feel too. Certainly not going to object if folks find "ignore the contents of" clearer.

draft-ietf-tls-tls13.md
Andrei-Popov: Off-topic? Somewhat true.
Not an attack? I respectfully disagree; will keep thwarting:).

ekr: The current version of TLS includes MUST language here, so I suggest that (pending people raising other concerns in the next few days) we land this PR and then we can have a separate issue about whether to relax this requirement.

Andrei-Popov: Sounds like a plan!

PR #293, opened 2015-10-15
State the required salt length for RSA-PSS signatures

ttaubert:


draft-ietf-tls-tls13.md
```diff
@@ -753,7 +753,8 @@ In RSA signing, the opaque vector contains the signature generated
 using the RSASSA-PSS signature scheme defined in {{RFC3447}} with
 MGF1. The digest used in the mask generation function MUST be the same
 as the digest which is being signed (i.e., what appears in
-algorithm.signature). Note that previous versions of TLS used
+algorithm.signature). The length of the salt MUST be equal to the octet
+length of the digest output. Note that previous versions of TLS used
```
briansmith: I think this is a good change.

NIT: You use "length" to refer to the length of the salt but "octet length" to refer to the length of the digest output. I suggest you be consistent there, and be consistent with whatever is done in the rest of the document.

PR #296, opened 2015-10-17
group & key share recommendations

davegarrett:
These are a revision of the recommendations I had in earlier drafts of my PRs. They're now neutral as to group priority, as you said that had not been sufficiently discussed & decided.

Note that one important issue is addressed here: how to reconcile the priorities of the "supported_groups" and "key_share" extensions. Both have priorities, but those could conflict leading to no clear answer as to what the indicated priority should be. I've fixed that with a simple "SHOULD" to have them just match up. If they still don't match up, it's ambiguous; implementers can decide for themselves how to handle that, which I think is fine.

This also adds a "MUST NOT" offer shares for unsupported groups.

There's also a "MUST NOT" use ffdhe2048 with TLS 1.3+, as it's in there primarily to increase adoption of the extension with older servers that will be using TLS 1.2 and it is relatively weak.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2496,6 +2496,23 @@ different curves for (i) the ECDSA or EdDSA key in its certificate,
 and (ii) the ephemeral ECDH key in its "key_share" extension.  The server
 must consider the supported groups in both cases.
 
+Servers SHOULD avoid negotiation of groups for key exchange that have
+an estimated security level significantly higher or lower than that of
+the negotiated symmetric bulk cipher key size. The RECOMMENDED groups
+are as follows:
+
+  * 128-bit level:
+    X25519, secp256r1, ffdhe3072, ffdhe4096
+
+  * 192/256-bit level:
+    X448, secp384r1, secp521r1, ffdhe8192
+
+The estimated symmetric-equivalent strength of ffdhe2048 is only 103-bit,
+and thus MUST NOT be negotiated for TLS 1.3 or later, however it MAY
+be offered for use with prior TLS versions supporting this extension.
+
+[[TODO: Priority recommendations]]
+
 [[TODO: IANA Considerations.]]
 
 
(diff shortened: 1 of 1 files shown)


davegarrett:
Rebased & fixed conflict with recent changes.

draft-ietf-tls-tls13.md
```diff
@@ -2496,6 +2496,22 @@ different curves for (i) the ECDSA or EdDSA key in its certificate,
...
+
+  * 128-bit level:
+    X25519, secp256r1, ffdhe3072, ffdhe4096
+
+  * 192/256-bit level:
+    X448, secp384r1, secp521r1, ffdhe8192
+
+The estimated symmetric-equivalent strength of ffdhe2048 is only 103-bit,
+and thus MUST NOT be negotiated for TLS 1.3 or later, however it MAY
+be offered for use with prior TLS versions supporting this extension.
```
ekr: I think we would need consensus for this.

draft-ietf-tls-tls13.md
```diff
@@ -2583,8 +2601,23 @@ the "supported_groups" extension, then the server MUST reply with a
...
+In general, it is RECOMMENDED to send at least one group of each security
+level corresponding to the key lengths of the offered bulk ciphers (see
+{{negotiated-groups}}), preferably the highest priority supported groups
+applicable to the highest priority supported cipher suites.
+Groups stated as mandatory to implement for ciphers offered by the client
+are RECOMMENDED to be sent in most, if not all, offers.
+(see {{mti-cipher-suites}}) Clients with prior expectations of what groups
+are likely to be accepted by the server will generally want to include
+those groups in addition to, or instead of, this general set of
+recommendations.
```
ekr: I'm not sure I think this is the right recommendation, since it would imply
(for instance) that one should send x448 when you are offering ChaCha/Poly (because it has a 256-bit key). That doesn't seem necessary. My sense would be to recommend sending one or two groups plus "prior expectations"

draft-ietf-tls-tls13.md
davegarrett: If I remember correctly, this sentence was added due to a statement made by the FFDHE draft author about 2048. I can bring this to the list after a rewrite, at some point.

ekr: Oh, it's definitely @dkg's opinion. But I'm not sure it's consensus.

draft-ietf-tls-tls13.md
davegarrett: It probably should, though I do agree with your point that X25519 is likely to be considered fine in this case.

davegarrett:
This PR is obsolete and not useful to be open anymore. (see also mailing list discussion related to parts here)

PR #297, opened 2015-10-17
helloretryrequest fixups

davegarrett:
@ekr: You addressed the main ambiguity in https://github.com/tlswg/tls13-spec/commit/74c4c7da1bf5bcea4f75cfc38c1aedb406098d3d#diff-9d84740dcc569a0a5a359d0fba461a05L2097 . This fixes a few remaining issues. An itemized list:
* "handshake_failure" alert is fatal
* note it must be appended, not just "used", in the selected_group description (just using instead of appending creates a downgrade attack due to lack of authentication for this message, as discussed previously)
* clean up styling of variable names; the dominant style uses no quotes around variable names, but quotes around enum values (granted, this may not be the ideal style, but it is more or less consistently the case elsewhere in the document)
* note that the same rule as ServerHello applies: no extensions the client did not offer first
* update text to deal with possibility of omitted extension
* minor wording simplification
* specify error to use if "MUST" fails (could be a "SHOULD" error, here)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2419,14 +2419,16 @@ must consider the supported groups in both cases.
 The "key_share" extension contains the endpoint's cryptographic parameters
 for non-PSK key establishment methods (currently DHE or ECDHE).
 
-Clients which offer one or more (EC)DHE cipher suites
-MUST send at least one supported KeyShare value and
-servers MUST NOT negotiate any of these cipher suites unless a supported
+Clients which offer one or more (EC)DHE cipher suites MUST send this
+extension and SHOULD send at least one supported KeyShareEntry value.
+Servers MUST NOT negotiate any of these cipher suites unless a supported
 value was provided.
-If this extension is not provided in a ServerHello or retried ClientHello,
+If this extension is not provided in a ServerHello or ClientHello,
 and the peer is offering (EC)DHE cipher suites, then the endpoint MUST close
 the connection with a fatal "missing_extension" alert.
 (see {{mti-extensions}})
+Clients MAY send an empty client_shares vector in order to request
+group selection from the server at the cost of an additional round trip.
 
 %%% Key Exchange Messages
        struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2087,18 +2087,21 @@ Structure of this message:
 [[OPEN ISSUE: Merge in DTLS Cookies?]]
 
 selected_group
-: The group which the client MUST use for its new ClientHello.
+: The group which the client MUST append to its KeyShare list in its
+  new ClientHello.
```
ekr: It says "appended" below.

draft-ietf-tls-tls13.md
```diff
@@ -2107,15 +2110,16 @@ to any second HelloRetryRequest which was sent in the same connection
...
 HelloRetryRequest).
 
-Otherwise, the client MUST send a ClientHello with a new KeyShare
-extension to the server. The client MUST append a new KeyShareEntry
-list which is consistent with the "selected_group" field to the groups
-in its original KeyShare.
+Otherwise, the client MUST send a ClientHello with an updated KeyShare
+extension to the server, or a new extension if it was previously omitted.
+The client MUST append a new KeyShareEntry for the selected_group
+to the groups in its original KeyShare list.
```
ekr: selected_group is a field so "for" seems weird. If you prefer "for" then you need to say "for the group indicated in selected group"

davegarrett: On Saturday, October 17, 2015 04:46:27 pm ekr wrote:
> selected_group is a field so "for" seems weird. If you prefer "for" then you need to say "for the group indicated in selected group"

Makes sense; will change.

davegarrett:
@ekr: Repushed commit with changes for comments.
* Changed description to not say used/amended, but rather describe what it actually represents directly: the group the server intends to negotiate and is requesting a retry for. (we don't need duplicated MUSTs here, and stating its actual meaning more explicitly is preferable)
* Changed text wrt selected group as suggested.

davegarrett:
Rebased on top of PR #349 (which still merges cleanly onto master, as of this moment) to avoid flip-flopping again on the omit vs. empty KeyShare for retry request.

Issue #309, opened 2015-10-17
CELLOS: Resumption & PSK 1


seanturner:
Note that the client supplies a ClientKeyShare to the server as well, which allows the server to decline resumption and fall back to a full handshake." Here PSK is ignored.

Solution proposed: Replace this sentence by "Note that the client supplies a ClientKeyShare to the server as well, which allows PSK + (EC)DHE or the server to decline resumption and fall back to a full handshake"

ekr:
PR welcome.

ekr:
Already fixed.

PR #312, opened 2015-10-17
mention (EC)DHE+PSK (issue #309)

davegarrett:
Quick PR for proposed change for issue #309.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1828,8 +1828,8 @@ Subsequent Handshake:
 As the server is authenticating via a PSK, it does not
 send a Certificate or a CertificateVerify. PSK-based resumption cannot
 be used to provide a new ServerConfiguration. Note that the client
-supplies a KeyShare to the server as well, which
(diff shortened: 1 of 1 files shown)


ekr:
Actually, looking at this again, this proposal is not really what's wanted here.

The natural case here is that the client will offer only:
Sig + ECDHE (full handshake) and PSK (resumption)

davegarrett:
I thought we were using (EC)DHE+PSK for resumption? I mentioned previously that I think plain PSK for resumption is appropriate, but I vaguely remember someone saying we were going to use PFS+PSK here anyway. I don't see a reason to not support it, though.

ekr:
The client can use either plain PSK or (EC)DHE+PSK for resumption, at its discretion (this is one of the advantages of a more decoupled design). However, the natural counterpart is pure PSK.

davegarrett:
Well, if you think the current text is preferred, I have no objection to just closing this and the issue.

The PSK resumption story really does need to be clarified as to explicitly how offer and negotiate which cipher suites, by the way.

ekr:
I don't think this is what's needed.

PR #313, opened 2015-10-18
Renaming section to be generic and making description of msg agnostic.

seanturner:
It's always bugged me that there's no section title "client finished" just one called "server finished".  So, how about we make this one applicable to both by a title change and some tweaks to the initial description.


ekr:
I've got a major rewrite of this section pending and it includes a change like this.

PR #316, opened 2015-10-21
Wip client auth revision

ekr:
This is still a WIP but is posted so that people can review.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -326,6 +326,8 @@ draft-11
 
 - Update IANA considerations section and relax some of the policies.
 
+- Unify authentication modes. Add post-handshake client authentication.
+
 
 draft-10
 
@@ -1532,55 +1534,68 @@ derivation scheme for all cipher modes.
 The basic TLS Handshake for DH is shown in {{tls-full}}:
 
 ~~~
-     Client                                               Server
-
-     ClientHello
-       + KeyShare              -------->
-                                                     ServerHello
-                                                      + KeyShare
-                                           {EncryptedExtensions}
-                                          {ServerConfiguration*}
-                                                  {Certificate*}
-                                           {CertificateRequest*}
-                                            {CertificateVerify*}
-                               <--------              {Finished}
-     {Certificate*}
-     {CertificateVerify*}
-     {Finished}                -------->
-     [Application Data]        <------->      [Application Data]
-
-            +  Indicates extensions sent in the
-               previously noted message.
-
-            *  Indicates optional or situation-dependent
-               messages that are not always sent.
-
-            {} Indicates messages protected using keys
-               derived from the ephemeral secret.
-
-            [] Indicates messages protected using keys
-               derived from the master secret.
+       Client                                               Server
+  
+Key  / ClientHello                                                  
+Exch \  + KeyShare               -------->                           
+                                                       ServerHello  \ Key
+                                                        + KeyShare  / Exch
+                                             {EncryptedExtensions}  ^ 
+                                             {CertificateRequest*}  | Server
+                                            {ServerConfiguration*}  v Params
+                                                    {Certificate*}  ^
+                                              {CertificateVerify*}  | Auth
+                                 <--------              {Finished}  v
+     ^ {Certificate*}
+Auth | {CertificateVerify*}
+     v {Finished}                -------->
+       [Application Data]        <------->      [Application Data]
+  
+              +  Indicates extensions sent in the
+                 previously noted message.
+  
+              *  Indicates optional or situation-dependent
+                 messages that are not always sent.
+  
+              {} Indicates messages protected using keys
+                 derived from the ephemeral secret.
+  
+              [] Indicates messages protected using keys
+                 derived from the master secret.
 ~~~
 {: #tls-full title="Message flow for full TLS Handshake"}
 
+The handshake can be thought of as having three phases, indicated
+in the diagram above.
 
-The first message sent by the client is the ClientHello {{client-hello}} which contains
+Key Exchange: establish shared keying material and select the
+   cryptographic parameters. Everything after this phase is
+   encrypted.
+
+Server Parameters: establish other handshake parameters 
+(whether the client is authenticated, support for 0-RTT, etc.)
+
+Authentication: authenticate the server (and optionally the client)
+   and provide key confirmation and handshake integrity.
+
+In the Key Exchange phase, the client sends the ClientHello ({{client-hello}})
+message, which contains
 a random nonce (ClientHello.random), its offered protocol version,
 cipher suite, and extensions, and one or more Diffie-Hellman key
 shares in the KeyShare extension {{key-share}}.
 
 The server processes the ClientHello and determines the appropriate
 cryptographic parameters for the connection. It then responds with
-the following messages:
+its own ServerHello which indicates the negotiated connection parameters. [{{server-hello}}]
+If DH is in use, this will contain a KeyShare extension with the
+server's ephemeral Diffie-Hellman share which MUST be in the same group
+as one of the shares offered by the client. The server's KeyShare and
+the client's KeyShare corresponding to the negotiated key exchange
+are used together to derive the Static Secret and Ephemeral
+ Secret (in this mode they are the same).  [{{key-share}}]
 
-ServerHello
-: indicates the negotiated connection parameters. [{{server-hello}}]
-  If DH is in use, this will contain a KeyShare extension with the
-  server's ephemeral Diffie-Hellman share which MUST be in the same group
-  as one of the shares offered by the client. The server's KeyShare and
-  the client's KeyShare corresponding to the negotiated key exchange
-  are used together to derive the Static Secret and Ephemeral
-  Secret (in this mode they are the same).  [{{key-share}}]
+
+The server then sends three messages to establish the Server Parameters:
 
 EncryptedExtensions
 : responses to any extensions which are not required in order to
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1532,55 +1532,68 @@ derivation scheme for all cipher modes.
...
-
-            {} Indicates messages protected using keys
-               derived from the ephemeral secret.
-
-            [] Indicates messages protected using keys
-               derived from the master secret.
+       Client                                               Server
+  
+Key  ^ ClientHello                                                  
+Exch v  + KeyShare               -------->
```
martinthomson: The ^ and v without a line joining them are a little hard to follow.  What about:

```
v---- Key exchange
 ClientHello
       + KeyShare              -------->
                                                       ServerHello
                                                      + KeyShare
v---- Server parameters
                                             {EncryptedExtensions}
                                          {ServerConfiguration*}
                                           {CertificateRequest*} 
v---- Authentication
                                                    {Certificate*}
                                            {CertificateVerify*}
                                <--------              {Finished}
     {Certificate*}
     {CertificateVerify*}
     {Finished}                -------->
```

Also, this doesn't allow the server to speak first, is that intentional?

draft-ietf-tls-tls13.md
```diff
@@ -1601,39 +1610,34 @@ CertificateRequest
...
-  client is not authenticating via a certificate. [{{client-certificate-verify}}]
+: a signature over the entire handshake using the public key
+  in the Certificate message. This message will be omitted if the
+  server is not authenticating via a certificate. [{{certificate-verify}}]
 
 Finished
-: a MAC over the entire handshake computed using the Static Secret
-  and providing key confirmation. [{{server-finished}}]
+: a MAC over the entire handshake. This message provides key confirmation, binds the endpoint's identity
+  to the exchanged keys, and in some modes (see {{zero-rtt-exchange}})
```
martinthomson: Do you mean to reference the 0RTT handshake, or do you mean to refer to PSK/resumption?

draft-ietf-tls-tls13.md
ekr: Why do you think it doesn't allow that?

draft-ietf-tls-tls13.md
ekr: Both, actually.

draft-ietf-tls-tls13.md
martinthomson: Because it shows the server's application data after the server has received the Finished.

draft-ietf-tls-tls13.md
```diff
@@ -1703,42 +1707,62 @@ described below.
...
+               [] Indicates messages protected using keys
+                  derived from the master secret.
 ~~~
 {: #tls-0-rtt title="Message flow for a zero round trip handshake"}
 
+As shown in Figure {{tls-0-rtt}}, the Zero-RTT data is just added
+to the 1-RTT handshake in the first flight. Specifically, the
+client sends its Authentication messages after the ClientHello,
+followed by any application data. The rest of the handshake
+messages are the same as with Figure {{tls-full}}. This implies
```
martinthomson: `Figure {{figure-ref}}` puts the word "Figure" into the final text twice.

draft-ietf-tls-tls13.md
```diff
@@ -1703,42 +1707,62 @@ described below.
...
+that the server can request client authentication even if the
+client offers a certificate on its first flight. This isn't
+necessarily useful but is consistent with the server being
```
martinthomson: I think that you might want to avoid statements like "this isn't necessarily useful"

draft-ietf-tls-tls13.md
```diff
@@ -2887,6 +2791,8 @@ Structure of this message:
            CertificateExtension certificate_extensions<0..2^16-1>;
        } CertificateRequest;
 
+[[TODO: We agreed to add a context string here.]]
```
martinthomson: a context string, fresh entropy, or both?

draft-ietf-tls-tls13.md
```diff
@@ -3025,17 +2931,253 @@ processing the server's Certificate, CertificateVerify, and
...
+
+The following table defines the Handshake Context and MAC Key
+for each scenario:
+
+~~~~
+Mode             Handshake Context                        Base Key
+----             -----------------                        --------
+0-RTT            ClientHello + ServerConfiguration        xSS
+                             + Server Certificate
+                             + CertificateRequest
```
martinthomson: I think that this needs to make it clear that these extras are cached, otherwise ServerConfiguration here can be confused for ServerConfiguration in the 1-RTT case.

draft-ietf-tls-tls13.md
```diff
@@ -3025,17 +2931,253 @@ processing the server's Certificate, CertificateVerify, and
...
+Post-Handshake   ClientHello ... ClientFinished +         master_secret
+                 CertificateRequest
+~~~~
+Note 1: The ServerConfiguration, CertificateRequest, and Server Certificate
+in the 0-RTT case are the messages from the handshake where the
+ServerConfiguration was established.
+
+Note 2: The Handshake Context for the last three rows does not include
+any 0-RTT handshake messages, regardless of whether 0-RTT is used.
+
```
martinthomson: An extra note required to point out that ClientHello ... HelloRetryRequest ... ClientHello is included in all cases.

draft-ietf-tls-tls13.md
```diff
@@ -3025,17 +2931,253 @@ processing the server's Certificate, CertificateVerify, and
...
+> The certificate MUST be appropriate for the negotiated cipher suite's key
+exchange algorithm and any negotiated extensions.
+
+Structure of this message:
+
+%%% Authentication Messages
+       opaque ASN1Cert<1..2^24-1>;
+
+       struct {
+           ASN1Cert certificate_list<0..2^24-1>;
```
martinthomson: TODO: add correlator/context id

draft-ietf-tls-tls13.md
```diff
@@ -3025,17 +2931,253 @@ processing the server's Certificate, CertificateVerify, and
...
+
+All certificates provided by the server MUST be signed by a
+hash/signature algorithm pair that appears in the "signature_algorithms"
+extension provided by the client, if they are able to provide such
+a chain (see {{signature-algorithms}}).
+If the server cannot produce a certificate chain that is signed only via the
+indicated supported pairs, then it SHOULD continue the handshake by sending
+the client a certificate chain of its choice that may include algorithms
+that are not known to be supported by the client. This fallback chain MAY
+use the deprecated SHA-1 hash algorithm.
```
martinthomson: ... if the client indicated that it supported SHA-1 in combination with the signature algorithm in "signature_algorithms".

draft-ietf-tls-tls13.md
ekr: Still an open issue.

draft-ietf-tls-tls13.md
ekr: I am trying to decide whether it should go here or in CertificateVerify.

draft-ietf-tls-tls13.md
ekr: Actually, I'm not sure that this is right, since we already agreed you can just send one that violates the rules.

draft-ietf-tls-tls13.md
```diff
@@ -3107,101 +3267,45 @@ Structure of this message:
...
+hash.
 
 finished_label
 : For Finished messages sent by the client, the string
   "client finished".  For Finished messages sent by the server,
-  the string "server finished".
-{:br }
+  the string "server finished" [[OPEN ISSUE: Do we still need these?
+  labels? Should we expand them to indicate which phase we are
+  sending the messages in?]]
```
martinthomson: If you do keep the labels, then something needs to change.  Having both "server finished" and "server_finished" is going to be quite confusing.

draft-ietf-tls-tls13.md
```diff
@@ -3381,50 +3495,36 @@ The following table describes the inputs to the key calculation for
...
-messages from the handshake where the configuration was established (including the
-type and length fields). Note that
-this requires the client and server to memorize these values.
-{:br }
+The handshake hash is defined as the hash of all handshake messages
+sent or received, starting at ClientHello up to the present time,
+with the exception of the Finished message, including the type and
+length fields of the handshake messages. This is the concatenation
+of all the exchanged Handshake structures in plaintext form (even if
+they were encrypted on the wire).
```
martinthomson: Need to explicitly exclude post-handshake messages. NewSessionTicket, CertificateRequest, etc...

draft-ietf-tls-tls13.md
martinthomson: Ack, I don't think that it matters, but CertificateVerify has a certain merit to it.  Of course, we should merge all the messages and then it will matter much less.

draft-ietf-tls-tls13.md
martinthomson: I think that this is important though.  If we allow inclusion of sha-1 in signature_algorithms, and not permit violations of the rules, then we have a way to signal that a client doesn't want sha-1 any more.  If you let servers just use sha-1 without signaling, then you are in a perpetual sha-1 limbo.

ekr: This is in the existing text, so why don't you send a PR. I suspect
@davegarrett would approve :)

draft-ietf-tls-tls13.md
ekr: Also the 0-RTT messages.

ekr:
I've rebased this version and made some updates. I need to go over the diff again before I merge so there may be some issues, but I wanted to put it up so others could comment ahead of that. Expected merge tonight or tomorrow.

PR #317, opened 2015-10-21
Control SHA-1 with signature_algorithms

martinthomson:
The current draft permits the use of SHA-1 in the certificate chain, which gives SHA-1 a free pass indefinitely.  Since we expressly forbid the use of SHA-1 for signing in TLS itself, we can just permit clients to include it in "signature_algorithms" and use that to determine whether SHA-1 is acceptable.

That means that clients that want to disable SHA-1 (real soon now, we promise), can signal that preference cleanly.

This is built on top of #316 to avoid a nasty rebase when that merges.  Just look at the final commit.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2255,11 +2255,13 @@ hash
   and SHA-512 {{SHS}}, respectively. The "none" value is provided for
   future extensibility, in case of a signature algorithm which does
   not require hashing before signing.  Previous versions of TLS
-  supported MD5 and SHA-1. These algorithms are now deprecated and
-  MUST NOT be offered by TLS 1.3 implementations.  SHA-1 SHOULD NOT be
-  offered, however clients willing to negotiate use of TLS 1.2 MAY
-  offer support for SHA-1 for backwards compatibility with old
-  servers.
+  supported MD5 and SHA-1.  These algorithms are now deprecated.
+  MD5 MUST NOT be offered by TLS 1.3 implementations; SHA-1 SHOULD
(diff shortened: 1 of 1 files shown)


martinthomson:
Fixed nits and rebased.

davegarrett:
Travis is still mad at this. I think it's because there's 2 "%%### Key Exchange Messages" appendix, and the generator script is erroring on that.
https://travis-ci.org/tlswg/tls13-spec/builds/89405121

This is from @ekr's PR #316 commits on which this is based, though, not the ones for this issue. That PR doesn't have any Travis result.

ekr:
@martinthomson are you happy with the basic text here? If so, I can rebase.

martinthomson:
@ekr, I reviewed the thread and didn't find anything actionable other than Russ' suggestion to refer to RFC 5280 for terminology.  In doing so, I noticed that we use "certificate chain" whereas 5280 uses "certification path".  I also notice that we duplicate some of the advice regarding certificate validation and the "bad_certificate" alert.

BTW, I have no idea on how you might rebase this.  I tried against several base points but was unsuccessful in even applying the changes.

seanturner:
After reviewing the list discussion, PR#317 (https://github.com/tlswg/tls13-spec/pull/317) is okay to merge. Initially, there was some confusion about exactly what changes were being proposed (PR#317 is built on PR#231) but that seems to have no be cleared up; text about not checking the signature on a self-signed certificates has been added in various places.

For reference:
https://mailarchive.ietf.org/arch/msg/tls/8O1zcuoGwbNjrZZr6Z0wJrTUYww

PR #319, opened 2015-10-21
Issue225 deprecated extensions

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -3595,6 +3595,16 @@ for general use and should be assumed to be potentially unsafe. The set
...
 currently deployed and properly configured TLS implementations.
 
+#### Deprecated Extensions
+
+The following extensions are no longer applicable to TLS 1.3, although
+TLS 1.3 clients MAY send them if they are willing to negotiate them
+with prior versions of TLS. TLS 1.3 servers MUST ignore these
+extensions if they are negotiating TLS 1.3: max_fragment_length,
+truncated_hmac, ec_point_formats, srp, encrypt_then_mac,
+extended_master_secret, SessionTicket, and renegotiation_info.
```
martinthomson: Probably want to include citations here for each of these.

davegarrett:
Can we just cite the registry itself, which links to the various RFCs?
https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#tls-extensiontype-values-1

This would let us avoid adding informative references to stuff we're explicitly telling people to ignore.

Technically, the only info really needed to follow this would be the ID numbers, which could be added for each.

ekr:
Eh, I already did it and it seems like useful context.

PR #326, opened 2015-10-29
cleanup ClientHello message placeholders

davegarrett:
I noticed that a definition of the SessionID type is missing from the current draft. Here's a few little cleanups to deal with the legacy ClientHello bits like this.

* drop empty orphaned "%%% Hello Messages" block
* drop CompressionMethod enum in favor of just saying it's exactly one zero byte (defining an enum to tell someone that a value is zero is just clutter at this point)
* change the placeholder session_id & compression_methods fields to opaque vectors (that's all they are to TLS 1.3 implementations)
* rename the placeholders to have a "legacy_" prefix so it's clear that they're just placeholders now (also makes it clear here that session resumption has changed)
* RFC2119 capitalize a "should"

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1895,8 +1895,6 @@ random_bytes
 Note: Versions of TLS prior to TLS 1.3 used the top 32 bits of
 the Random value to encode the time since the UNIX epoch.
 
-
-%%% Hello Messages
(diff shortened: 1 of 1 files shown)


ekr:
Manually rebased and merged separately.

davegarrett:
@ekr: Thanks :)

PR #327, opened 2015-11-02
Issue318 278 additional data

ekr:


zakar:
Why additional data input is empty in TLS1.3?

seanturner:
Starts in this thread:

https://mailarchive.ietf.org/arch/msg/tls/bPccVqZyu2vO_JE-91cLS9n8gXg

spt

PR #332, opened 2015-11-04
Add Barnes

IETF-WG-Chairs-Training:


IETF-WG-Chairs-Training:
You should use a different affiliation.

draft-ietf-tls-tls13.md
```diff
@@ -28,6 +28,11 @@ author:
        name: Eric Rescorla
        organization: RTFM, Inc.
        email: ekr@rtfm.com
+   -
+       ins: R. Barnes
+       name: Richard Barnes
+       organization: This is a demo, Inc.
```
IETF-WG-Chairs-Training: Should be Mozilla

PR #335, opened 2015-11-05
Use correct pronoun for "client" and "attacker"

ttaubert:
The spec always uses the gender-neutral possessive pronoun "its" for "client", so I think we should definitely fix that occurrence.

Also, it seems that throughout the document we try to not use gender-specific pronouns for "attacker". I'm not sure about "their" though, as it looks like we usually try to avoid that by restructuring sentences, but I may be mistaken. We could also refer to multiple attackers?

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1582,7 +1582,7 @@ Finished
   the Static Secret. [{{server-finished}}]
 {:br }
 
-Upon receiving the server's messages, the client responds with his final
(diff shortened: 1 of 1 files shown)


davegarrett:
In technical writing, we should generally use "they" and "their" everywhere when (potentially) referring to a person or persons. ("it" is best for denoting clients/servers) The word "they" gives both gender & quantity neutral language; no assumptions. (this use is older than Shakespeare; it's just underused)

PR #347, opened 2015-11-20
iana considerations

seanturner:
Adding a note to registry so that those that start there understand why the cipher suites are marked the way they are in column.


ekr:
@seanturner it seems like there's not yet consensus on the list for the semantics here. Holding off on merging this PR pending chair declaration one way or the otehr

Issue #277, opened 2015-10-05
prohibit immediate plain PSK negotiation if no usable (EC)DHE key is provided, but could be, and any (EC)DHE suite is offered at a higher priority


davegarrett:
There's a specific risk with plain PSK (no (EC)DHE) that needs explicit handling here. Assume the following:
1) Client offers both plain PSK & any other (EC)DHE suite (e.g. (EC)DHE-PSK)
2) Server supports both as well
3) The client does not provide a (EC)DHE key share that the server can use

There is a risk that the server will:
1) Skip all (EC)DHE suites due to lack of usable key share
2) Continue checking other suites, and then negotiate plain PSK if supported

Instead, we need to explicitly make sure that this scenario triggers a HelloRetryRequest so that the client has a chance to get PFS and not accidentally be downgraded to plain PSK. (especially if plain PSK is lowest priority, as it should be)

Edited to add:
Of course, this also needs to handle the case where there is no group support overlap and plain PSK is the correct choice.

ekr:
This doesn't seem right, because in order to do resumption, you offer plain PSK and your ordinary ECDHE (non-PSK) cipher suite and you expect the server to pick one or the other, and quite
possible that you want them to prefer PSK-only.

It might be possible to restrict to not allowing ECDHE-PSK and plain PSK together, but then you basically couldn't resume ECDHE-PSK.

davegarrett:
If plain PSK is top priority, then it should get picked and that's that. If it's not, e.g. the client cipher list is (EC)DHE-CERT-AES, (EC)DHE-PSK, plain PSK, (in this descending priority order) and there's no usable group in the client's key share, but there is a usable PSK identity, the point I'm making is that the correct route is to request a retry to try to get a usable (EC)DHE key to do one of the first two rather than just going straight to plain PSK. In the case of resumption, the client knows what works with the server and has no reason to offer an unusable offer set. If that changes due to a config change, resumption really isn't an option anymore anyway.

The retry mechanism explicitly sets up (EC)DHE negotiation as 1RTT ideal, or 2RTT as fallback. I'm concerned that also offering plain PSK could lead to short-circuiting that and servers just ditching PFS to save a RTT (possibly accidentally).

davegarrett:
I'll add the word "immediate" to the title to make it more clear. If the retry fails, then plain PSK is correct. It should attempt it, though.

davegarrett:
[...]

PR #349, opened 2015-11-26
always require key_share extension with (EC)DHE suites

davegarrett:
The current text is a mix of omitting and sending empty, as pointed out by Ilari on the mailing list. This fixes things and switches to an always-send model.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2419,14 +2419,16 @@ must consider the supported groups in both cases.
 The "key_share" extension contains the endpoint's cryptographic parameters
 for non-PSK key establishment methods (currently DHE or ECDHE).
 
-Clients which offer one or more (EC)DHE cipher suites
-MUST send at least one supported KeyShare value and
-servers MUST NOT negotiate any of these cipher suites unless a supported
+Clients which offer one or more (EC)DHE cipher suites MUST send this
+extension and SHOULD send at least one supported KeyShareEntry value.
+Servers MUST NOT negotiate any of these cipher suites unless a supported
 value was provided.
-If this extension is not provided in a ServerHello or retried ClientHello,
+If this extension is not provided in a ServerHello or ClientHello,
 and the peer is offering (EC)DHE cipher suites, then the endpoint MUST close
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2430,8 +2432,8 @@ The "extension_data" field of this extension contains a
 
 client_shares
 : A list of offered KeyShareEntry values in descending order of client preference.
-  This vector MUST NOT be empty. Clients not providing a KeyShare MUST instead
-  omit this extension from the ClientHello.
+  Servers MUST NOT negotiate any connection if this vector is empty and instead
+  MUST respond with a HelloRetryRequest (see {{hello-retry-request}}).
```
ekr: This doesn't seem quite correct. What if you offered another cipher suite that *was* negotiable without any shares?

ekr:
This looks good to me modulo the issue noted above

draft-ietf-tls-tls13.md
davegarrett: See also issue #277. If the (EC)DHE suite is higher priority, it MUST NOT try and find another lower priority suite. That just risks downgrading when the client was trying to request a server group selection via retry. If there's a higher priority suite than the (EC)DHE one(s), then we don't even get to the point of caring about this extension. This could be stated in detail.

davegarrett: Even with that, there's still a conflict on how to handle a lack of mutually supported key share when there is a mutually supported group: retry or negotiate lower priority suite? This complication is an example of why I don't really like the retry mechanic.

davegarrett:
@ekr: Ok, I've pushed a second commit here which clarifies that point and also reconciles the fact that the two extensions here both claim to determine group priority. Suggestions to reduce verbosity without reducing clarity would be appreciated. (this commit fixes issue #277)

davegarrett:
Typo fixed & rebased.

davegarrett:
Since the list is in favor of "forward secrecy" over "forward security", I've changed my use of the term in this PR to the former (a commit just landed to change the existing uses elsewhere). Not rebased again; still merges cleanly (for the moment, at least).

PR #354, opened 2015-12-01
Implement rekey. Fixes #125, Fixes #4

### LINKED ISSUE #4: Rekeying
If we ditch renegotiation (#3), then we will need a way to rekey sessions to extend their lifetime.

Nico suggests that sending ChangeCipherSpec and then using an increasing counter to extract new keys (i.e., add a counter to the key expansion PRF input), which would seem to be sufficient.

### LINKED ISSUE #125: Data transfer limitation per connection
The counter has a size of (2^64-1).The records are max (2^14 + 2048). 
So, my bold calculation would come to (2^64-1) / (2^14 + 2048) / 2^40 = 910.22 TiB of data.

In Big Data, video streaming and other mass data cases 910TB data can be a limitation.

I would like to propose the counter to go to 2^128-1, which would result in a connection that can reach (2^128-1) / (2^14 + 2048) / 2^80 = 1.52 * 10^10 YiB.

If byte preservation is an issue, I'm willing to accept 2^96-1 as the counter max, resulting in (2^96-1) / (2^14 + 2048) / 2^80 = 3.55 YiB as maximum connection.

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -3532,6 +3571,25 @@ schedule with a black-box HKDF API, albeit at some loss of efficiency
...
+update its sending traffic keys using the KeyUpdate handshake message
+{{key-update}}.  The next generation of traffic keys is computed by
+generating traffic_secret_N+1 from traffic_secret_N as described in
+this section then re-deriving the traffic keys as described in
+{{traffic-key-calculation}}.
+
+The next-generation traffic_secret is computed as:
+
+  traffic_secret_N+1 = HKDF-Expand-Label(traffic_secret_N,
+                                       "traffic_secret", "", L)
```
davegarrett: The string probably shouldn't have the underscore, as the other secrets including the prior traffic secret N do not.

draft-ietf-tls-tls13.md
```diff
@@ -3429,6 +3431,35 @@ flight, the server MUST be prepared to receive an arbitrary number
...
 
 
+#### Key Update 
+
+struct {} KeyUpdate;
+
+The KeyUpdate handshake message is used to indicate that the sender is
+updating its sending cryptographic keys. This message can be sent by
+the server after sending its first flight and the client after sending
+its second flight. Implementations prior to this time MUST generate a
```
martinthomson: "Endpoints that receive a KeyUpdate message prior to receiving a Finished message MUST generate a fatal ..."

draft-ietf-tls-tls13.md
```diff
@@ -3429,6 +3431,35 @@ flight, the server MUST be prepared to receive an arbitrary number
...
+fatal "unexpected_message" alert.  After sending a KeyUpdate, the
```
martinthomson: After sending a KeyUpdate *message*, the sender

martinthomson:
Is good.

draft-ietf-tls-tls13.md
```diff
@@ -3429,6 +3431,35 @@ flight, the server MUST be prepared to receive an arbitrary number
...
+sender SHALL send all its traffic using the next generation of keys,
+computed as described in {{updating-traffic-keys}}. Upon receiving a
+KeyUpdate, the receiver MUST update their receiving keys and if they
+have not already updated their sending state up to or past the then
+current receiving generation MUST immediately send their own
```
kaduk: What does "immediately" mean, and what are the consequences if the receiver does not actually send this KeyUpdate?  If there is no KeyUpdate should the original sender send an alert and close the connection?  If so, under what conditions?

draft-ietf-tls-tls13.md
```diff
@@ -1909,6 +1909,7 @@ processed and transmitted as specified by the current active session state.
            client_key_exchange_RESERVED(16),
            server_configuration(17),
            finished(20),
+           key_update(21),
```
hannesm: handshake_type 21 is already taken by [rfc4366](https://tools.ietf.org/html/rfc4366#section-3.3) for `Certificate URL` (first free seems to be 24)

PR #359, opened 2015-12-08
Added note that 0-RTT is compatible with PSK

waywardgeek:
This is a one-sentence addition to clarify that 0-RTT is compatible with PSK.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1789,6 +1789,8 @@ client offers a certificate on its first flight. This
 is consistent with the server being
 able to ask for client authentication after the handshake is
 complete. See {{post-handshake-authentication}}.
+Note that a Zero-RTT handshake can either create a new session, or
+resume a previous PSK session.
 
 Note: because sequence numbers continue to increment between the
 initial (early) application data and the application data sent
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1789,6 +1789,8 @@ client offers a certificate on its first flight. This
 is consistent with the server being
 able to ask for client authentication after the handshake is
 complete. See {{post-handshake-authentication}}.
+Note that a Zero-RTT handshake can either create a new session, or
+resume a previous PSK session.
```
martinthomson: This isn't quite right.  PSK and resumption are the same thing.  I think that you mean to say: 0-RTT handshakes can be used with pre-shared keys.

The concept of a session in the sense you use here is no longer relevant.

davegarrett:
I'd prefer it to be more clear on what's actually supposed to happen here. The flow chart is currently only applicable to the non-PSK case. Instead of the currently proposed text, I'd suggest something akin to:

"When offering PSK support, the PreSharedKeyExtension will be used instead of (or in addition to) the KeyShare extension as specified above."

ekr:
Added something more along the lines of what Garrett proposed.

PR #360, opened 2015-12-08
Added suggestions for mitigating security weaknesses when using 0-RTT

waywardgeek:
This needs some work, I think.  I do think it is important to offer some guidance, as the spec currently reads as if we are discouraging 0-RTT handshakes due to security weaknesses.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1817,6 +1817,17 @@ the 0-RTT data without detection. If the client's ephemeral share
 is compromised and client authentication is used, then the attacker
 can impersonate the client on subsequent connections.
 
+It is up to the implementation and application layers above TLS to mitigate
+the impact of these weaknesses.  For example, to support strong client
+authentication and forward secrecy over 0-RTT connections, it is possible to
+issue a new ticket on the first 1-RTT connection, and on every resumed session
+after that.  Each co-located group of servers could ensure that tickets they
+issue are used only once by maintaining a local cache of ticket decryption
+keys.  An accurate clock on the client could ensure freshness of client
+signatures.  In addition it is a good idea to make all request idempotent,
+even POST requests, to mitigate the impact of packet replay, which happen
+often as a result of network errors rather than malicious attacks.
+
 
 ### Resumption and PSK
 


draft-ietf-tls-tls13.md
```diff
@@ -1817,6 +1817,17 @@ the 0-RTT data without detection. If the client's ephemeral share
...
 can impersonate the client on subsequent connections.
 
+It is up to the implementation and application layers above TLS to mitigate
+the impact of these weaknesses.  For example, to support strong client
+authentication and forward secrecy over 0-RTT connections, it is possible to
+issue a new ticket on the first 1-RTT connection, and on every resumed session
+after that.  Each co-located group of servers could ensure that tickets they
+issue are used only once by maintaining a local cache of ticket decryption
+keys.  An accurate clock on the client could ensure freshness of client
+signatures.  In addition it is a good idea to make all request idempotent,
```
martinthomson: I have no idea how an accurate clock will help here.  Or you haven't made that case very well.

draft-ietf-tls-tls13.md
```diff
@@ -1817,6 +1817,17 @@ the 0-RTT data without detection. If the client's ephemeral share
...
+even POST requests, to mitigate the impact of packet replay, which happen
+often as a result of network errors rather than malicious attacks.
```
martinthomson: This idempotence thing is completely without context and you certainly shouldn't talk about POST requests, which are HTTP-specific (just one of many users of TLS).

draft-ietf-tls-tls13.md
```diff
@@ -1817,6 +1817,17 @@ the 0-RTT data without detection. If the client's ephemeral share
 is compromised and client authentication is used, then the attacker
 can impersonate the client on subsequent connections.
 
+It is up to the implementation and application layers above TLS to mitigate
+the impact of these weaknesses.  For example, to support strong client
+authentication and forward secrecy over 0-RTT connections, it is possible to
+issue a new ticket on the first 1-RTT connection, and on every resumed session
+after that.  Each co-located group of servers could ensure that tickets they
+issue are used only once by maintaining a local cache of ticket decryption
+keys.  An accurate clock on the client could ensure freshness of client
```
martinthomson: This suggestion has some problems: 

it assumes that there is exactly one connection at any point in time, and not multiple 0-RTT connections established based on a single 1-RTT connection (which is highly likely.

The tickets need to be perfectly atomic and non-portable.

I'd rather that you didn't attempt to give advice like this.  This is very hard to get right.  It's better to say that it is hard and to concentrate on precisely characterizing the problem (as the existing text does).

Note that strong client authentication and forward secrecy are possible on 0-RTT connections, but both require an absolute guarantee of single-use.  If you insist on describing a "solution" to the limitations, you need to be very clear what conditions it to be a genuine solution.  Either way, attempting to describe a mechanism in even as much detail as you do is very risky.

draft-ietf-tls-tls13.md
waywardgeek: A traditional proof of possession provides two things: uniqueness and freshness.  The server knows the signature is unique since the server picked a unique challenge to sign.  We can get the same effect by ensuring the servers will never accept the same signed nonce twice.

As you point out, the more difficult trait we want is freshness.  Traditional challenge-response happens in less than a second, proving that the key was available to the client for signing within the last 1 second.  OTP keys have accurate clocks, which is how they are able to show you a pin to type in when prompted.  They are accurate to a few seconds typically.  If a user kept his client certificate in a security key with a similar clock, the key would create the nonce and timestamp when it creates the signature.  The company issuing the keys can be confident that only an attacker with the access and skills to hack the security key could fake the signed nonce and timestamp.  This is not quite as strong as a real proof of possession.  I call it "compelling evidence".

Anyway, with good secure clocks, we can indeed get the freshness property we need.

waywardgeek: I think I left out an important detail that might be confusing.  The server needs to reject the CertificateVerify if the signed timestamp is too old, or too far in the future.  a 30 second to 1 minute window is common, I think, with OTPs.  This is not as strong as the typical 1-ish second window we get with a real proof of possession, but we're just suggesting techniques for mitigating weakness here.

draft-ietf-tls-tls13.md
waywardgeek: It is up to the application layer to mitigate the impact of replayed requests.  The most common application is HTTPS, which is why I think it is OK to use it as an example.  My writing skill is very poor, and I know this needs work...

draft-ietf-tls-tls13.md
waywardgeek: I agree with your concerns, but at the same time, I feel the document would be improved with some guidance about mitigation.  As it stands, the warnings read like a Stay Away sign.  Another direction is to accept the forward secrecy limitations and implement QUIC-inspired strike registers, without using tickets.  This can give client certificates more value, and many people don't care about a few days of lost forward secrecy.

In any case, dealing with replay when implementing HTTPS involves making all requests idempotent at the application layer, even POST requests.  This is true for TLS 1.2 as well.  All requests are subject to being replayed, either due to network errors or because of a malicious attack.

draft-ietf-tls-tls13.md
martinthomson: I think that you will find that clock sync is basically unattainable in practice.  However, I don't think that it's necessary.  Since the server generates the challenge (even if it's on a previous connection), it is entirely in control of freshness.  All it has to do is tune the window over which it will accept 0-RTT.  This makes the client clock irrelevant here.

draft-ietf-tls-tls13.md
martinthomson: Explicitly making it an example helps: For example, in HTTP [RFC7230], limiting the use of 0-RTT to idempotent requests [RFC7231, relevant section] ensures that replay attacks do not cause problems.  Your advice on POST is generically sound (it's a principle I try to abide by, certainly), but it doesn't belong here.

PR #363, opened 2015-12-11
Receiving TLS 1.2 ClientHello:  MAY/may to might

MikeBishop:
When talking about handling downlevel clients, the spec variously says a server "MAY" receive such messages, or "may" receive such messages.  Since this is discussing a possibility (not a permitted behavior), the normative language isn't correct; lower-case versions of the RFC2119 terms have the potential to confuse readers.  Suggest changing all of these to "might" to denote a possible occurrence server implementers will need to handle.


davegarrett:
Yes, this looks like an improvement.

ekr:
LGTM

PR #366, opened 2015-12-15
Use tables

martinthomson:
This was using figures.  Using tables isn't that much harder on the eyes in the markdown form and much nicer in the HTML format.


davegarrett:
When viewing richdiff to get a sense of what it will look like, the font in the tables is notably larger than the surrounding text. This easily fixable?

PR #383, opened 2015-12-22
Clarifying text specifying the definition of the master secret

huitema:
In section 6.2, the diagrams make reference to three secrets: static, ephemeral, and master. But the master secret was not defined in the introduction of the section.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1542,7 +1542,7 @@ TLS supports three basic key exchange modes:
 - A combination of a symmetric key and Diffie-Hellman
 
 Which mode is used depends on the negotiated cipher suite. Conceptually,
-the handshake establishes two secrets which are used to derive all the
+the handshake establishes three secrets which are used to derive all the
 keys.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1553,8 +1553,12 @@ Ephemeral Secret (ES): A secret which is derived from fresh (EC)DHE
 Static Secret (SS): A secret which may be derived from static or
    semi-static keying material, such as a pre-shared key or the
    server's semi-static (EC)DH share.
+   
+Master Secret (SS): A secret derived from both the static
```
davegarrett: This should be "MS" here instead of "SS". That said, I think I'd prefer just dropping this parenthetical as it's not abbreviated like that anywhere like the other two are.

martinthomson: I'm actually thinking that the abbreviation would be useful.  The inconsistency is a little troublesome in a lot of places (tables and charts primarily).

ekr:
Merged separately.

PR #384, opened 2015-12-24
More explicitly require group matching between 0-RTT and 1-RTT

ekr:
Per a question by Ilari

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2764,7 +2764,12 @@ MUST perform the following checks:
 - The configuration_id matches a known server configuration.
 - The client's cryptographic determining parameters match the
   parameters that the server has negotiated based on the
-  rest of the ClientHello.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2764,7 +2764,12 @@ MUST perform the following checks:
 - The configuration_id matches a known server configuration.
 - The client's cryptographic determining parameters match the
   parameters that the server has negotiated based on the
-  rest of the ClientHello.
+  rest of the ClientHello. If (EC)DHE is selected, this includes
+  verifying that (1) the ClientHello contains a key from the
+  same group that is indicated by the server configuration and
+  (2) that the server has negotiated that group (and will
+  therefore include a share from that group in its own
+  "key_share" extension.
```
davegarrett: missing a closing paren, though just dropping the opening paren is probably better

PR #385, opened 2015-12-24
Reuse KeyShareEntry in ServerConfiguration. No wire format change

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -2979,8 +2979,7 @@ Structure of this Message:
           struct {
               opaque configuration_id<1..2^16-1>;
               uint32 expiration_date;
-              NamedGroup group;
-              opaque server_key<1..2^16-1>;
+              KeyShareEntry key_share;
```
davegarrett: Please name this something at least a little different from the extension's type name. Something like "static_key_share" would also be a little more descriptive.

ekr: Sorry, you mean !KeyShareEntry?

davegarrett: No, I mean:
KeyShareEntry static_key_share;
rather than:
KeyShareEntry key_share;

It's just that we refer directly to "key_share" by name quite a few times and I don't want to risk conflating the two slightly, even though we're reusing a struct from it.

ekr: OK, sold.

PR #394, opened 2015-12-28
minor fixes

davegarrett:
This should be "chose" not "choosed", which isn't really an accepted word.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2781,7 +2781,7 @@ trial decrypt each record with the 1-RTT handshake keys
 until it finds one that decrypts properly, and then pick up
 the handshake from that point.
 
-If the server choosed to accept the "early_data" extension,
+If the server chooses to accept the "early_data" extension,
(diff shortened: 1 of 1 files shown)


davegarrett:
Added another commit here for a minor wording fix that snuck in with changing all extensions to be referred to by type ID name.

draft-ietf-tls-tls13.md
```diff
@@ -2781,7 +2781,7 @@ trial decrypt each record with the 1-RTT handshake keys
 until it finds one that decrypts properly, and then pick up
 the handshake from that point.
 
-If the server choosed to accept the "early_data" extension,
+If the server chose to accept the "early_data" extension,
```
ekr: I meant for this to be chooses

ekr: I agree that "choosed" isn't a word

davegarrett:
Ok, I switched it to "chooses" instead & squashed my little fixes together here. There's also a fix for the "will MAY" bit that I think we discussed at some point but its fix didn't get in.

PR #398, opened 2015-12-31
fix whitespace

davegarrett:
Fix the alignment in the Handshake switch. Just a whitespace patch.


davegarrett:
Appending another quick whitespace fix changeset here, as more trailing whitespace crept back in. (easily generated by my text editor)

PR #404, opened 2016-01-22
Revise signature algorithm negotiation.

davidben:
This PR contains proposed specification text for the mailing list thread here:
https://www.ietf.org/mail-archive/web/tls/current/msg19060.html

SignatureAlgorithm and HashAlgorithm are merged into a single u16 value. Fold signing curves into it as well. This is wire-compatible with TLS 1.2, generalizes better to Ed25519 which combines all parameters into a single algorithm, and allows the server to report signature curve preferences to the client.

Per discussion, there are only three ECDSA values, one per curve, so mismatching curve and hash is no longer allowed. They use the same values as the old TLS 1.2 ECDSA code points. (If the WG decides differently here, that can be tweaked.)

Otherwise, I tried to keep SHOULD NOTs and MUST NOTs the same. MUST NOT values (MD5 and SHA-224) are unallocated and remain at MUST NOT. DSA and SHA-1 remain allocated as they're merely SHOULD NOT.

Note: SHA-1 values are limited to rsa_pkcs1_sha1 (formerly {sha1, rsa}) and dsa_sha1 (formerly {sha1, dsa}). TLS 1.2's {sha1, ecdsa} does not have a corresponding TLS 1.3 value due to the curve/hash matching. (Thus it implicitly falls in the unallocated pair MUST NOT bucket.)

Finally, with NamedGroups limited to (EC)DH, this also removes the 'ecdh_' prefix on ecdh_x25519 and ecdh_x448. X25519 and X448 already refer exclusively to Diffie-Hellman primitives and now there is no need to disambiguate between signing curves anyway.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2316,20 +2316,20 @@ The "extension_data" field of this extension contains a
 %%% Signature Algorithm Extension
        enum {
            // RSASSA-PKCS-v1_5 algorithms.
-           rsa_pkcs1_sha1 (0x0101),
-           rsa_pkcs1_sha256 (0x0301),
-           rsa_pkcs1_sha384 (0x0401),
+           rsa_pkcs1_sha1 (0x0201),
+           rsa_pkcs1_sha256 (0x0401),
+           rsa_pkcs1_sha384 (0x0501),
            rsa_pkcs1_sha512 (0x0601),
 
            // DSA algorithms (deprecated).
-           dsa_sha1 (0x0102),
-           dsa_sha256 (0x0302),
-           dsa_sha384 (0x0402),
+           dsa_sha1 (0x0202),
+           dsa_sha256 (0x0402),
+           dsa_sha384 (0x0502),
            dsa_sha512 (0x0602),
 
            // ECDSA algorithms.
-           ecdsa_secp256r1_sha256 (0x0303),
-           ecdsa_secp384r1_sha384 (0x0403),
+           ecdsa_secp256r1_sha256 (0x0403),
+           ecdsa_secp384r1_sha384 (0x0503),
            ecdsa_secp521r1_sha512 (0x0603),
 
            // RSASSA-PSS algorithms.
@@ -2342,10 +2342,10 @@ The "extension_data" field of this extension contains a
            ed448 (0x0704),
 
            // Reserved Code Points.
-           obsolete_RESERVED (0x0000..0x0100),
-           obsolete_RESERVED (0x0103..0x0300),
-           obsolete_RESERVED (0x0304..0x0400),
-           obsolete_RESERVED (0x0404..0x0600),
+           obsolete_RESERVED (0x0000..0x0200),
+           obsolete_RESERVED (0x0203..0x0400),
+           obsolete_RESERVED (0x0404..0x0500),
+           obsolete_RESERVED (0x0504..0x0600),
            obsolete_RESERVED (0x0604..0x06FF),
            private_use (0xFE00..0xFFFF),
            (0xFFFF)
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -226,6 +226,15 @@ informative:
          ins: T. van der Merwe
        -
          ins: S. Scott
+  SLOTH:
+       title: "Transcript Collision Attacks: Breaking Authentication in TLS, IKE, and SSH"
+       author:
+       -
+         ins: K. Bhargavan
+       -
+         ins: G. Leurent
+       seriesinfo: Network and Distributed System Security Symposium (NDSS 2016)
+       date: 2016
 
 --- abstract
 
@@ -761,13 +770,12 @@ current session state (see {{connection-states}}).
 A digitally-signed element is encoded as a struct DigitallySigned:
 
        struct {
-          SignatureAndHashAlgorithm algorithm;
+          SignatureAlgorithm algorithm;
           opaque signature<0..2^16-1>;
        } DigitallySigned;
 
 The algorithm field specifies the algorithm used (see {{signature-algorithms}}
-for the definition of this field). Note that the algorithm
-field was introduced in TLS 1.2, and is not in earlier versions. The signature is a digital signature
+for the definition of this field). The signature is a digital signature
 using those algorithms over the contents of the element. The contents
 themselves do not appear on the wire but are simply calculated. The length of
 the signature is specified by the signing algorithm and key.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2323,102 +2315,138 @@ The "extension_data" field of this extension contains a
...
-           (255)
+           // Reserved Code Points.
+           obsolete_RESERVED (0x0000..0x0100),
+           obsolete_RESERVED (0x0103..0x0300),
+           obsolete_RESERVED (0x0304..0x0400),
+           obsolete_RESERVED (0x0404..0x0600),
+           obsolete_RESERVED (0x0604..0x06FF),
+           private_use (0xFE00..0xFFFF),
+           (0xFFFF)
        } SignatureAlgorithm;
```
davegarrett: I don't think we should be doing out-of-order enums and I don't want the obsolete ranges to be unnamed either. I would much rather we stick to the same ordering, as I first proposed.
https://www.ietf.org/mail-archive/web/tls/current/msg18035.html

If we really want a list sorted by cert, we can stick one in an appendix.

davidben: I tried that originally, but it's much harder to read. This works better with the "rsa_pkcs1_sha1, etc." lines below it. (That, in turn, matches how NamedGroup values are defined.) I figure for future allocations, the 0x0700..(whatever) range is what actually matters anyway and that's sorted.

 The Enumerateds section also explicitly says:

> Since the elements of the enumerated are not ordered, they can be assigned any unique value, in any order.

Though it's true that none are out-of-order right now.

draft-ietf-tls-tls13.md
```diff
@@ -2323,102 +2315,138 @@ The "extension_data" field of this extension contains a
...
-Note: TLS 1.3 servers might receive TLS 1.2 ClientHellos which do not contain
-this extension. If those servers are willing to negotiate TLS 1.2, they MUST
-behave in accordance with the requirements of {{RFC5246}} when negotiating
-that version.
+* TLS 1.2 ClientHellos may omit this extension.
+
+* In TLS 1.2, the extension contained hash/signature pairs. The pairs are
+  encoded in two octets, so SignatureAlgorithm values have been allocated to
+  align with TLS 1.2's encoding. Unallocated pairs are deprecated as of TLS
+  1.3. These values MUST NOT be offered or negotiated by any implementation.
```
davidben: It occurs to I don't actually know that all existing TLS 1.2 ECDSA servers are willing to sign SHA-256 or better.  This probably wants some SHOULD NOT'd obsolete_ecdsa_sha1(0x0103) allocation that's ignored in TLS 1.3 but might be advertised for TLS 1.2, at least until we have some evidence that this is not needed for TLS 1.2 compatibility.

(Whether ECDSA-SHA-1 certificates exist in the Web PKI is largely answerable with CT, but my initial explorations suggested that some servers also insist on signing ServerKeyExchange with SHA-1. Almost all RSA of course, but I didn't think to separate Chrome's current metrics by hash type.)

draft-ietf-tls-tls13.md
```diff
@@ -2323,102 +2315,138 @@ The "extension_data" field of this extension contains a
...
        enum {
-           none(0),
-           md5_RESERVED(1),
-           sha1(2),
-           sha224_RESERVED(3),
-           sha256(4), sha384(5), sha512(6),
-           (255)
-       } HashAlgorithm;
+           // RSASSA-PKCS-v1_5 algorithms.
+           rsa_pkcs1_sha1 (0x0101),
```
ekr: Isn't this MD5/RSA?

draft-ietf-tls-tls13.md
```diff
@@ -2323,102 +2315,138 @@ The "extension_data" field of this extension contains a
...
+           rsa_pkcs1_sha256 (0x0301),
+           rsa_pkcs1_sha384 (0x0401),
+           rsa_pkcs1_sha512 (0x0601),
```
ekr: I'm kind of confused by these code point assignments, which seem to not actually match their matching SignatureAndHashAlgorithm values from 1.2.

draft-ietf-tls-tls13.md
davidben: Whoops. Those should end in 2. I'll go fix that.

davidben: Er, by that I mean this should be 0x0201. I think the others in this group are correct?

draft-ietf-tls-tls13.md
davidben: Fixed. I seem to have shifted the hash values by one. Sorry about that!

draft-ietf-tls-tls13.md
```diff
@@ -2323,102 +2315,138 @@ The "extension_data" field of this extension contains a
...
-: This field indicates the hash algorithms which may be used.  The
-  values indicate support for unhashed data, SHA-1, SHA-256, SHA-384,
-  and SHA-512 {{SHS}}, respectively. The "none" value is provided for
-  signature algorithms which do not require hashing before signing,
-  such as EdDSA.
-
-  Previous versions of TLS
-  supported MD5, SHA-1, and SHA-224.  These algorithms are now deprecated.
-  MD5 and SHA-224 MUST NOT be offered by TLS 1.3 implementations; SHA-1 SHOULD
-  NOT be offered.
```
davegarrett: This deprecation paragraph is not in the revised version. Please add the explicit deprecation of these hashes back somewhere here. In particular, SHA-224 was listed as a "MUST NOT", but this removes the only mention of that without replacing it.

davidben: That's in the paragraph in line 2430. Since we're no longer maintaining continuity with TLS 1.2's registry, the things that are MUST NOT (rather than SHOULD NOT) seemed simpler to just not define at all. This was done per mailing list feedback here:
http://www.ietf.org/mail-archive/web/tls/current/msg19070.html

(Although it cited MD5 and DSA. DSA is still mentioned because the existing text gave it a SHOULD NOT rather than MUST NOT.)

davegarrett: On Sunday, March 20, 2016 03:39:06 pm David Benjamin wrote:
> That's in the paragraph in line 2430. Since we're no longer maintaining continuity with TLS 1.2's registry, the things that are MUST NOT (rather than SHOULD NOT) seemed simpler to just not define at all. This was done per mailing list feedback here:
> http://www.ietf.org/mail-archive/web/tls/current/msg19070.html
> 
> (Although it cited MD5 and DSA. DSA is still mentioned because the existing text gave it a SHOULD NOT rather than MUST NOT.)

Explicit is better than implicit. It's too easy to not notice this change. Please add a sentence somewhere explicitly stating that SHA-224 is deprecated alongside MD5. It's the only way to ensure implementations to take note of this.

davidben: Tweaked line 2430 slightly to mention SHA-224.

davegarrett: On Sunday, March 20, 2016 04:21:17 pm David Benjamin wrote:
> Tweaked line 2430 slightly to mention SHA-224.

Updated text looks good. Thanks.

draft-ietf-tls-tls13.md
```diff
@@ -2323,102 +2315,139 @@ The "extension_data" field of this extension contains a
...
-  value refers solely to signatures which appear in certificates.
-  The use of DSA and anonymous is deprecated. Previous versions
-  of TLS supported DSA. DSA is deprecated as of TLS 1.3 and
-  SHOULD NOT be offered or negotiated by any implementation.
+       SignatureAlgorithm supported_signature_algorithms<2..2^16-2>;
+
+[[TODO: IANA considerations for new SignatureAlgorithm scheme]]
+
+[[TODO: It's unfortunate that this collides with the old one. Maybe
+SignatureScheme?]]
```
ekr: Yes I think that would be better. I will make the change when I merge

PR #406, opened 2016-01-23
Require RFC6979 deterministic ECDSA

ctz:
I think TLS1.3 should require implementations to use RFC6979 deterministic ECDSA. This eradicates the well-known flaw in ECDSA, and makes ECDSA k selection testable for correctness.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -797,7 +797,7 @@ of the digest output. Note that previous versions of TLS used RSASSA-PKCS1-v1_5,
 not RSASSA-PSS.
 
 All ECDSA computations MUST be performed according to ANSI X9.62 {{X962}}
-or its successors.  In addition, signatures must be deterministically made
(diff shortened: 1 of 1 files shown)


davegarrett:
I get the feeling from discussion on the mailing list that the WG is leaning towards consensus for a "SHOULD" not a "MUST", as this is an unverifiable & unenforceable requirement, and systems with true RNGs can actually do things properly. I suggest changing this PR to reflect this, including leaving the pitfalls section as it is in the current draft.

Mentioning TRNGs could be a good idea here, to differentiate that case from when doing things deterministically his highly preferred. e.g., something to the effect of:

"ECDSA signatures SHOULD be deterministically made using the scheme specified in {{RFC6979}} or generated using randomness provided by a secure hardware entropy source."

(not sure of the best way to word this)

A squash of all commits might also be a good idea.

ctz:
This PR is superseded by https://github.com/tlswg/tls13-spec/pull/408

PR #409, opened 2016-02-03
spell-out first occurances of PSK

equalsJeffH:
the PSK acronym occurs first in the TOC, and later in S 6 without being spelled-out -- these modest changes fix that. 

[ this is also a small test to see if I have the tlswg+github process figured out ]

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1241,7 +1241,7 @@ cipher spec
 
 resumption master secret
 : a secret shared between the client and server that can be used
-  as a PSK in future connections.
+  as a pre-shared symmetric key (PSK) in future connections.
 {:br }
 
 These items are then used to create security parameters for use by the record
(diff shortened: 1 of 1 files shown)


equalsJeffH:
ok, I see the problem -- tho i ought to install travis, kramdown, et al hereabouts to be able to try a build locally before pushing again to here.

davegarrett:
You don't need to install everything unless you're testing the build yourself, which usually isn't needed.

draft-ietf-tls-tls13.md
```diff
@@ -1825,7 +1825,7 @@ is compromised and client authentication is used, then the attacker
 can impersonate the client on subsequent connections.
 
 
-### Resumption and PSK
+### Resumption and Pre-Shared Key (PSK)
```
davegarrett: Changing this actually breaks the auto-generated anchor links. Rather than make the ref longer, please append to this line the following:
" {#resumption-and-psk}"
(without the quotes, but with the leading space after the new text; note, only 1 set of braces)

This will keep the same anchor name whilst allowing the title to be changed.

equalsJeffH:
Oh, sorry @davegarrett, i didn't see your suggestion above until after I'd figured out how to fix the one xref the section title had broken -- but the fix can be easily changed to how you suggest if you all wish. also thx for the hint wrt toolchain.

davegarrett:
I'd prefer just renaming the ref, as parens don't end up in the autogenerated one, so it looks a little odd. ("{{resumption-and-pre-shared-key-psk}}") If it works, though, I don't have a strong objection, as it doesn't really matter.

PR #423, opened 2016-02-23
purpose should be lowercase

hannesm:
It seems that both Karthik and I misinterpreted this "IV" to be uppercase, but it should be lowercase... Thus fixing this table so that others won't make the very same mistake..


ekr:
This does in fact seem like a bug in the spec.

PR #434, opened 2016-03-22
append "_RESERVED" to DSA SignatureScheme value names

davegarrett:
This appends a "_RESERVED" to each name for the DSA values in the SignatureScheme enum to have them hidden up top and only shown in the appendix, as they're obsolete and their full description was removed in a prior changeset (https://github.com/tlswg/tls13-spec/commit/bed72816a2cbcb2695718c3936c44b78498e07da).

The alternative would be to remove them completely and expand the obsolete_RESERVED ranges to encompass them (these ranges are already hidden up top). I think keeping them as separate reserved values in the appendix is a bit better.

Note that I have to delete this block's header comment as well, as the auto-pruning won't catch that.


ekr:
Thanks. Good fix.

PR #436, opened 2016-03-24
Patch some typos, missing words and punctuations

beurdouche:


beurdouche:
Re-pushed to squash commits...

draft-ietf-tls-tls13.md
```diff
@@ -1106,7 +1106,7 @@ and in TLS 1.3 the additional data input is empty (zero length).
 opaque_type
 : The outer opaque_type field of a TLSCiphertext record is always set to the
   value 23 (application_data) for outward compatibility with
-  middleboxes used to parsing previous versions of TLS.  The
```
davegarrett: The current line is correct English. If we would like to simplify the language here, the suggested replacement is not ideal (boxes will likely continue to parses old TLS).

draft-ietf-tls-tls13.md
```diff
@@ -3335,7 +3335,7 @@ and for a client signature is "TLS 1.3, client CertificateVerify". A
 hash of the handshake messages is signed rather than the messages themselves
 because the digitally-signed format requires padding and context bytes at the
 beginning of the input. Thus, by signing a digest of the messages, an
-implementation need only maintain one running hash per hash type for
+implementation only needs to maintain a single running hash per hash type for
```
davegarrett: Both versions are correct. The suggested replacement could be considered simpler phrasing, which looks fine to me.

draft-ietf-tls-tls13.md
```diff
@@ -3531,8 +3531,8 @@ the traffic keys (though not the traffic secret) for the previous
 generation of keys until it receives the KeyUpdate from the other
 side.
 
-Both sender and receiver must encrypt their KeyUpdate 
-messages with the old keys. Additionally, both sides MUST enforce that 
+Both sender and receiver must encrypt their KeyUpdate
```
davegarrett: While we're here, "must" should be capitalized (or simply dropped).

davegarrett:
Aside from the one noted change, lots of good typo fixes here.

draft-ietf-tls-tls13.md
beurdouche: "with middleboxes used to parse ..." ?

davegarrett: First Google hit for "used to": http://www.5minuteenglish.com/mar20.htm
In this context, "used to" is equivalent to "accustomed to".

The simplest change that would clarify this for those not familiar with this wording would be to just swap in "accustomed" for "used" with no other changes:

"middleboxes accustomed to parsing previous versions of TLS"

beurdouche:
Good to go.

PR #439, opened 2016-04-11
Add a section on 0-RTT backwards compatibility.

davidben:
0-RTT is *not* backwards compatible because of the external application data records. Add some guidance for both clients and servers to mitigate this. Ideally the server SHOULD could be a MUST and that would be the end of it, but I do not expect all servers to get this right and so clients will want a fallback.

This also lets us provide guidance on the exact client-side fallback trigger and explicitly say not to change the version. Conveniently, there is a reliable signal available. (Historically, clients have used network errors or generic protocol errors as fallback triggers which masks new server bugs.)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4324,28 +4324,22 @@ MUST always be ignored.
 
 ## Zero-RTT backwards compatibility
 
-A TLS 1.3 ClientHello with 0-RTT data is not compatible with older servers
-implementing previous versions of TLS. This can cause compatibility issues if a
-client first connects to a 0-RTT-capable TLS 1.3 server and later to an older
-server, both behind the same service. For example, a multi-server deployment
-may deploy TLS 1.3 gradually with some servers implementing TLS 1.3 and some
-implementing TLS 1.2. This may also occur if a TLS 1.3 deployment gets reverted
-to TLS 1.2.
+A TLS 1.3 ClientHello with 0-RTT data is not compatible with older servers.
+This can cause compatibility issues if a client attempts to use 0-RTT with a
+server cluster that contains servers without TLS 1.3 support. For example, a
+multi-server deployment may deploy TLS 1.3 gradually with some servers
+implementing TLS 1.3 and some implementing TLS 1.2. This may also occur if a
+TLS 1.3 deployment gets reverted to TLS 1.2.
 
 If a TLS 1.3 client sends a ClientHello with 0-RTT to an older server, the
-server will respond with a ServerHello containing an older version number.
-(The handshake will later fail at the 0-RTT data.) Clients which accept older
-versions of TLS SHOULD then retry the connection advertising TLS 1.3 without
-offering 0-RTT.
+server will respond with a ServerHello containing an older version number. The
+handshake will later fail at the 0-RTT data. Clients which accept older
+versions of TLS MAY retry the connection without 0-RTT in response to this
+ServerHello. It is NOT RECOMMENDED to retry the connection in response to a
+more generic error or advertise lower versions of TLS.
 
-Note that it is not necessary to retry based on network error or arbitrary
-protocol error. It is also not necessary to advertise lower versions of
-TLS.
-
-Services in transition from an older TLS version SHOULD ensure that all older
-servers have upgraded to TLS 1.3 without 0-RTT and wait before there is minimal
-chance of rollback before offering 0-RTT on any server.
-
+Services with multiple servers SHOULD ensure a stable deployment of TLS 1.3
+without 0-RTT prior to enabling 0-RTT.
 
 ## Backwards Compatibility Security Restrictions
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4285,6 +4285,9 @@ client agrees to use this version, the negotiation will proceed as appropriate
 for the negotiated protocol. A client resuming a session SHOULD initiate the
 connection using the version that was previously negotiated.
 
+Note that a 0-RTT TLS 1.3 ClientHello is not compatible with older servers.
+See {{zero-rtt-backwards-compatibility}}.
+
 If the version chosen by the server is not supported by the client (or not
 acceptable), the client MUST send a "protocol_version" alert message and close
 the connection.
@@ -4317,6 +4320,31 @@ Note that earlier versions of TLS did not clearly specify the record layer
 version number value in all cases (TLSPlaintext.record_version). Servers
 will receive various TLS 1.x versions in this field, however its value
 MUST always be ignored.
+
+
+## Zero-RTT backwards compatibility
+
+A TLS 1.3 ClientHello with 0-RTT data is not compatible with older servers
+implementing previous versions of TLS. This can cause compatibility issues if a
+client first connects to a 0-RTT-capable TLS 1.3 server and later to an older
+server, both behind the same service. For example, a multi-server deployment
+may deploy TLS 1.3 gradually with some servers implementing TLS 1.3 and some
+implementing TLS 1.2. This may also occur if a TLS 1.3 deployment gets reverted
+to TLS 1.2.
+
+If a TLS 1.3 client sends a ClientHello with 0-RTT to an older server, the
+server will respond with a ServerHello containing an older version number.
+(The handshake will later fail at the 0-RTT data.) Clients which accept older
+versions of TLS SHOULD then retry the connection advertising TLS 1.3 without
+offering 0-RTT.
+
+Note that it is not necessary to retry based on network error or arbitrary
+protocol error. It is also not necessary to advertise lower versions of
+TLS.
+
+Services in transition from an older TLS version SHOULD ensure that all older
+servers have upgraded to TLS 1.3 without 0-RTT and wait before there is minimal
+chance of rollback before offering 0-RTT on any server.
 
 
 ## Backwards Compatibility Security Restrictions
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4338,8 +4338,8 @@ versions of TLS MAY retry the connection without 0-RTT in response to this
 ServerHello. It is NOT RECOMMENDED to retry the connection in response to a
 more generic error or advertise lower versions of TLS.
 
-Services with multiple servers SHOULD ensure a stable deployment of TLS 1.3
-without 0-RTT prior to enabling 0-RTT.
+Services with multiple servers are RECOMMENDED to ensure a stable deployment of
+TLS 1.3 without 0-RTT prior to enabling 0-RTT.
 
 ## Backwards Compatibility Security Restrictions
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4285,7 +4285,7 @@ client agrees to use this version, the negotiation will proceed as appropriate
 for the negotiated protocol. A client resuming a session SHOULD initiate the
 connection using the version that was previously negotiated.
 
-Note that a 0-RTT TLS 1.3 ClientHello is not compatible with older servers.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,31 @@ will receive various TLS 1.x versions in this field, however its value
 MUST always be ignored.
 
 
+## Zero-RTT backwards compatibility
+
+A TLS 1.3 ClientHello with 0-RTT data is not compatible with older servers
+implementing previous versions of TLS. This can cause compatibility issues if a
```
martinthomson: s/older servers/servers/  -- this is redundant with "implementing previous versions"

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,31 @@ will receive various TLS 1.x versions in this field, however its value
...
+client first connects to a 0-RTT-capable TLS 1.3 server and later to an older
+server, both behind the same service. For example, a multi-server deployment
```
martinthomson: This can cause compatibility issues if a client attempts to use 0-RTT with a server cluster that contains servers without TLS 1.3 support.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,31 @@ will receive various TLS 1.x versions in this field, however its value
...
+may deploy TLS 1.3 gradually with some servers implementing TLS 1.3 and some
+implementing TLS 1.2. This may also occur if a TLS 1.3 deployment gets reverted
+to TLS 1.2.
+
+If a TLS 1.3 client sends a ClientHello with 0-RTT to an older server, the
+server will respond with a ServerHello containing an older version number.
+(The handshake will later fail at the 0-RTT data.) Clients which accept older
```
martinthomson: Remove parentheses.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,31 @@ will receive various TLS 1.x versions in this field, however its value
...
+versions of TLS SHOULD then retry the connection advertising TLS 1.3 without
+offering 0-RTT.
```
martinthomson: Oooo, that's nasty.  Are you advocating the fallback dance?  I guess that it's safe now.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,31 @@ will receive various TLS 1.x versions in this field, however its value
...
+
+Note that it is not necessary to retry based on network error or arbitrary
+protocol error. It is also not necessary to advertise lower versions of
+TLS.
+
+Services in transition from an older TLS version SHOULD ensure that all older
+servers have upgraded to TLS 1.3 without 0-RTT and wait before there is minimal
+chance of rollback before offering 0-RTT on any server.
```
martinthomson: Services with multiple servers SHOULD ensure a stable deployment of TLS 1.3 without 0-RTT prior to enabling 0-RTT.

draft-ietf-tls-tls13.md
davidben: Not the same one as before. Rather than falling back on generic error and insecurely dropping the version, this is just a retry on a particular signal and punting on 0-RTT. The ClientHello version should be unchanged. (Though the old fallback is probably going to come back too. I probed for 1.3 intolerance way back and it's everywhere. :-( )

Agreed that this is incredibly nasty. I can't think of a better solution though, assuming we want to continue allowing streaming 0-RTT. I think it's unrealistic to expect that servers will get the deployment considerations correct. (I'm thinking people who just install their favorite HTTP server from their favorite OS distribution and put a few of them behind a hostname.)

Given that, I think it's better that we spell out the fallback we want clients to implement so they don't make worse mistakes (dropping ClientHello.version or using an overly broad fallback signal that masks future implementation bugs).

martinthomson: Yeah, I guess we have to decide whether we want to put that in this document, or just do it quietly so that we can hide our shame.

davegarrett: Please change this "SHOULD" to a "MAY", as I think that's more appropriate. If a client sends a hello to a server it knows supports 0RTT, yet it borks on it, it should not always be expected to retry as this could easily be considered evidence that it's not connecting to the server it thinks it is. (e.g. a controlled environment where a known set of clients are interoperating with a known set of servers) If the client wants to fail here, that's up to the implementers. Some will generally want to retry; others could be fine with failing. Going into more detail would be outside the scope of this document.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,31 @@ will receive various TLS 1.x versions in this field, however its value
...
+implementing TLS 1.2. This may also occur if a TLS 1.3 deployment gets reverted
+to TLS 1.2.
+
+If a TLS 1.3 client sends a ClientHello with 0-RTT to an older server, the
+server will respond with a ServerHello containing an older version number.
+(The handshake will later fail at the 0-RTT data.) Clients which accept older
+versions of TLS SHOULD then retry the connection advertising TLS 1.3 without
+offering 0-RTT.
+
+Note that it is not necessary to retry based on network error or arbitrary
```
davegarrett: "NOT RECOMMENDED" instead of "not necessary" makes a bit more sense here, I think. (caps just because of RFC 2119, not that I think the point particularly needs stressing) I'd also merge the two sentences to "[...] protocol error, or advertise lower versions of TLS." and stick this just as the last sentence of the previous paragraph.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,31 @@ will receive various TLS 1.x versions in this field, however its value
 MUST always be ignored.
 
 
+## Zero-RTT backwards compatibility
```
davegarrett: This should probably be a subsection under "Negotiating with an older server". (add another '#' and move this all to before "Negotiating with an older client") Also negates the need for the redundant language added to that section.

draft-ietf-tls-tls13.md
```diff
@@ -4285,6 +4285,9 @@ client agrees to use this version, the negotiation will proceed as appropriate
 for the negotiated protocol. A client resuming a session SHOULD initiate the
 connection using the version that was previously negotiated.
 
+Note that a 0-RTT TLS 1.3 ClientHello is not compatible with older servers.
+See {{zero-rtt-backwards-compatibility}}.
```
davegarrett: I think this is excessive redundancy. (see comment below on sticking the new text under this section as a subsection)

draft-ietf-tls-tls13.md
davidben: Removed "implementing previous versions ..." and kept "older" to be consistent with the other section.

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md
davidben: Tweaked the text a bit.

draft-ietf-tls-tls13.md
davidben: I considered that, but this section contains both client and server advice. Splitting them up is confusing because the reasoning is the same (and the server advice does not fall under "Negotiating with an older client"). I think it's best in its own section because this is a peculiar beast we have written ourselves into.

draft-ietf-tls-tls13.md
davidben: (See response in other bit.)

davegarrett:
Your rationale for sectioning like this does make sense; fair enough. Updated text looks good.

One issue, though: use of "SHOULD" with regard to upgrade deployment should probably be changed to a "RECOMMENDED". I agree with the statement, but multi-server rollout of updates and new features is rather outside the scope of this document. It's the same level of requirement per RFC2119 language, but I think the later is technically preferred in this context. (might just be my interpretation; not a strong objection)

davidben:
Done. I know little of RFC2119 minutia and will defer to you all on that. :-)

draft-ietf-tls-tls13.md
ekr: Technically it's not the ClientHello that's not compatible but rather everything afterwards.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,25 @@ will receive various TLS 1.x versions in this field, however its value
 MUST always be ignored.
 
 
+## Zero-RTT backwards compatibility
+
+A TLS 1.3 ClientHello with 0-RTT data is not compatible with older servers.
```
ekr: I think it would be helpful to explain what the problem is here. It's not the ClientHello but the need to skip past the ciphertext.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,25 @@ will receive various TLS 1.x versions in this field, however its value
...
+This can cause compatibility issues if a client attempts to use 0-RTT with a
```
ekr: maybe just "issues" since the previous sentence told us they were compat issues.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,25 @@ will receive various TLS 1.x versions in this field, however its value
...
+server cluster that contains servers without TLS 1.3 support. For example, a
+multi-server deployment may deploy TLS 1.3 gradually with some servers
+implementing TLS 1.3 and some implementing TLS 1.2. This may also occur if a
```
ekr: Nit: I would merge these sentences, as in "or, a TLS 1.3 ..."

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,25 @@ will receive various TLS 1.x versions in this field, however its value
...
+TLS 1.3 deployment gets reverted to TLS 1.2.
+
+If a TLS 1.3 client sends a ClientHello with 0-RTT to an older server, the
+server will respond with a ServerHello containing an older version number. The
+handshake will later fail at the 0-RTT data. Clients which accept older
```
ekr: Maybe say how it will fail.

draft-ietf-tls-tls13.md
```diff
@@ -4319,6 +4322,25 @@ will receive various TLS 1.x versions in this field, however its value
...
+versions of TLS MAY retry the connection without 0-RTT in response to this
+ServerHello. It is NOT RECOMMENDED to retry the connection in response to a
+more generic error or advertise lower versions of TLS.
```
ekr: Maybe elaborate on the security properties here. Specifically, if you negotiate an ephemeral cipher suite, then the signature from the server should validate the lack of downgrade.

draft-ietf-tls-tls13.md
davidben: Tweaked.

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md
davidben: (Done as part of changes to earlier bits.)

draft-ietf-tls-tls13.md
davidben: I'm not sure I follow. Do you mean security properties of the 0-RTT retry or of advertising lower versions of TLS? I'm saying not to do that and to retry at TLS 1.3. (Of course, all us browsers will certainly do a version fallback too. It seems 1.3 intolerance is rampant. But that's another matter and, yes, we have the server_random signaling and whatnot to help there.)

I don't know of any particular security considerations for this retry. I guess it's another vector for 0-RTT data to be replayable, but we're already accepting that it is, right? (Hrm. Actually... this would defeat any server-side stateful 0-RTT replay protection... I'm dubious anyone would bother with that, but still.)

PR #444, opened 2016-04-27
Remove 0-RTT (EC)DHE and 0-RTT client authentication

ekr:
Not ready for landing, but for @martinthomson initial review.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -114,7 +114,6 @@ informative:
   RFC6091:
   RFC6520:
   RFC6961:
-  RFC6962:
   RFC7301:
   RFC7250:
   RFC7366:
@@ -286,10 +285,10 @@ receives its first byte of data. The TLS Handshake Protocol provides connection
 security that has three basic properties:
 
 
-- The peer's identity can be authenticated using asymmetric, or
-  public key, cryptography (e.g., RSA {{RSA}}, ECDSA {{ECDSA}}).  This
-  authentication can be made optional, but is generally required for
-  at least one of the peers.
+- The peer's identity can be authenticated using asymmetric (public key)
+  cryptography (e.g., RSA {{RSA}}, ECDSA {{ECDSA}}) or a pre-shared
+  symmetric key. The TLS server is always authenticated; client authentication
+  is optional.
 
 - The negotiation of a shared secret is secure: the negotiated
   secret is unavailable to eavesdroppers, and for any authenticated
@@ -340,6 +339,14 @@ server: The endpoint which did not initiate the TLS connection.
 draft-13
 
 - Allow server to send SupportedGroups.
+
+- Remove 0-RTT client authentication
+
+- Remove (EC)DHE 0-RTT.
+
+- Flesh out 0-RTT PSK mode and shrink EarlyDataIndiation
+
+- Turn PSK-resumption response into an index to save room.
 
 draft-12
 
@@ -1578,23 +1585,29 @@ Static Secret (SS): A secret which may be derived from static or
 Master Secret (MS): A secret derived from both the static
    and the ephemeral secret.
 
-In some cases, as with the DH handshake shown in {{tls-full}},
-the ephemeral and shared
-secrets are the same, but having both allows for a uniform key
-derivation scheme for all cipher modes.
+There are three basic cases:
 
-The basic TLS Handshake for DH is shown in {{tls-full}}:
+- In a pure (EC)DHE cipher suite, ES and SS will both be
+  the (EC)DHE shared secret.
+  
+- In a pure PSK cipher suite, ES and SS will both be the PSK.
+
+- In an (EC)DHE-PSK cipher suite, ES will be the (EC)DHE
+  shared secret and SS will be the PSK.
+
+{{tls-full}} below shows the basic full TLS handshake.
 
 ~~~
        Client                                               Server
 
 Key  / ClientHello
-Exch \  + key_share              -------->
-                                                       ServerHello  \ Key
-                                                       + key_share  / Exch
-                                             {EncryptedExtensions}  ^
-                                             {CertificateRequest*}  | Server
-                                            {ServerConfiguration*}  v Params
+Exch<  + key_share*
+     \ + pre_shared_key*        -------->
+                                                       ServerHello  \  Key
+                                                      + key_share*   > Exch
+                                                 + pre_shared_key*  /
+                                             {EncryptedExtensions}  ^  Server
+                                             {CertificateRequest*}  v  Parameters
                                                     {Certificate*}  ^
                                               {CertificateVerify*}  | Auth
                                                         {Finished}  v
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1574,23 +1583,29 @@ Static Secret (SS): A secret which may be derived from static or
...
 
 Key  / ClientHello
-Exch \  + key_share              -------->
-                                                       ServerHello  \ Key
-                                                       + key_share  / Exch
-                                             {EncryptedExtensions}  ^
-                                             {CertificateRequest*}  | Server
-                                            {ServerConfiguration*}  v Params
+Exch<  + key_share*
+     \ + pre_shared_key*        -------->
```
martinthomson: BTW your notation for these things is inconsistent.

```
^
|
v
```
OR
```
 /
<
 \
```

draft-ietf-tls-tls13.md
```diff
@@ -1622,29 +1637,31 @@ Key Exchange: establish shared keying material and select the
...
-message, which contains
-a random nonce (ClientHello.random), its offered protocol version,
-cipher suite, and extensions, and one or more Diffie-Hellman key
-shares in the "key_share" extension {{key-share}}.
+In the Key Exchange phase, the client sends the ClientHello
+({{client-hello}}) message, which contains a random nonce
+(ClientHello.random), its offered protocol version, cipher suite, and
+extensions, and in general either one or more Diffie-Hellman key shares (in the
+"key_share" extension {{key-share}}), one or more pre-shared key labels (in the
+"pre_shared_key" extension {{pre-shared-key-extension}}), or both.
```
martinthomson: The ClientHello also includes the client's parameters.

draft-ietf-tls-tls13.md
```diff
@@ -1622,29 +1637,31 @@ Key Exchange: establish shared keying material and select the
...
-The server then sends three messages to establish the Server Parameters:
+its own ServerHello which indicates the negotiated connection parameters. [{{server-hello}}].
+The combination of the ClientHello and the ServerHello determines
+the values of ES and SS, as described above. If either a pure
+(EC)DHE or (EC)DHE-PSK cipher suite is in use, then the ServerHello
+will will contain a "key_share" extension with the server's ephemeral
+Diffie-Hellman share which MUST be in the same group.
+If a pure PSK or an (EC)DHE-PSK cipher suite is negotiated, then the
+ServerHello will contain a "pre_shared_key" extension indicating
+which if the client's offered PSKs was selected.
```
martinthomson: I like how this works out:

if (ECDHE) add(key_share)
if (PSK) add(pre_shared_key)

draft-ietf-tls-tls13.md
```diff
@@ -1657,23 +1674,20 @@ CertificateRequest
...
-
 Finally, the client and server exchange Authentication messages. TLS
 uses the same set of messages every time that authentication is needed.
 Specifically:
 
 Certificate
-: the certificate of the endpoint. This message is omitted if certificate
-  authentication is not being used. [{{certificate}}]
+: the certificate of the endpoint. This message is omitted if the
+  server is not authenticating via a certificate (i.e.,
```
martinthomson: s/via/with/

Also, cached-info and bare public keys.

draft-ietf-tls-tls13.md
```diff
@@ -1707,14 +1721,16 @@ a 255-bit ECDHE key exchange with a host whose certificate
...
 against algorithmic attacks, at least in the year 2015.]]
 
-
 ### Incorrect DHE Share
 
 If the client has not provided an appropriate "key_share" extension (e.g. it
 includes only DHE or ECDHE groups unacceptable or unsupported by the
 server), the server corrects the mismatch with a HelloRetryRequest and
 the client will need to restart the handshake with an appropriate
 "key_share" extension, as shown in Figure 2:
```
martinthomson: Trailing colon here.  Might need to be restructured.

draft-ietf-tls-tls13.md
```diff
@@ -1906,10 +1823,78 @@ Subsequent Handshake:
 {: #tls-resumption-psk title="Message flow for resumption and PSK"}
 
 As the server is authenticating via a PSK, it does not send a
-Certificate or a CertificateVerify. PSK-based resumption cannot be
-used to provide a new ServerConfiguration. Note that the client
-supplies a KeyShare to the server as well, which allows the
-server to decline resumption and fall back to a full handshake.
+Certificate or a CertificateVerify. When a client offers resumption
+via PSK it SHOULD also supply a KeyShare to the server as well; this
```
martinthomson: KeyShare or key_share?  I thought the latter.

draft-ietf-tls-tls13.md
```diff
@@ -1906,10 +1823,78 @@ Subsequent Handshake:
...
+                  derived from the ephemeral secret.
+
+               [] Indicates messages protected using keys
+                  derived from the master secret.
+~~~
+{: #tls-0-rtt title="Message flow for a zero round trip handshake"}
+
+[[OPEN ISSUE: Should it be possible to combine 0-RTT with the
+server authenticating via a signature
+https://github.com/tlswg/tls13-spec/issues/443]]
```
martinthomson: Yes, and I think that this is easy (just use the right cipher suites).

draft-ietf-tls-tls13.md
```diff
@@ -1906,10 +1823,78 @@ Subsequent Handshake:
...
+
+2. There are no guarantees of non-replay between connections.
+Unless the server takes special measures outside those provided by TLS (See
+{{replay-properties}}), the server has no guarantee that the same
+0-RTT data was not transmitted on multiple 0-RTT connections.
+This is especially relevant if the data is authenticated either
+with TLS client authentication or inside the application layer
+protocol. However, 0-RTT data cannot be duplicated within a connection (i.e., the server
+will not process the same data twice for the same connection) and also
+cannot be sent as if it were ordinary TLS data.
```
martinthomson: What is the point of the last bit here? "also cannot be sent as if it were ordinary TLS data".

Is this a reference to API requirements?

draft-ietf-tls-tls13.md
```diff
@@ -1986,9 +1969,7 @@ ServerHello that selects cryptographic parameters that don't match the
 client's "key_share" extension. In that case, the client MUST send the same
 ClientHello (without modification) except including a new KeyShareEntry
 as the lowest priority share (i.e., appended to the list of shares in
-the KeyShare message). [[OPEN ISSUE: New random values? See:
-https://github.com/tlswg/tls13-spec/issues/185]]
-If a server receives a ClientHello at any other time, it MUST send
+the KeyShare message). If a server receives a ClientHello at any other time, it MUST send
```
martinthomson: As above, KeyShare or key_share?

draft-ietf-tls-tls13.md
```diff
@@ -2682,12 +2663,17 @@ The "extension_data" field of this extension contains a
...
 
                case server:
-                   psk_identity identity;
+                   uint16 selected_identity;
            }
        } PreSharedKeyExtension;
 
-identity
-: An opaque label for the pre-shared key.
+identities
```
martinthomson: Would it make sense to use a different name for these.  It's a PSK identifier.  Using the word "identity" might still be a good idea to trigger the right privacy-related immune responses though.

draft-ietf-tls-tls13.md
```diff
@@ -2720,57 +2705,30 @@ The "extension_data" field of this extension contains an
...
 application_data, and alert respectively) but are protected under
 different keys. After all the 0-RTT application data messages (if
 any) have been sent, a "end_of_early_data" alert of type
 "warning" is sent to indicate the end of the flight.
-Clients which do 0-RTT MUST always send "end_of_early_data" even
-if the ServerConfiguration indicates that no application
-data is allowed (EarlyDataType of "client_authentication"),
-though in that case it MUST NOT send any non-empty data records
-(i.e., those which consist of anything other than padding).
+Clients which do 0-RTT MUST always send "end_of_early_data".
```
martinthomson: This could be read as having to send the alert on every connection.

Maybe "0-RTT data MUST be followed by an "end_of_early_data" alert."

draft-ietf-tls-tls13.md
```diff
@@ -2783,20 +2741,21 @@ can behave in one of two ways:
...
-  rest of the ClientHello. If (EC)DHE is selected, this includes
-  verifying that (1) the ClientHello contains a key from the
-  same group that is indicated by the server configuration and
-  (2) that the server has negotiated that group and will
-  therefore include a share from that group in its own
-  "key_share" extension.
+- The "padding" extension {{RFC7685}} MUST be ignored for this purpose.
 
+- The values of "key_share", "pre_shared_key", and "early_data", which MUST
+  be as defined in this document.
```
martinthomson: Is the above list the result of a complete analysis.  For instance, I don't know SRP, but is there any sense that the srp extension needs to be the same?

For something I do know, it seems like signature_algorithms could be different without affecting security.

martinthomson: That said, if the advice is "if you change anything, then don't resume", that's a reasonable position to take.

draft-ietf-tls-tls13.md
```diff
@@ -3108,24 +2960,14 @@ for each scenario:
 
 | Mode | Handshake Context | Base Key |
 |------|-------------------|----------|
-| 0-RTT | ClientHello + ServerConfiguration + Server Certificate + CertificateRequest (where ServerConfiguration, etc. are from the previous handshake) | xSS |
-| 1-RTT (Server) | ClientHello ... ServerConfiguration | master secret |
+| 0-RTT | ClientHello | xSS |
+| 1-RTT (Server) | ClientHello ... CertificateVerify | master secret |
```
martinthomson: Is this right?  I think that you mean CertificateRequest/EncryptedExtensions.

martinthomson:
A bunch of nits, though some of these are on existing text.

draft-ietf-tls-tls13.md
ekr: Sorry, which parameters? It says cipher suite, etc. above.

draft-ietf-tls-tls13.md
ekr: Agreed, I just wanted to do this in pieces.

draft-ietf-tls-tls13.md
ekr: This is what the PSK document used so I kept it. I think I'll leave for nwo

draft-ietf-tls-tls13.md
ekr: I'm adopting a conservative position.

draft-ietf-tls-tls13.md
ekr: No, not presently. See S 7.1

martinthomson: Assuming that's [Key Schedule](https://tlswg.github.io/tls13-spec/#key-schedule), I have no idea what you are talking about.

My point was that the CertificateVerify and Finished already already Certificate, and Finished includes CertificateVerify.  They can't depend on themselves.

PR #445, opened 2016-04-27
Enhanced new session ticket

ekr:
@martinthomson this builds on the previous PR.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -114,7 +114,6 @@ informative:
   RFC6091:
   RFC6520:
   RFC6961:
-  RFC6962:
   RFC7301:
   RFC7250:
   RFC7366:
@@ -286,10 +285,10 @@ receives its first byte of data. The TLS Handshake Protocol provides connection
 security that has three basic properties:
 
 
-- The peer's identity can be authenticated using asymmetric, or
-  public key, cryptography (e.g., RSA {{RSA}}, ECDSA {{ECDSA}}).  This
-  authentication can be made optional, but is generally required for
-  at least one of the peers.
+- The peer's identity can be authenticated using asymmetric (public key)
+  cryptography (e.g., RSA {{RSA}}, ECDSA {{ECDSA}}) or a pre-shared
+  symmetric key. The TLS server is always authenticated; client authentication
+  is optional.
 
 - The negotiation of a shared secret is secure: the negotiated
   secret is unavailable to eavesdroppers, and for any authenticated
@@ -340,6 +339,14 @@ server: The endpoint which did not initiate the TLS connection.
 draft-13
 
 - Allow server to send SupportedGroups.
+
+- Remove 0-RTT client authentication
+
+- Remove (EC)DHE 0-RTT.
+
+- Flesh out 0-RTT PSK mode and shrink EarlyDataIndiation
+
+- Turn PSK-resumption response into an index to save room.
 
 draft-12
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3290,10 +3291,31 @@ suite. Servers may send multiple tickets on a single connection, for
...
+
+     enum { (65535) } TicketExtensionType;
+
+     struct {
+         TicketExtensionType extension_type;
+         opaque extension_data<0..2^16-1>;
+     } TicketExtension;
+
+     enum {
+       allow_dhe_resumption(1),
```
martinthomson: wots dis?

martinthomson: Oh, nevermind.  I get it.

draft-ietf-tls-tls13.md
```diff
@@ -3306,19 +3328,36 @@ ticket_lifetime
...
+  suite
+
+allow_psk_resumption
+: When resuming this session, the client MAY offer a pure PSK cipher
+  suite.
+
+allow_early_data
+: When resuming this session, the client MAY send data in its
+first flight (early data) encrypted under a key derived from
+this PSK.
```
martinthomson: Do we need a third flag: when resuming the client MAY offer a PSK+signature cipher suite?

draft-ietf-tls-tls13.md
```diff
@@ -3306,19 +3328,36 @@ ticket_lifetime
...
+{:br }
 
+The meanings of the flags are as follows:
 
-[[TODO: Should we require that tickets be bound to the existing
-symmetric cipher suite. See the TODO above about early_data and
-PSK.??]
+allow_dhe_resumption
+: When resuming this session, the client MAY offer an (EC)DHE-PSK cipher
+  suite
```
martinthomson: I think that this formulation isn't quite right.  It's not that the client MAY offer these cipher suites, because that is always possible.  It is "this ticket can be used with PSK cipher suites that have the following properties: PSK-only key exchange, PSK+DHE key exchange, (and later) PSK+signature authentication."

draft-ietf-tls-tls13.md
```diff
@@ -3306,19 +3328,36 @@ ticket_lifetime
...
+
+allow_early_data
+: When resuming this session, the client MAY send data in its
+first flight (early data) encrypted under a key derived from
+this PSK.
+{:br }
 
+In all cases, the PSK or (EC)DHE-PSK cipher suites that the client
+offers/uses MUST have the same symmetric parameters (cipher/hash) as
+the cipher suite negotiated for this connection.
```
martinthomson: You need to point out what the client does with a ticket that doesn't include any of the cipher-suite flags.

draft-ietf-tls-tls13.md
martinthomson: Also, I would move allow_early_data first.  It's special.

draft-ietf-tls-tls13.md
ekr: I think we will once we figure out what the exact semantics are. The issue here is I don't want to add this function until we specify how it works.

draft-ietf-tls-tls13.md
ekr: Sorry I don't understand this comment. Can you elaborate?

martinthomson: If a server provides a ticket and it doesn't include allow_dhe_resumption (should that be allow_dhe_psk_resumption?) OR allow_psk_resumption.  Then the ticket is useless to it.  Is that something that we want the client to generate an error for?

Or is it possible that new flags might be added without prior arrangement, in which case the client should just ignore tickets that it can't use.

I think the latter, but that means we have an extension point here.  Which requires a little care.

PR #446, opened 2016-04-27
Allow server to send supported_groups. Fixes #415

### LINKED ISSUE #415: Allow servers to send KnownGroups
So that the clients can learn about what groups a server supports.

The idea here is to allow a client to offer P256 but then learn that
a server supports 25519 for next time without the server having to
reject.

ekr:


davegarrett:
You'll also need to update the last paragraph of the MTI Extensions section (currently section 8.2). Now that we're going to be down to only one client-only extension defined in this document, that whole paragraph can probably be cut in favor of a one-off note in the remaining extension's section (signature_algorithms will be the last).

draft-ietf-tls-tls13.md
```diff
@@ -2525,9 +2525,19 @@ extension type ("supported_groups" extension):
...
        00 0A 00 06 00 04 00 17 00 18
 
+
+As of TLS 1.3, servers are permitted to send the "supported_groups"
+extension to the client.  If the server has a group it prefers to the
+ones in the "key_share" extension but is still willing to accept the
+ClientHello, it SHOULD send "supported_groups" to update the client's
+view of its preferences.  Clients MUST NOT act upon any information
+found in "supported_groups" prior to successful completion of the
+handshake.
```
martinthomson: ...but they MAY use the information learned from a successfully completed handshake to change what groups they offer to a server in subsequent connections.

PR #448, opened 2016-05-02
Move CertificateStatus to an extension

ekr:


martinthomson:
I think that you should mark this document as updating 6066 and 6961.

PR #449, opened 2016-05-03
Encrypted extensions for client

ekr:
WIP, not ready to land. @martinthomson PTAL

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2801,14 +2801,14 @@ an accepted "early_data" extension MUST produce a fatal
 Implementations SHOULD determine the security parameters for the
 1-RTT phase of the connection entirely before processing the EncryptedExtensions
 and Finished, using those values solely to determine whether to
-accept or reject the "early_data" extension.
+accept or reject 0-RTT data.
 
 [[TODO: How does the client behave if the indication is rejected.]]
 
 ##### Processing Order
 
 Clients are permitted to "stream" 0-RTT data until they
-receive the ServerHello, only then sending the "end_of_early_data"
+receive the server's Finished, only then sending the "end_of_early_data"
 alert. In order to avoid deadlock, when accepting "early_data",
 servers MUST process the client's Finished and then immediately
 send the ServerHello, rather than waiting for the client's
@@ -2900,7 +2900,9 @@ extensions and if any are found MUST terminate the handshake with an
 
 The client's EncryptedExtensions apply only to the early data
 with which they appear. Servers MUST NOT use them to negotiate
-the rest of the handshake.
+the rest of the handshake. Only those extensions explicitly
+designated as being included in 0-RTT Encrypted Extensions
+in the IANA registry can be sent in the client's EncryptedExtensions.
 
 Structure of this message:
 
@@ -3815,7 +3817,7 @@ is listed below:
    that the server shall not send them. "Clear", indicating
    that they shall be in the ServerHello. "Encrypted", indicating that
    they shall be in the EncryptedExtensions block, "Early", indicating
-   that they shall be in the client's 0-RTT EncryptedExtensions block,
+   that they shall be only in the client's 0-RTT EncryptedExtensions block,
    and "No" indicating
    that they are not used in TLS 1.3. This column [shall be/has been]
    initially populated with the values in this document.
@@ -3856,7 +3858,7 @@ is listed below:
 | key_share [[this document]]              |         Yes |     Clear |
 | pre_shared_key [[this document]]         |         Yes |     Clear |
 | early_data [[this document]]             |         Yes |     Clear |
-| 
+| ticket_age [[this document]]             |         Yes |     Early |
 
 
 In addition, this document defines two new registries to be maintained
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1851,6 +1851,7 @@ in the first flight, the rest of the handshake uses the same messages.
          ClientHello
            + early_data
            + key_share*
+         (EncryptedExtensions)
          (Finished)
          (Application Data*)
          (end_of_early_data)        -------->
@@ -1895,7 +1896,7 @@ with the PSK.
 
 2. There are no guarantees of non-replay between connections.
 Unless the server takes special measures outside those provided by TLS (See
-{{replay-properties}}), the server has no guarantee that the same
+{{replay-time}}), the server has no guarantee that the same
 0-RTT data was not transmitted on multiple 0-RTT connections.
 This is especially relevant if the data is authenticated either
 with TLS client authentication or inside the application layer
@@ -2242,6 +2243,7 @@ The extension format is:
            key_share(40),
            pre_shared_key(41),
            early_data(42),
+           ticket_age(43),
            (65535)
        } ExtensionType;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2795,13 +2800,25 @@ specified for all records when processing early data records.
 Specifically, decryption failure of any 0-RTT record following
 an accepted "early_data" extension MUST produce a fatal
 "bad_record_mac" alert as per {{record-payload-protection}}.
+Implementations SHOULD determine the security parameters for the
+1-RTT phase of the connection entirely before processing the EncryptedExtensions
+and Finished, using those values solely to determine whether to
+accept or reject the "early_data" extension.
```
martinthomson: It's not the extension that is being accepted or rejected, it's the 0-RTT data.

draft-ietf-tls-tls13.md
```diff
@@ -2795,13 +2800,25 @@ specified for all records when processing early data records.
...
 
+##### Processing Order
 
-##### Replay Properties
+Clients are permitted to "stream" 0-RTT data until they
+receive the ServerHello, only then sending the "end_of_early_data"
+alert. In order to avoid deadlock, when accepting "early_data",
+servers MUST process the client's Finished and then immediately
+send the ServerHello, rather than waiting for the client's
+"end_of_early_data" alert.
```
martinthomson: Is the `ServerHello` the only message that a client needs in order to switch to application data keys?  Because otherwise we should allow clients to stream until they receive the server's `Finished`.

draft-ietf-tls-tls13.md
```diff
@@ -2810,32 +2827,82 @@ configuration identifier for each connection, thus preventing
...
+in {{server-hello}} or designated in the IANA registry MUST only
+appear in EncryptedExtensions. Extensions which are designated to
+appear in ServerHello MUST NOT appear in EncryptedExtensions. Clients
+MUST check EncryptedExtensions for the presence of any forbidden
+extensions and if any are found MUST terminate the handshake with an
+"illegal_parameter" alert.
+
+The client's EncryptedExtensions apply only to the early data
+with which they appear. Servers MUST NOT use them to negotiate
+the rest of the handshake.
```
martinthomson: Maybe add, "Only those extensions explicitly designated as being included in 0-RTT Encrypted Extensions in the IANA registry can be sent in the client's EncryptedExtensions."

I think that the check  above can be made mandatory for both client and server.  I have no problem tearing a connection down if 0-RTT includes bad extensions, though rejecting 0-RTT might be less disruptive.

draft-ietf-tls-tls13.md
```diff
@@ -3787,7 +3854,9 @@ is listed below:
    1.3" column with the following four values: "Client", indicating
    that the server shall not send them. "Clear", indicating
    that they shall be in the ServerHello. "Encrypted", indicating that
-   they shall be in the EncryptedExtensions block, and "No" indicating
+   they shall be in the EncryptedExtensions block, "Early", indicating
+   that they shall be in the client's 0-RTT EncryptedExtensions block,
+   and "No" indicating
```
martinthomson: You need to be clear that Early means ONLY 0-RTT.

draft-ietf-tls-tls13.md
```diff
@@ -3827,6 +3896,7 @@ is listed below:
 | key_share [[this document]]              |         Yes |     Clear |
 | pre_shared_key [[this document]]         |         Yes |     Clear |
 | early_data [[this document]]             |         Yes |     Clear |
+|
```
martinthomson: Extra '|'

draft-ietf-tls-tls13.md
martinthomson: Unless you think that we need to allow Early and Client, or Early and Clear, or Early and Encrypted.

draft-ietf-tls-tls13.md
```diff
@@ -2810,32 +2827,82 @@ configuration identifier for each connection, thus preventing
...
+       struct {
+           uint64 ticket_age;
+       } TicketAge;
+
+When the client sends the "early_data" extension, it MUST also send
+a "ticket_age" extension in its EncryptedExtensions block. This value
+contains the time elapsed since the client learned about the server
+configuration that it is using, in milliseconds.  This value can
+be used by the server to limit the time over which early data can
+be replayed.
```
davegarrett: I would really prefer if some recommended time window(s) were specified, rather than leaving this with no default sense of scale. A fairly large hard limit, in addition to more reasonable shorter recommendations, could also be good. I don't know what an ideal set of recommendations would be here, but leaving it unbounded doesn't feel great. Likewise, a general definition of what constitutes a "small allowance for errors" is would be helpful.

ekr: Me too, but this is your text. Recommendations wanted

martinthomson: I think that it's my text.  And I sense a really big bike shed looming.

If I had to pick, I'd say 100ms, but I don't know how good RTT estimates are.

davegarrett: Most likely Martin's text. I don't think I've touched any of the text in this area of the spec. (My guess is you just didn't notice that one of the comments here wasn't his. ;)

And yeah, 'tis a bikesheddy topic, but without any actual numbers here, my worry is that some implementation will happily accept a replay a year and a half after the fact.

To be more precise, I'm asking for about 4 values here: recommended & hard max sanity checks, for both acceptable error & total age. Changing the field to an unsigned 32-bit int, instead of 64-bit, would set an effectively unavoidable max at a bit under 50 days. (2^32-1 ms ~= 49.7 days)


On Wednesday, May 04, 2016 09:03:54 am ekr wrote:
> Me too, but this is your text. Recommendations wanted

draft-ietf-tls-tls13.md
davegarrett: Actually, it looks like a blank table line which should instead list the new ticket age extension.

draft-ietf-tls-tls13.md
martinthomson: Yes, I did the same math and concluded that 32 bits was not just sufficient, but a good idea to limit the total age of tickets.

ekr: We need to distinguish between two different kinds of limits:

1. How large this value can be.
2. How large the *delta* between this value and the age of the ticket can be.

Any limits on how large the age of the ticket can be belong in NewSessionTicket, not here. With that said, since NST is limited to 86400 seconds, it seems like we could make this uint32.

WRT to how much delta/error is allowed, I'd rather leave that to the discretion of the server, as it seems like it's very domain specific. For instance, if you have a single machine with an anti-replay cache and restarts take 2 minutes, then multiple seconds would be reasonabl

draft-ietf-tls-tls13.md
```diff
@@ -2793,47 +2798,107 @@ specified for all records when processing early data records.
...
+time prior to sending the NewSessionTicket message and account for
+that in the value it saves.
+
+There are several potential sources of error that make an exact
+measurement of time difficult.  Variations in client and server clocks
+are likely to be minimal, outside of gross time corrections.  Network
+propagation delays are most likely causes of a mismatch in legitimate
+values for elapsed time.  Both the NewSessionTicket and ClientHello
+messages might be retransmitted and therefore delayed, which might be
+hidden by TCP.
```
davidben: The client could also have since moved to another network. Or perhaps got load-balanced to a different server. Or maybe I got the ticket at 4am when no one was awake and now it's noon and the network's busier.

I'm not sure how useful it is to pin down all the sources of error and recommend that servers SHOULD measure the RTT and put it in NewSessionTicket. I think more realistic is that servers will pick some vaguely reasonable amount of slop and use that.

I also suspect the difference between "an attacker can replay this for 5 seconds" and "an attacker can replay this for 1 minute" isn't much and so keeping the clock tight isn't that valuable. If you're online-ish, you can get a lot of replays in anyway. I'd probably buy "an attacker can replay this for 1 minute" vs. "an attacker can replay this for however long this ticket key is alive" is meaningful since that's now offline-ish.

PR #453, opened 2016-05-11
Wip new key schedule

ekr:
First cut at a new, more linear key schedule. WIP so possible I missed something serious.

@martinthomson @bifurcation, comments, please? Does this seem more straightforward

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3495,7 +3495,8 @@ based on HKDF {{RFC5869}}:
   - HkdfLabel.hash_value is HashValue.
 
   Add-Secret(Secret, InputSecret) =
-       HKDF-Extract(HKDF-Expand(Secret, "", L), InputSecret))
+       HKDF-Extract(HKDF-Expand-Label(Secret, "add secret", "", L),
+                                      InputSecret))
 
   where L denotes the length of the underlying hash function for
   HKDF.
@@ -3528,8 +3529,12 @@ This produces a full key derivation schedule as below.
 (EC)DHE ->  Add-Secret()
                  |
                  v
-           Master Secret --> Derive-Secret("traffic secret",
+           Master Secret --> Derive-Secret("handshake traffic secret",
                  |                         ClientHello + ServerHello) ->
+                 |                         handshake_traffic_secret
+                 |
+                 +---------> Derive-Secret("application traffic secret",
+                 |                         ClientHello...Server Finished) ->
                  |                         traffic_secret_0
                  |                        
                  +---------> Derive-Secret("exporter master secret",
@@ -3546,7 +3551,11 @@ of the diagram are just raw entropy without context, whereas the
 secrets down the right side include handshake context and therefore
 can be used to derive working keys without additional context. With
 the exception of finished keys, the secrets down the left are never
-used to generate working keying material.
+used to generate working keying material. Note that the different
+calls to Derive-Secret() may take different Messages arguments,
+even with the same secret. In a 0-RTT exchange, Derive-Secret is
+called with four distinct transcripts; in a 1-RTT only exchange
+with three distinct transcripts.
 
 If no PSK is available, then the first Add-Secret() stage is omitted
 and the (EC)DHE secret is mixed with the empty secret (0). Of course, this
@@ -3597,7 +3606,7 @@ each class of traffic keys:
 |:------------|--------|-------|
 | 0-RTT Handshake   | early_traffic_secret | "early handshake key expansion" |
 | 0-RTT Application | early_traffic_secret | "early application data key expansion" | 
-| Handshake         | traffic_secret_0 | "handshake key expansion" |
+| Handshake         | handshake_traffic_secret | "handshake key expansion" |
 | Application Data  | traffic secret_N | "application data key expansion" | 
 
 The following table indicates the purpose values for each type of key:
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3548,7 +3548,7 @@ In this diagram, the following formatting conventions apply:
       0 ->  Add-Secret()
                  |
                  v
-                MS2
+            Master Secret
                  |
                  +---------> Derive-Secret(., "application traffic secret",
                  |                         ClientHello...Server Finished)
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1587,15 +1587,15 @@ Exch | + key_share*
                                                        ServerHello  ^ Key
                                                       + key_share*  | Exch
                                                  + pre_shared_key*  v
-                                             [EncryptedExtensions]  ^  Server
-                                             [CertificateRequest*]  v  Parameters
-                                                    [Certificate*]  ^
-                                              [CertificateVerify*]  | Auth
-                                                        [Finished]  v
+                                             {EncryptedExtensions}  ^  Server
+                                             {CertificateRequest*}  v  Parameters
+                                                    {Certificate*}  ^
+                                              {CertificateVerify*}  | Auth
+                                                        {Finished}  v
                                  <--------     [Application Data*]
-     ^ [Certificate*]
-Auth | [CertificateVerify*]
-     v [Finished]                -------->
+     ^ {Certificate*}
+Auth | {CertificateVerify*}
+     v {Finished}                -------->
        [Application Data]        <------->      [Application Data]
 
               +  Indicates extensions sent in the
@@ -1604,7 +1604,11 @@ Auth | [CertificateVerify*]
               *  Indicates optional or situation-dependent
                  messages that are not always sent.
 
-              [] Indicates encrypted messages
+              {} Indicates messages protected using keys
+                 derived from handshake_traffic_secret.
+
+              [] Indicates messages protected using keys
+                 derived from traffic_secret_N
 ~~~
 {: #tls-full title="Message flow for full TLS Handshake"}
 
@@ -1727,15 +1731,15 @@ fatal alert (see {{alert-protocol}}).
            + key_share             -------->
                                                          ServerHello
                                                          + key_share
-                                               [EncryptedExtensions]
-                                               [CertificateRequest*]
-                                                      [Certificate*]
-                                                [CertificateVerify*]
-                                                          [Finished]
+                                               {EncryptedExtensions}
+                                               {CertificateRequest*}
+                                                      {Certificate*}
+                                                {CertificateVerify*}
+                                                          {Finished}
                                    <--------     [Application Data*]
-         [Certificate*]
-         [CertificateVerify*]
-         [Finished]                -------->
+         {Certificate*}
+         {CertificateVerify*}
+         {Finished}                -------->
          [Application Data]        <------->     [Application Data]
 ~~~
 {: #tls-restart title="Message flow for a full handshake with mismatched parameters"}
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3450,60 +3409,74 @@ the sources from the table above.
...
-                                         "exporter master secret",
-                                         handshake_hash, L)
-
-
-  Where handshake_hash contains the entire handshake up to
-  and including the client's Finished, but not including
-  any 0-RTT handshake messages or post-handshake messages.
-  AT this point master_secret SHOULD be securely deleted.
-~~~
+  Add-Secret(Secret, InputSecret) =
```
martinthomson: Which is Secret and which is InputSecret in the diagram below?  I assume that you have an existing secret (the one from the top) and you take a new secret in (InputSecret).  It's not clear though.

draft-ietf-tls-tls13.md
```diff
@@ -3450,60 +3409,74 @@ the sources from the table above.
...
+           Early Secret  --> Derive-Secret("early traffic secret",
+                 |                         ClientHello) -> early_traffic_secret
+                 |                         
+                 v
+(EC)DHE ->  Add-Secret()
+                 |
+                 v
+           Master Secret --> Derive-Secret("handshake traffic secret",
+                 |                         ClientHello + ServerHello) ->
+                 |                         handshake_traffic_secret
```
martinthomson: I'm not sure that this formatting works that well, the handshake_traffic_secret looks like it is an argument to Derive-Secret with this indent.

draft-ietf-tls-tls13.md
```diff
@@ -1857,24 +1832,21 @@ in the first flight, the rest of the handshake uses the same messages.
...
 
                () Indicates messages protected using keys
-                  derived from the static secret.
-
-               {} Indicates messages protected using keys
-                  derived from the ephemeral secret.
+                  derived from the PSK.
 
                [] Indicates messages protected using keys
                   derived from the master secret.
```
martinthomson: Rather than split keys based on what they are derived from, use early_traffic_secret, handshake_traffic_secret, and traffic_secret_0 ?

draft-ietf-tls-tls13.md
```diff
@@ -3450,60 +3409,74 @@ the sources from the table above.
...
-~~~
+  Add-Secret(Secret, InputSecret) =
+       HKDF-Extract(HKDF-Expand-Label(Secret, "add secret", "", L),
+                                      InputSecret))
+
+  where L denotes the length of the underlying hash function for
+  HKDF.
+
+  Derive-Secret(Secret, Label, Messages) =
+       HKDF-Extract(HKDF-Expand(Secret, Label, Hash(Messages), L))
```
svince: Is HKDF-Extract is defined to take two arguments. Waht is the value of the salting term?

svince:
After this change, the keys used to encrypt the handshake traffic are derived from both ES and SS. (As before, they were derived only from ES)
Does this bring any advantages security-wise?

In the current version, the Finished keys contain the handshake context up to CertificateVeryServer, by being transitively derived from mES and mSS. Now these keys don't contain handshake context any more. What is the reasoning behind this change?

draft-ietf-tls-tls13.md
davegarrett: I'd suggest something simpler without the word "Secret" on this line 3 times, e.g.
Combine-Secret(old, new)

draft-ietf-tls-tls13.md
ekr: Yes. The value is L*0. Thanks for the catch.

ekr:
On Wed, May 11, 2016 at 9:53 PM, svince <notifications@github.com> wrote:

> After this change, the keys used to encrypt the handshake traffic are
> derived from both ES and SS. (As before, they were derived only from ES)
> Does this bring any advantages security-wise?
>
Some, because then if you have 0-RTT then you get protection from the PSK.
but mostly, it's just convenience.

In the current version, the Finished keys contain the handshake context up
> to CertificateVeryServer, by being transitively derived from mES and mSS.
> Now these keys don't contain handshake context any more. What is the
> reasoning behind this change?
>
Simplicity. Since they're going to be used anyway to hash the handshake
transcript
it seemed unnecessary. But we could do it if there was strong reason to the
contrary

martinthomson:
This is pretty good.  The simplification is dramatic, which is the number one thing.  Well, the number two thing, given that number one is sign-off from the cryptographers; my limited knowledge of this says that we are using the same basic source of keying material for all the important decisions.  Critically, exporters and resumption branch after the client Finished.

I think that you can revert the bracketing changes by changing the description of each instead.  The fact that there are different keys for different phases is quite important to have in people's faces, I think.

martinthomson:
Note that I had some concerns with clarity of the PR: formatting, the description of `Add-Secret()` and a few things.  I think that those are pretty important and probably need a little attention before landing this.

draft-ietf-tls-tls13.md
```diff
@@ -3450,60 +3409,74 @@ the sources from the table above.
...
+present version of TLS 1.3, secrets are added in the following order:
+
+- PSK
+- (EC)DHE shared secret
+
+This produces a full key derivation schedule as below.
+
+~~~~ 
+                 0
+                 |
```
davegarrett: Trivial nit: it's missing a 'v' arrowhead here.

draft-ietf-tls-tls13.md
davegarrett: ASCII art bikeshedding:
```
           Master Secret --> Derive-Secret("handshake traffic secret",
                 |                         ClientHello + ServerHello)
                 |                         = handshake_traffic_secret
                 |

```
And, all derivations would have their named result on its own line, including early_traffic_secret which is currently in only 2 total lines rather than 3.

I was initially considering suggesting more precisely pointing arrows, but that's just overthinking it and makes this potentially confusing. Just stick each name on its own line and use an equals sign. Simple and easy to parse.

svince:
I'm currently modeling the handshake using the Tamarin Prover. Now i need to decide how to model key generation. Should I rather already use the new key generation, or use the current one?
How likely will this pull request be merged in the next time?

ekr:
I expect to have this resolved and in (though perhaps not identical to
this) by this
time next week.

draft-ietf-tls-tls13.md
ekr: I think I prefer Add, but otherwise, yes.

draft-ietf-tls-tls13.md
```diff
@@ -3535,60 +3498,84 @@ the sources from the table above.
...
-
-
-  Where handshake_hash contains the entire handshake up to
-  and including the client's Finished, but not including
-  any 0-RTT handshake messages or post-handshake messages.
-  AT this point master_secret SHOULD be securely deleted.
-~~~
+  Add-Secret(Old, New) =
+       HKDF-Extract(HKDF-Expand-Label(Secret, "add secret", "", L),
+                   InputSecret))
```
davegarrett: Also need to change the two parameter names in the definition, too.

draft-ietf-tls-tls13.md
```diff
@@ -3535,60 +3498,84 @@ the sources from the table above.
...
+
+  where L denotes the length of the underlying hash function for
```
davegarrett: Can probably just drop the "where" here and it'll fit on one line as a sentence rather than a sentence fragment.

draft-ietf-tls-tls13.md
```diff
@@ -3535,60 +3498,84 @@ the sources from the table above.
...
+  HKDF.
+
+  Derive-Secret(Secret, Label, Messages) =
+       HKDF-Extract(0, HKDF-Expand(Secret, Label, Hash(Messages), L))
+~~~~       
+
+Given a set of n InputSecrets, the final "master secret" is computed
+by iteratively invoking Add-Secret with InputSecret_1, InputSecret_2,
+etc.  The initial secret is simply a string of 0s as long as the size
```
davegarrett: nit: "zeroes" is probably better here

draft-ietf-tls-tls13.md
```diff
@@ -3535,60 +3498,84 @@ the sources from the table above.
...
+called with four distinct transcripts; in a 1-RTT only exchange
+with three distinct transcripts.
 
-Note: Although the steps above are phrased as individual HKDF-Extract
-and HKDF-Expand operations, because each HKDF-Expand operation is
-paired with an HKDF-Extract, it is possible to implement this key
-schedule with a black-box HKDF API, albeit at some loss of efficiency
-as some HKDF-Extract operations will be repeated.
+If a given secret is not available, then the 0-value consisting of
+a string of L 0s is used.
```
davegarrett: nit: see other comment WRT "zeroes"

draft-ietf-tls-tls13.md
```diff
@@ -3619,24 +3606,22 @@ The traffic keying material is generated from the following input values:
...
-| 0-RTT Handshake   | xSS | "early handshake key expansion" | ClientHello  |
-| 0-RTT Application | xSS | "early application data key expansion" | ClientHello |
-| Handshake         | xES | "handshake key expansion" | ClientHello... ServerHello |
-| Application Data  | traffic secret | "application data key expansion" | ClientHello... Server Finished |
+| Record Type | Secret | Phase | 
+|:------------|--------|-------|
+| 0-RTT Handshake   | early_traffic_secret | "early handshake key expansion" |
+| 0-RTT Application | early_traffic_secret | "early application data key expansion" | 
+| Handshake         | handshake_traffic_secret | "handshake key expansion" |
+| Application Data  | traffic secret_N | "application data key expansion" |
```
davegarrett: missing underscore after "traffic"

draft-ietf-tls-tls13.md
```diff
@@ -3602,8 +3589,8 @@ this section then re-deriving the traffic keys as described in
 
 The next-generation traffic_secret is computed as:
 
-  traffic_secret_N+1 = HKDF-Expand-Label(traffic_secret_N,
-                                       "traffic secret", "", L)
+  traffic_secret_N+1 = Derive-Secret(traffic_secret_N,
+                                     "traffic secret", "", L)
```
davegarrett: Derive-Secret() does not take a length parameter.

draft-ietf-tls-tls13.md
davegarrett: Actually, better yet, drop this entirely and just use "key_length" instead of "L" everywhere, as it doesn't need frequent repeating anymore.

draft-ietf-tls-tls13.md
```diff
@@ -3708,13 +3693,9 @@ the TLS PRF. This document replaces the PRF with HKDF, thus requiring
 a new construction. The exporter interface remains the same, however
 the value is computed as:
 
-    HKDF-Expand-Label(HKDF-Extract(0, exporter_secret),
+    HKDF-Expand-Label(exporter_secret
                       label, context_value, length)
```
davegarrett: nit: I suggest "key_length" instead, here, for consistency with the previous use of HKDF-Expand-Label() a bit above.

draft-ietf-tls-tls13.md
```diff
@@ -3708,13 +3693,9 @@ the TLS PRF. This document replaces the PRF with HKDF, thus requiring
 a new construction. The exporter interface remains the same, however
 the value is computed as:
 
-    HKDF-Expand-Label(HKDF-Extract(0, exporter_secret),
+    HKDF-Expand-Label(exporter_secret
```
davegarrett: nit: missing comma after first arg and might all just fit on one line now

draft-ietf-tls-tls13.md
```diff
@@ -3067,9 +3046,9 @@ for each scenario:
 
 | Mode | Handshake Context | Base Key |
 |------|-------------------|----------|
-| 0-RTT | ClientHello | xSS |
-| 1-RTT (Server) | ClientHello ... later of EncryptedExtensions/CertificateRequest | master secret |
-| 1-RTT (Client) | ClientHello ... ServerFinished     | master secret |
+| 0-RTT | ClientHello | Early Secret |
```
davegarrett: nit: capitalization inconsistency; either use lowercase here or uppercase for each line below

PR #458, opened 2016-05-17
Specify whether DHE public keys are zero-padded.

davidben:
TLS 1.2 did not specify this and some implementations have sporadic interoperability issues as a result. Specify that, in TLS 1.3, public keys should be zero-padded. This is believed to be the more compatible TLS 1.2 variant, and having cryptographic messages be fixed-width fields where possible (not that it matters for public keys) seems preferable to saving one byte 1/256 of the time.


davidben:
(Er, that should say "DHE public keys". I'll go do a force-push to fix the commit description.)

PR #466, opened 2016-05-21
Allow multiple tickets in NewSessionTicket

ekr:
@martinthomson Was this what you had in mind?

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3352,7 +3352,7 @@ traffic key.
 
 At any time after the server has received the client Finished message, it MAY send
 a NewSessionTicket message. This message creates a pre-shared key
-(PSK) binding between the ticket value and the following two values derived
+(PSK) binding between the ticket value(s) and the following two values derived
 from the resumption master secret:
 
 ~~~~
@@ -3366,10 +3366,20 @@ from the resumption master secret:
 The client MAY use this PSK for future handshakes by including
 the ticket value in the "pre_shared_key" extension in its ClientHello
 ({{pre-shared-key-extension}}) and supplying a suitable PSK cipher
-suite. Servers may send multiple tickets on a single connection, for
-instance after post-handshake authentication. For handshakes that
+suite. For handshakes that
 do not use a resumption_psk, the resumption_context is a string of
 L zeroes.
+
+The NewSessionTicket message may include multiple ticket
+labels bound to the same resumption_psk and context; this allows
+a client to initiate multiple concurrent connections in parallel
+without exposing themselves to tracking of the PSK label, which
+is carried in the clear. Servers MAY send multiple NewSessionTicket
+messages on a single connection, for
+instance after post-handshake authentication. When a client
+receives a NewSessionTicket message it SHOULD discard any existing
+tickets from this connection.
+
 
 %%% Ticket Establishment
 
@@ -3385,12 +3395,14 @@ L zeroes.
        allow_dhe_resumption(2),
        allow_psk_resumption(4)
      } TicketFlags;
-     
+
+     opaque Ticket<1..2^16-1>;
+  
      struct {
          uint32 ticket_lifetime;
          uint32 flags;
          TicketExtension extensions<2..2^16-2>;
-         opaque ticket<0..2^16-1>;
+         Ticket tickets<2..2^16-1>;
      } NewSessionTicket;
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3387,8 +3387,15 @@ from the resumption master secret:
 The client MAY use this PSK for future handshakes by including
 the ticket value in the "pre_shared_key" extension in its ClientHello
 ({{pre-shared-key-extension}}) and supplying a suitable PSK cipher
-suite. Servers may send multiple tickets on a single connection, for
-instance after post-handshake authentication.
+suite. The NewSessionTicket message may include multiple ticket
```
martinthomson: MAY?

draft-ietf-tls-tls13.md
```diff
@@ -3387,8 +3387,15 @@ from the resumption master secret:
...
+labels bound to the same resumption_psk and context; this allows
+a client to initiate multiple concurrent connections in parallel
+without exposing themselves to tracking of the PSK label, which
+is carried in the clear. Servers may send multiple NewSessionTicket
+messages on a single connection, for
+instance after post-handshake authentication. When a client
+receives a NewSessionTicket message it should discard any existing
```
martinthomson: s/should/MUST/

draft-ietf-tls-tls13.md
```diff
@@ -3404,12 +3411,16 @@ instance after post-handshake authentication.
        allow_dhe_resumption(2),
        allow_psk_resumption(4)
      } TicketFlags;
-     
+
+     struct {
+       opaque ticket<1..2^16-1>;
+     } Ticket;
```
martinthomson: Isn't a typedef enough?

martinthomson:
Yes, this is what I was thinking.

draft-ietf-tls-tls13.md
ekr: Yeah Im onboard witht hat

draft-ietf-tls-tls13.md
ekr: Actually the notation is, believe it or not.
opaque Ticket<1..2^16-1>;

draft-ietf-tls-tls13.md
```diff
@@ -3385,12 +3395,14 @@ L zeroes.
...
-     
+
+     opaque Ticket<1..2^16-1>;
+  
      struct {
          uint32 ticket_lifetime;
          uint32 flags;
          TicketExtension extensions<2..2^16-2>;
-         opaque ticket<0..2^16-1>;
+         Ticket tickets<2..2^16-1>;
```
kaduk: The smallest permitted contents of tickets would be 3 octets, since a Ticket must have at least one octet of content.

draft-ietf-tls-tls13.md
```diff
@@ -3413,11 +3425,12 @@ ticket_extensions
 : A placeholder for extensions in the ticket. Clients MUST ignore
   unrecognized extensions.
 
-ticket
-: The value of the ticket to be used as the PSK identifier.
-The ticket itself is an opaque label. It MAY either be a database
-lookup key or a self-encrypted and self-authenticated value. Section
-4 of {{RFC5077}} describes a recommended ticket construction mechanism.
+tickets
+: One or more ticket values, each of which is to be used as the PSK
```
kaduk: Maybe explicitly say that the multiple values are just concatenated one after each other with no additional framing other than that from the opaque Ticket<>?  I don't see how else someone would interpret this, but just in case...

draft-ietf-tls-tls13.md
```diff
@@ -3366,11 +3366,21 @@ from the resumption master secret:
...
 L zeroes.
 
+The NewSessionTicket message may include multiple ticket
+labels bound to the same resumption_psk and context; this allows
+a client to initiate multiple concurrent connections in parallel
+without exposing themselves to tracking of the PSK label, which
+is carried in the clear. Servers MAY send multiple NewSessionTicket
+messages on a single connection, for
+instance after post-handshake authentication. When a client
+receives a NewSessionTicket message it SHOULD discard any existing
```
kaduk: If we have this text, then a server should not send multiple NewSessionTickets in close succession, as the removed text would have had it do.  Do we want to say anything about the frequency of sending NewSessionTicket?  I am inclined to say that we do not, since it would depend on the use case for the given connection.

ekr:
I think this is lamer than the generation alternative

PR #468, opened 2016-05-21
Cookie for hrr

ekr:
@martinthomson PTAL

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2171,8 +2171,6 @@ Structure of this message:
            Extension extensions<0..2^16-1>;
        } HelloRetryRequest;
 
-[[OPEN ISSUE: Merge in DTLS Cookies?]]
-
 selected_group
 : The mutually supported group the server intends to negotiate and
   is requesting a retried ClientHello/KeyShare for.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2310,6 +2311,34 @@ be taken into account when designing new extensions:
...
+  transports (see {{?RFC6347}} for an example of this).
+
+- Allowing the server to offload state to the client, thus allowing it to send
+  a HelloRetryRequest without storing any state.
+
+The client MUST provide a Cookie extension in every ClientHello it
+sends. If no cookies are available, the extension MUST be empty (i.e.,
+contain an empty-length cookie, allowing the server to send a Cookie in the
+HelloRetryRequest. A server MAY supply a Cookie in the EncryptedExtensions
+block to allow the client to use it for future connections.
```
martinthomson: I see what you are doing.  Until I saw that last bit, I was going to recommend that it MUST NOT be present.  Are you really sure that you want to include the Cookie in subsequent connections?  Won't the PSK ID be enough for that?  Do you think that you need a tracker across connections when PSK isn't in use?

ekr:
I was thinking if you wanted to have a cookie but not PSK for some reason. I could go either way on this. Not allowing it would have the advantage that it would always be strictly for a given connection and if QUIC wanted to have some sort of source address token it could put it elsewhere (like in EE)

Issue #478, opened 2016-05-23
define hash_value in the definition of Hkdf-Expand-Label


martinthomson:
It never actually says that hash_value is H(messages).

ekr:
This should be clear now.

PR #479, opened 2016-05-23
fixes for issue #478

davegarrett:
Some small tweaks and fixes for issue #478:

* Move HKDF-Extract citing to a sentence and bring HKDF-Extract with it.
* Change the Messages arg to HashValue, as the actual hashing is stated in the definition of Derive-Secret which is passing this function a hash, not messages. (fixes issue #478)
* More concise defining of HkdfLabel.
* Use the word "zeroes" instead of "0s".

Edited to add:
* Also, appended a commit to fix an extra trailing paren.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3548,26 +3548,23 @@ that remains is to calculate the key schedule.
 
 The TLS handshake establishes one or more input secrets which
 are combined to create the actual working keying material, as detailed
-below. The key derivation process makes use of the following functions,
-based on HKDF {{RFC5869}}:
+below.
+The key derivation process makes use of the HKDF-Extract and HKDF-Expand
+functions as defined for HKDF {{RFC5869}}, as well as the functions
+defined below:
 
 ~~~~
-  HKDF-Extract(Salt, IKM) as defined in {{RFC5869}}.
-
-  HKDF-Expand-Label(Secret, Label, Messages, Length) =
+  HKDF-Expand-Label(Secret, Label, HashValue, Length) =
        HKDF-Expand(Secret, HkdfLabel, Length)
 
   Where HkdfLabel is specified as:
 
-  struct HkdfLabel {
-    uint16 length;
-    opaque label<9..255>;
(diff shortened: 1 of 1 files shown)


davegarrett:
The other possibility is that a hash of the concatenation of hashes was to be the new HashValue. Simplest change is just to make the arg for what's currently used in the definition and passed in.

draft-ietf-tls-tls13.md
```diff
@@ -3546,35 +3546,32 @@ that remains is to calculate the key schedule.
...
+below.
+The key derivation process makes use of the HKDF-Extract and HKDF-Expand
+functions as defined for HKDF {{RFC5869}}, as well as the functions
+defined below:
 
 ~~~~
-  HKDF-Extract(Salt, IKM) as defined in {{RFC5869}}.
-  
-  HKDF-Expand-Label(Secret, Label, Messages, Length) =
+  HKDF-Expand-Label(Secret, Label, HashValue, Length) =
```
martinthomson: This is a regression, I think.  I think that this should take Messages (a list of messages) as input still, but then define that HkdfLabel.hash_value is H(concat(Messages))

ekr: This was meant to be (generally) Hash(Messages) + Hash(resumption_context).

davegarrett: On Monday, May 23, 2016 05:55:47 pm ekr wrote:
> This was meant to be (generally) Hash(Messages) + Hash(resumption_context).

This is the result of this change, as the 3rd arg passed here below in Derive-Secret is "Hash(Messages) + Hash(resumption_context)". Leaving this as it was and changing hash_value to "Hash(Messages)" results in "Hash(Hash(Messages) + Hash(resumption_context))" (hash of concat of hashes). All other uses of HKDF-Expand-Label provide either an empty string for the 3rd arg or in the case of Exporters, context_value.

martinthomson: Ahh, right.  That means that this is good.

davegarrett:
@ekr: Rebased. Is this ok to merge now?

PR #481, opened 2016-05-24
Use underscores consistently for key/iv parameters

colmmacc:
This change uses underscores consistently for the key/iv paramaters
to aid searching.


martinthomson:
What about the table in Section 7.3.1?

PR #488, opened 2016-06-01
Define ecdsa_sha1 and mark it as a SHOULD NOT alongside other SHA-1 algorithms

davidben:
I'm going to see about getting rid of it anyway, but I think, for the spec, this is probably the better text, as sad as it makes me.

---

The revised signature algorithm mechanism left the {sha1, ecdsa}
SignatureAndHashAlgorithm without a corresponding SignatureScheme defined at
all. SHA-1 does not have a corresponding curve, so defining it alongside
ecdsa_p256_sha256 and friends was a nuisance.

But omitting it, combined with other text, implicitly means it "MUST NOT be
offered or negotiated by any implementation".  Clients (ideally) send only one
ClientHello, so this implies a constraint on TLS 1.2 as well. SHA-1 needs to
go, but I think I was a tad overzealous here.

I wouldn't be surprised {sha1, ecdsa} can be removed from TLS 1.2 right now.
Certainly it will be much much easier to remove than {sha1, rsa}. But this is
probably better a SHOULD NOT rather than a MUST NOT. We wouldn't like to
discover after the fact that a MUST NOT in TLS 1.3 is untenable due to some
ecosystem problem. Notably, old versions of some popular server software ignore
signature algorithms and only sign SHA-1. This mostly affects RSA since ECDSA
isn't that common, but any such ECDSA deployments would be affected.

This is also a little odd for servers as TLS 1.2 says missing sigalgs means the
client is assumed to support {sha1, X}. (Although, strictly speaking, the spec
does not say the server must do anything useful with that advertisement.)

In contrast, dsa_sha1 is already not offered by many major browsers but is
merely a SHOULD NOT.

Define ecdsa_sha1_RESERVED as a companion to dsa_*_RESERVED to bring it out of
the implicit MUST NOT. Then include it in the SHOULD NOT lists alongside
rsa_pkcs1_sha1 and dsa_sha1.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2374,8 +2374,9 @@ The "extension_data" field of this extension contains a
            dsa_sha256_RESERVED (0x0402),
            dsa_sha384_RESERVED (0x0502),
            dsa_sha512_RESERVED (0x0602),
+           ecdsa_sha1_RESERVED (0x0203),
            obsolete_RESERVED (0x0000..0x0200),
-           obsolete_RESERVED (0x0203..0x0400),
+           obsolete_RESERVED (0x0204..0x0400),
            obsolete_RESERVED (0x0404..0x0500),
            obsolete_RESERVED (0x0504..0x0600),
            obsolete_RESERVED (0x0604..0x06FF),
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3302,8 +3303,9 @@ message.
 in the sender's end-entity certificate. RSA signatures MUST use an
 RSASSA-PSS algorithm, regardless of whether RSASSA-PKCS-v1_5 algorithms
 appear in "signature_algorithms". SHA-1 MUST NOT be used in any signatures in
-CertificateVerify. (Note that rsa_pkcs1_sha1 and dsa_sha1, the only defined
-SHA-1 signature algorithms, are undefined for CertificateVerify signatures.)
+CertificateVerify. (Note that all SHA-1 signature algorithms in this
+specification are defined solely for use in certificates, not CertificateVerify
+signatures.)
```
davegarrett: I suggest also simplifying/clarifying this slightly while we're here. This doesn't really need to be in parenthesis, either. (minor preexisting nit)

Suggested replacement:

All SHA-1 signature algorithms in this specification are defined solely for use in legacy certificates, and are not valid for CertificateVerify signatures.

PR #489, opened 2016-06-01
Rename HandshakeType.session_ticket to HandshakeType.new_session_ticket.

davidben:
This is purely aesthetic, but apparently no one realizes the name is different, not even IANA, so we may as well go fix it.

(I don't know how to write IANA considerations, so I've left it a TODO. But if you give me text, I'll update this PR to include it.)

---
For consistency. At a glance, it seems most TLS implementations did not realize
the HandshakeType value was actually named session_ticket and assumed it was
named new_session_ticket anyway. Moreover, the IANA entry was named neither and
actually named NewSessionTicket since RFC 4507's IANA considerations used the
wrong name.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1943,9 +1943,6 @@ processed and transmitted as specified by the current active session state.
            } body;
        } Handshake;
 
-[[TODO: Add IANA Considerations to rename the 'NewSessionTicket' HandshakeType
-value to 'new_session_ticket'.]]
-
 The TLS Handshake Protocol messages are presented below in the order they
 MUST be sent; sending handshake messages in an unexpected order
 results in an "unexpected_message" fatal error. Unneeded handshake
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1938,11 +1938,14 @@ processed and transmitted as specified by the current active session state.
...
                case finished:              Finished;
-               case session_ticket:        NewSessionTicket;
+               case new_session_ticket:    NewSessionTicket;
                case key_update:            KeyUpdate;
            } body;
        } Handshake;
 
+[[TODO: Add IANA Considerations to rename the 'NewSessionTicket' HandshakeType
+value to 'new_session_ticket'.]]
+
```
ekr: Do you think you could do this TODO?

davidben: Sure. I just don't know how to write IANA considerations. But I'm happy to include the right sentence wherever.

seanturner: Think of the IANA considerations as instructions for somebody changing a value in a registry table that knows nada about the protocol.

Is this what you're trying to do (add this in s11):

`IANA is requested to rename item 4 in the TLS HandshakeType Registry to 'new_session_ticket', i.e., change the table entry for value 4 to 'NewSessionTicket (renamed new_session_ticket)'.`

davidben: Oh, I figured there might have been some very particular format or something.

I've added a sentence to the bit at the bottom that talks about HandshakeType. I see the other entries use "[SHALL update/has updated]" so I mimicked them. Does that look good?

seanturner: lgtm

PR #490, opened 2016-06-02
Use more capital letters to describe version negotiation.

davidben:
People will continue getting this wrong until the end of time, but add an extra
MUST in the server_version half on the off chance someone reads it.

Also clarify that client_version is a maximum version. ("The version of the TLS
protocol by which the client wishes to communicate during this session"
suggests it should be an exact match.)


draft-ietf-tls-tls13.md
```diff
@@ -2095,9 +2094,10 @@ one extension (at least "key_share" or "pre_shared_key").
 
 server_version
 : This field will contain the lower of that suggested by the client
-  in the ClientHello and the highest supported by the server.  For
-  this version of the specification, the version is { 3, 4 }.  (See
-  {{backward-compatibility}} for details about backward compatibility.)
+  in the ClientHello and the highest supported by the server.
+  Servers MUST accept ClientHello versions higher than those supported.
```
ekr: Perhaps "Servers MUST accept ClientHello messages with versions higher than those supported and negotiate the highest mutually supported version"?

davidben: Done. Tweaked some of the surrounding words a bit so it didn't seem quite as repetitive.

draft-ietf-tls-tls13.md
```diff
@@ -2093,11 +2093,13 @@ As of TLS 1.3, all clients and servers will send at least
 one extension (at least "key_share" or "pre_shared_key").
 
 server_version
-: This field will contain the lower of that suggested by the client
-  in the ClientHello and the highest supported by the server.
-  Servers MUST accept ClientHello versions higher than those supported.
-  For this version of the specification, the version is { 3, 4 }.
-  (See {{backward-compatibility}} for details about backward compatibility.)
+: This field contains the version of TLS negotiated for this session.  Servers
+  MUST select the lower of the highest supported version and the version
```
davegarrett: "highest supported version" is too ambiguous; "highest supported server version" would be preferable.

The whole sentence could be a little more clear, however. I suggest something like:

"Servers MUST select the lower of the two endpoints' supported versions: either the highest supported server version or the version offered by the client in its ClientHello."

PR #491, opened 2016-06-03
It really can't be used without encryption now, 

martinthomson:
and arguably should never be used in that fashion.

I know that this stuff badly needs a rewrite, but this is just egregious.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -273,10 +273,6 @@ that has two basic properties:
 - The connection is reliable.  Messages include an authentication
   tag which protects them against modification.
 
-Note: The TLS Record Protocol can operate in an insecure mode but is generally
-only used in this mode while another protocol is using the TLS Record
-Protocol as a transport for negotiating security parameters.
-
 The TLS Record Protocol is used for encapsulation of various higher-level
 protocols. One such encapsulated protocol, the TLS Handshake Protocol, allows
 the server and client to authenticate each other and to negotiate an encryption


draft-ietf-tls-tls13.md
```diff
@@ -273,10 +273,6 @@ that has two basic properties:
 - The connection is reliable.  Messages include an authentication
   tag which protects them against modification.
 
-Note: The TLS Record Protocol can operate in an insecure mode but is generally
-only used in this mode while another protocol is using the TLS Record
-Protocol as a transport for negotiating security parameters.
-
```
ekr: This is still true for the HEllo messages.

martinthomson: Oh, so that's worth putting in the *introduction*?  This is talking about integrity-only modes, which are dead.  Don't split hairs on me.

And I would contest the point that the handshake is insecure just because it is sent in the clear.  Conveying the necessary level of nuance to be able to understand this point isn't worth spending introductory text on.

davegarrett: Maybe just replace with a short sentence in here somewhere stating that the handshake starts in the clear, it switches to encrypted transport as soon as possible, and is retroactively authenticated before handshake completion?

martinthomson: You don't need to say anything at all.  It's the introduction.  How much of that is more important or even as important as things like "servers are always authenticated and clients are sometimes authenticated" or "messages have confidentiality and integrity protection" or "0-RTT is fast, but risky".

PR #492, opened 2016-06-03
Remove cruft from Record Layer section, reorganize a little

martinthomson:
The section on connection states was almost all cruft.  I made a new section for nonce construction and moved the piece on sequence numbers there.

The only text I modified was the first three paragraphs of the nonce section, which needed to be scrubbed for references to "connection state", and the lead-in section on the AEAD, which I moved to where the other AEAD stuff was and expanded a little to include some forward references.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -916,128 +916,10 @@ itself is not protected by encryption unless the sender uses the
 supplied padding mechanism -- see {{record-padding}} for more details.
 
 
-##  Connection States
-
-[[TODO: I plan to totally rewrite or remove this. IT seems like just cruft.]]
-
-A TLS connection state is the operating environment of the TLS Record
-Protocol.  It specifies a record protection algorithm and its
-parameters as well as the record protection keys and IVs for the
-connection in both the read and the write directions. The security
-parameters are set by the TLS Handshake Protocol, which also determines
-when new cryptographic keys are installed and used for record
-protection.
-The initial current state always specifies that records are
-not protected.
-
-The security parameters for a TLS Connection read and write state are set by
-providing the following values:
-
-{:br: vspace="0"}
-
-connection end
-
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -916,129 +916,11 @@ itself is not protected by encryption unless the sender uses the
...
-: A 32-byte value provided by the client.
-
-server random
-
-: A 32-byte value provided by the server.
-{: br}
-
-These parameters are defined in the presentation language as:
-
-%%% Security Parameters
```
davegarrett: This fails to build:
https://travis-ci.org/tlswg/tls13-spec/builds/134950214

Looks like you also need to delete the section in the appendix that transcludes this. (currently section A.5)
https://tools.ietf.org/html/draft-ietf-tls-tls13-13#appendix-A.5
https://tlswg.github.io/tls13-spec/#the-security-parameters

draft-ietf-tls-tls13.md
```diff
@@ -3662,7 +3563,7 @@ implementations SHOULD delete traffic_secret_N. Once the directional
 keys are no longer needed, they SHOULD be deleted as well.
 
 
-## Traffic Key Calculation
+## Traffic Key Calculation {#traffic-key-calculation}
```
davegarrett: This is not needed; I recommend reverting this one line. All sections have automatically generated anchor links based on their names in exactly this format. Note line 946 in the removed section above where this section was already cited without issue. You only need to give it a custom anchor if the title is long and a shorter link would be preferred. (e.g. {#nonce} above is fine, though the title itself isn't that long)

davegarrett:
LGTM

@ekr: Did you still want to replace any of this? The stuff being cut here is very obsolete and must confuse anyone who attempts to read through the whole spec.

ekr:
Yes, I'm working on it this week.

PR #493, opened 2016-06-04
Explicitly specify multiply concurrent tickets

ekr:


ekr:
We need to cover this at IETF.

kaduk:
To me, the ticket-encrypting-key change use case is the most compelling, and seems useful enough to support adding this now.

ekr:
This was resolved with separate changes.

PR #494, opened 2016-06-07
Implementations should return "handshake_failure" to peers sending SS…

ttaubert:
…L v3 in the {Client,Server}Hello message


davegarrett:
No, the fatal alert here is very clearly "protocol_version" and is stated multiple times in the document elsewhere.

My guess is you think this is required because SSL did not yet have the precision to say this and only had "handshake_failure", however I couldn't care less. SSL is still required to abort on a fatal error, even if it has no idea what it is, and special-casing this is a horrible idea. TLS implementations need to have very clear instructions on what to do for what contingency, and this is very clearly defined as the response here.

Definition of this error code in current draft:
https://tools.ietf.org/html/draft-ietf-tls-tls13-13#page-31

Requirement of full abort on all fatal alerts in SSL3:
https://tools.ietf.org/html/rfc6101#section-5.4.2

PR #499, opened 2016-06-15
Define the interaction between 0-RTT and multiple PSK identities.

davidben:
I would actually much rather we only offer one PSK identity since TLS 1.2 was perfectly fine offering just one and resumption has no use for offering multiple, but I suspect I will lose that fight. Absent that, at least make the protocol well-defined.

---
Unlike both TLS 1.2 resumption and TLS 1.2 PSK, TLS 1.3 increases the state
space and allows clients to offer multiple PSK identities. This mostly works,
but 0-RTT presumes that only one is offered. Define 0-RTT to presume the first
identity is selected and clarify that other shapes are possible due to the
increased state space.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1829,7 +1829,7 @@ negotiate an (EC)DHE-PSK cipher suite.
 When resuming via a PSK with an appropriate ticket (i.e., one with
 the "allow_early_data" flag), clients can also send data on their first
 flight ("early data"). This data is encrypted solely under keys
-derived using the PSK as the static secret.  As shown in
+derived using the first offered PSK as the static secret.  As shown in
 {{tls-0-rtt}}, the Zero-RTT data is just added to the 1-RTT handshake
 in the first flight, the rest of the handshake uses the same messages.
 
@@ -2691,7 +2691,9 @@ The "extension_data" field of this extension contains a
 
 identities
 : A list of the identities (labels for keys) that the client is willing
-  to negotiate with the server.
+  to negotiate with the server. If sent alongside the "early_data"
+  extension (see {{early-data-indication}}), the first identity is the
+  one used for 0-RTT data.
 
 selected_identity
 : The server's chosen identity expressed as a (0-based) index into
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2708,9 +2710,14 @@ suite, if available.
 If the server selects a PSK cipher suite, it MUST send a
 "pre_shared_key" extension with the identity that it selected.
 The client MUST verify that the server's selected_identity
-is within the range supplied by the client. If any other value
-is returned, the client MUST generate a fatal
-"unknown_psk_identity" alert and close the connection.
+is within the range supplied by the client. If the server supplies an
+"early_data" extension, the client MUST verify that the server selected the
+first offered identity. If any other value is returned, the client MUST
+generate a fatal "unknown_psk_identity" alert and close the connection.
```
martinthomson: This actually suggests an alternative possibility, which is that the selected PSK is used in place of the (EC)DHE secret for pure PSK suites and that the selection doesn't matter when PSK is only used for 0-RTT.  I guess that your position is that this is complicated and we don't need that complication.  I agree.

draft-ietf-tls-tls13.md
```diff
@@ -2708,9 +2710,14 @@ suite, if available.
...
+
+Note that, although 0-RTT data is encrypted with the first PSK identity, the
```
martinthomson: The first comma is redundant.

draft-ietf-tls-tls13.md
```diff
@@ -2782,9 +2789,11 @@ can behave in one of two ways:
   to accept only a subset of the early data messages.
 
 [[OPEN ISSUE: are the rules below correct? https://github.com/tlswg/tls13-spec/issues/451]]
-Prior to accepting the "early_data" extension, the server MUST
-validate that the session ticket parameters are consistent with its
-current configuration. It MUST also validate that the extensions
+Prior to accepting the "early_data" extension, the server MUST check the
+first PSK identity (see {{pre-shared-key-extension}}) corresponds to a
+valid ticket and validate that the session parameters are consistent
+with its current configuration.
```
martinthomson: I would state this more simply: If the server selects a PSK identity other than the first, it MUST reject early data.  

That's stronger than what you have, but I think that it also suggests an order of operations that is inherently safer: pick your parameters and only then work out whether to take this 0-RTT stuff.  I know that some implementations will want to maximize how often 0-RTT is used, but that isn't necessarily the best choice security-wise.

PR #503, opened 2016-06-22
Move ticket_age to the early_data extension

martinthomson:
This allows us to remove EncryptedExtensions from the 0-RTT flight.
Doing this means that the server has all the information that they
need to generate a ServerHello in the ClientHello.  That reduces
the complexity of the state machine needed considerably.

The 0-RTT Finished message is only needed for key confirmation for
0-RTT, which means that processing the Finished is not necessary.
If it's wrong, then the connection can fail.  That's different to
ticket_age, which would be "wrong" very often, so that we could
only reject 0-RTT.

This reduces generality somewhat, but since we've not identified
any other use for EncryptedExtensions, this isn't the end of the
world.  The other candidate use for EncryptedExtensions was
encrypted SNI, but it's not clear that a design based on
EncryptedExtensions is the right way to solve that problem.

A bad server can unmask clients by using a zero, fixed, or
predictable mask.  I don't believe that this is worth complicating
the design to counter.  This is a case where the dumbest possible
solution is the best.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1764,7 +1764,7 @@ Although TLS PSKs can be established out of band,
 PSKs can also be established in a previous session and
 then reused ("session resumption"). Once a handshake has completed, the server can
 send the client a PSK identity which corresponds to a key derived from
-the initial handshake (See {{new-session-ticket-message}}). The client
+the initial handshake (See {{NewSessionTicket}}). The client
 can then use that PSK identity in future handshakes to negotiate use
 of the PSK; if the server accepts it, then the security context of the
 original connection is tied to the new connection. In TLS 1.2 and
@@ -3025,7 +3025,7 @@ client authentication is not available in the 0-RTT case.
 
 CertificateVerify
 : A signature over the value Hash(Handshake Context + Certificate) + Hash(resumption_context)
-See {{new-session-ticket-message}} for the definition of resumption_context.
+See {{NewSessionTicket}} for the definition of resumption_context.
 
 Finished
 : A MAC over the value Hash(Handshake Context + Certificate + CertificateVerify) +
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2837,14 +2840,14 @@ As noted in {{zero-rtt-data}}, TLS provides only a limited
...
+The "masked_ticket_age" parameter in the client's "early_data" extension SHOULD be used by
 servers to limit the time over which the first flight might be
 replayed.  A server can store the time at which it sends a server
 configuration to a client, or encode the time in a ticket.  Then, each
-time it receives an early_data extension, it can check to see if the
+time it receives an "early_data" extension, it can unmask the value and check to see if the
 value used by the client matches its expectations.
 
-The "ticket_age" value provided by the client will be shorter than the
+The "masked_ticket_age" value provided by the client will be shorter than the
```
davidben: Nit: Perhaps change 'The "masked_ticket_age" value' to 'The unmasked value' or similar? Otherwise this reads like you're talking about what the client sent rather than the true value. (Whereas the paragraph above says to unmask it.)

draft-ietf-tls-tls13.md
```diff
@@ -3443,6 +3423,10 @@ ticket_lifetime
   for a shorter period of time than what is stated in the
   ticket_lifetime.
 
+ticket_age_mask
+: A randomly 32-bit value that is used to mask the age of the ticket that the
+  client includes in the "early_data" extension.
```
davidben: Do you think it's worth mentioning that this value needs to be part of the server's session state? (Either stashed into the ticket or in your stateful session cache.)

davidben:
> A bad server can unmask clients by using a zero, fixed, or predictable mask. I don't believe that this is worth complicating the design to counter.

(A bad server can equally easily unmask clients by making single-use tickets easily correlatable, so I think this is fine.)

draft-ietf-tls-tls13.md
martinthomson: I added a note to the section that explains the field.

draft-ietf-tls-tls13.md
martinthomson: Good point.  Done.

ekr:
Yes, this looks right. Why don't you propose it?

ekr:
merged by hand

PR #512, opened 2016-07-04
Added support for DTLS 1.3 to the TLS 1.3 specification

hannestschofenig:
Made various changes throughout the document in an attempt to propose the integration of DTLS 1.3 into the TLS 1.3 specification. Although a lot of the text of RFC 6347 has been re-used it is worth pointing out that the cookie mechanism is different in DTLS 1.3 compared to DTLS 1.2 due to the use of the cookie extension.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1296,7 +1296,7 @@ verify the TLS MAC.  The DTLS record format is shown below:
 
       struct {
            ContentType type;
-           ProtocolVersion version;
+           ProtocolVersion version = { 3, 1 };
            uint16 epoch;                         // DTLS-related field
            uint48 sequence_number;               // DTLS-related field
            uint16 length;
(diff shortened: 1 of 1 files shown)


ekr:
Hannes, can you fix the build so it compiles? See:
https://travis-ci.org/tlswg/tls13-spec/builds/142169418

davegarrett:
To start, this needs rebasing, as it looks like it reverts quite a few recent commits.

Personally, I think this is too much to dump into this document and just makes it harder to read. Keeping the two separate, but directly referencing eachother where appropriate, makes more sense to me. (e.g. do both specs in parallel, and release at same time)

hannestschofenig:
Added various review comments by Ilari:
 * Corrected text related to Length and payload description.
 * Removed text about 're-handshake'
 * Removed sentence that talked about  block padding and compression in PMTU section.
 * Indicated that ClientHello and HelloRetryRequest are included in handshake hashes.

Open issues: 
 * Where should the client put the cookie? (cookie extension or in the legacy_cookie field)?
 * What version number should be used in the DTLS handshake and in the DTLS record layer?

hannestschofenig:
Changes: 

 * Modified version number, as suggested by Ilari. 
 * Added placeholder for text about re-transmission of post-authentication messages. 
 * Added placeholder for text about rekeying. 
 * Indicated HelloVerifyRequest and KeyUpdate messages as depreciated for DTLS 1.3.
 * Updated design rational
 * Added cookie extension to the table in the IANA consideration section and indicated that it can be transmitted in clear as part of the ClientHello message.

ekr:
I definitely won't get to this by draft-14

ekr:
Hannes has started a separate spec.

boaks:
> Hannes has started a separate spec.

Is there a repo for the separate spec.?
Using CoAP I'm very interessted in a future version of DTLS.

PR #513, opened 2016-07-05
Make HKDF hash function and L explicit

FiloSottile:


davegarrett:
Thanks. I think there's no explanation of 'L' anywhere in here, currently, so this is a notable fix.

Doing a quick search for " L " shows another two places where it's used in a sentence. I think those are the only other two places you'd need to update.

FiloSottile:
Only found another one, what am I missing?

In general I think there needs to be in general more clarity as to what hash function is used where (which to my understanding is always the cipher suite one, except in crafting the digitally-signed data) but it's not as obvious. Opinions?

davegarrett:
On Tuesday, July 05, 2016 07:27:27 pm Filippo Valsorda wrote:
> Only found another one, what am I missing?

My first count was wrong; looks like you were only missing the one more, and the other must've been in the previous version of the commit already. I think you've got all the uses of 'L' now, so this should be good.

> In general I think there needs to be in general more clarity as to what hash function is used where (which to my understanding is always the cipher suite one, except in crafting the digitally-signed data) but it's not as obvious.

Unless I'm forgetting something, the only separately decided hash functions are the one for the cipher suite to be used with HKDF/HMAC and the one(s) for signatures (negotiated via signature_algorithms extension).

> Opinions?

My general opinion is always that explicit is better than implicit in spec docs. If there's a point that you think needs better clarification, please do submit a new PR with some updates.

ekr:
Merged by hand.

Issue #510, opened 2016-07-01
A lot of important stuff doesn't appear in the TOC


martinthomson:
...including extensions, and messages.

These are a really common thing to want to find.  A flatter structure would fix this; as would a deeper TOC, though my preference is for the former.

davegarrett:
After looking at it for a little bit, I think the subsections in the current Handshake Protocol section (6.3) can be pulled up a tier to be part of a standalone section just called Messages, and the remaining text that starts the current Handshake Protocol section (6.3) would be better placed closer to Handshake Protocol Overview (6.2).

https://tools.ietf.org/html/draft-ietf-tls-tls13-13#section-6.3
https://tools.ietf.org/html/draft-ietf-tls-tls13-13#section-6.2

Current layout:
6. The TLS Handshaking Protocols
    6.1. Alert Protocol
        6.1.1. Closure Alerts
        6.1.2. Error Alerts
    6.2. Handshake Protocol Overview
        6.2.1. Incorrect DHE Share
        6.2.2. Resumption and Pre-Shared Key (PSK)
        6.2.3. Zero-RTT Data
    6.3. Handshake Protocol
        6.3.1. Key Exchange Messages
        6.3.2. Hello Extensions
        6.3.3. Server Parameters
        6.3.4. Authentication Messages
        6.3.5. Post-Handshake Messages

Possible new layout:
6. The TLS Handshaking Protocols
    6.1. Alert Protocol
        6.1.1. Closure Alerts
        6.1.2. Error Alerts
    6.2. Handshake Protocol
        6.2.1. Incorrect DHE Share
        6.2.2. Resumption and Pre-Shared Key (PSK)
        6.2.3. Zero-RTT Data
7. Messages
    7.1. Key Exchange Messages
        [...]
    7.2. Hello Extensions
        [...]
    7.3. Server Parameters
    7.4. Authentication Messages
    7.5. Post-Handshake Messages

with 6.2 in the new absorbing a bit of the root of the previous 6.3.

I think this would be sufficient to get all the important stuff in the TOC and abolish the mild insanity of the 5 number subsection IDs.

PR #516, opened 2016-07-07
Flatter Sectioning (for issue #510)

davegarrett:
This is a PR for the section flattening scheme I noted in issue #510. The handshake structs are a little awkwardly positioned, but other than that, this is pretty straightforward. A little bit of excessive redundant text is also cut here. The reasons to do this are:
1) The TOC will actually list all the messages and extensions by name, which is fairly important information, and should make browsing the document easier.
2) No more 5 part subsection ID silliness.
3) The relevant handshake enum and struct are now defined before all the stuff that mentions them.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1844,20 +1844,20 @@ The contents and significance of each message will be presented in detail in
 the following sections.
 
 
-##  Handshake Protocol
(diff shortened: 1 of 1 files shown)


davegarrett:
Closing this in favor of the more drastic restructuring and improvements in PR #524.

PR #523, opened 2016-07-08
Refine rules for accepting 0-RTt. Fixes #451

### LINKED ISSUE #451: What are the exact rules for how closely 0-RTT must conform to the handshake where it was established
We currently have "identical" but Ilari argues we should have more specific rules

ekr:
As proposed by Ilari on the list.

I opted not to require that the various extensions where the server sends extra cert info (OCSP, etc.) be the same, as the client has already sent the data at this point and can make reliance decisions if those change.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2664,8 +2664,8 @@ obfuscated_ticket_age
 A server MUST validate that the ticket_age is within a small
 tolerance of the time since the ticket was issued (see {{replay-time}}).
 
-All of the parameters for the 0-RTT data (symmetric cipher suite,
-ALPN, etc.) MUST be those which were negotiated in the connection
+The parameters for the 0-RTT data (symmetric cipher suite,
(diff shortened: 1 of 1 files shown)


ekr:
@ilaril PTAL.

ilaril:
On Fri, Jul 08, 2016 at 11:07:00AM -0700, ekr wrote:
> @ilaril PTAL.

Yeah, much better than what there was previously.


I did run over all other known extensions to check if those are either
deprecated, (presumably[1]) connection control or related to signature
auth. All seem to be.


[1] Not all extensions explicitly say are those connection-scoped or
session-scoped in TLS 1.2.

davegarrett:
This commit makes Travis sad. You need to add RFC7443 to the informative references list up top for this to build properly.

ekr:
Thanks. I will fix pre-landing.

ekr:
Rebased at: https://github.com/tlswg/tls13-spec/pull/527

PR #525, opened 2016-07-09
Security analysis

ekr:
Here is a first cut at the security properties, built on the big restructure.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4230,7 +4230,7 @@ the handshake is complete, this does not compromise the security of the
 session key (See {{DOW92}}).
 
 Protection of endpoint identities.
-The server's identity (certificate) should be protected against passive
+: The server's identity (certificate) should be protected against passive
 attackers. The client's identity should be protected against both passive
 and active attackers.
 {:br}
@@ -4241,9 +4241,9 @@ establishment of a unique, secret, shared, key established by an
 the handshake transcript, as well as tied to the server's identity by
 a MAC. If the client is authenticated by a certificate, it also signs
 over the handshake transcript and provides a MAC tied to both
-identities. {{SIGMA}} provides an overview of the analysis of this
-type of key exchange protocol. If fresh (EC)DHE keys are used for each
-connection, then the output keys are forward secret.
+identities. {{SIGMA}} describes of the analysis of this type of key
+exchange protocol. If fresh (EC)DHE keys are used for each connection,
+then the output keys are forward secret.
 
 The PSK and resumption-PSK modes bootstrap from a long-term shared
 secret into a unique per-connection short-term session key. This
@@ -4263,7 +4263,9 @@ As soon as the client and the server have exchanged enough information
 to establish shared keys, the remainder of the handshake is encrypted,
 thus providing protection against passive attackers. Because the server
 authenticates before the client, the client can ensure that it only
-reveals its identity to an authentic server.
+reveals its identity to an authentic server. Note that implementations
+must use the provided record padding mechanism during the handshake
+to avoid leaking information about the identities due to length.
 
 The 0-RTT mode of operation generally provides the same security
 properties as 1-RTT data, with the two exceptions that the 0-RTT
@@ -4315,7 +4317,7 @@ Informally, TLS 1.3 provides these properties by AEAD-protecting the
 plaintext with a strong key. AEAD encryption protects confidentiality
 and integrity for the data. Non-replayability is provided by using
 a separate nonce for each record, with the nonce being derived from
-the record sequence number (Section {{nonce}}), with the sequence
+the record sequence number ({{nonce}}), with the sequence
 number being maintained independently at both sides thus records which
 are delivered out of order result in AEAD deprotection failures.
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -245,7 +245,6 @@ informative:
          ins: K. Paterson
        target: http://www.isg.rhul.ac.uk/~kp/TLS-AEbounds.pdf
        date: 2016
-
 --- abstract
 
 This document specifies Version 1.3 of the Transport Layer Security
@@ -4128,9 +4127,55 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
 been shown to be insecure in some scenarios.
 
 
-#  Security Analysis
+#  Overview of Security Properties {#security-analysis}
 
-[[TODO]]
+A complete security analysis of TLS is outside the scope of this document.
+In this section, we provide an informal description the desired properties
+as well as references to more detailed work in the research literature
+which provides more formal definitions.
+
+Because TLS is a modular protocol, we cover properties of the handshake
+separately from those of the record layer.
+
+## Handshake {#security-handshake}
+
+The TLS handshake is an Authenticated Key Exchange (AKE) protocol which
+is intended to provide both one-way authenticated (server-only) and
+mutually authenticated (client and server) functionality. At the completion
+of the handshake, each side outputs its view on the following properties.
+
+- A "session key" from which can be derived a set of working keys.
+- A set of cryptographic parameters (algorithms, etc.)
+- The identities of the communicating parties.
+
+The handshake should provide the following properties.
+
+Establishing the same key:
+: The handshake needs to output the same session key on both sides of the
+handshake (See {{CK01}}; defn 1, part 1).
+
+Secrecy of the session key: 
+: The shared session key should be known only to the communicating
+parties, not to the attacker (See {{CK01; defn 1, part 2).  Note that
+in a unilaterally authenticated connection, the attacker can establish
+its own session keys with the server, but those session keys are
+distinct from those established by the client.
+
+{:br}
+
+
+
+
+
+
+
+
+
+
+
+
+## Record Layer {#security-record-layer}
+
 
 # Working Group Information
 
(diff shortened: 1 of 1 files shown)


davegarrett:
Here's the GitHub branch diff link, just for ease of reading here:
https://github.com/ekr/tls13-spec/compare/restructure...ekr:security_analysis

ekr:
Rebased now that I landed big restructure. This should be much easier to read.

@tvdmerwe

ekr:
@bifurcation PTAL. I know this is incomplete but worth shipping as a first draft with -14

ekr:
@wbl PTAL.

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+The TLS handshake is an Authenticated Key Exchange (AKE) protocol which
+is intended to provide both one-way authenticated (server-only) and
+mutually authenticated (client and server) functionality. At the completion
+of the handshake, each side outputs its view on the following values:
+
+- A "session key" from which can be derived a set of working keys.
+- A set of cryptographic parameters (algorithms, etc.)
+- The identities of the communicating parties.
+
+We assume that the attacker has complete control of the network in
```
bifurcation: Maybe pull this up to the intro, since it applies to both handshake and records.

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+We cover properties of the handshake separately from those of the record layer.
+
+## Handshake {#security-handshake}
+
+The TLS handshake is an Authenticated Key Exchange (AKE) protocol which
+is intended to provide both one-way authenticated (server-only) and
+mutually authenticated (client and server) functionality. At the completion
+of the handshake, each side outputs its view on the following values:
+
+- A "session key" from which can be derived a set of working keys.
```
bifurcation: Would be nice if this actually mapped to something above.  `masterSecret`?

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+distinct from those established by the client.
+
+Peer Authentication.
+: The client's view of the peer identity should reflect the server's
+identity. If the client is authenticated, the server's view of the
+peer identity should match the client's identity.
+
+Uniqueness of the session key:
+: Any two distinct handshakes should produce distinct, unrelated session
+keys [REF: Need cite.]
```
bifurcation: Is the notion of "unrelated" here generally known?

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+
+Downgrade protection.
+: The cryptographic parameters should be the same on both sides and
+should be the same as if the peers had been communicating in the
+absence of an attack (See {{BBFKZG16}}; defns 8 and 9}).
+
+Forward secret
```
bifurcation: It seems like there's overlap between this and the uniqueness goal.  Either way, you're talking about information from one session not providing information about another session.  Perhaps there's a broader overall "independence of sessions" goal?, e.g.:

```
Independence of sessions.
: If the session key and all long-term authentication keys are compromised for a given session, this does not compromise the security of any other session. (This is a generalization of the common notion of Forward Secrecy, see {{DOW92}}.)
```

ekr: Why do you think that these are related?

It's trivial to design handshakes which provide for uniqueness but allow the attacker to control the algorithms: for instance, just allow two strong groups and don't integrity check the negotiation of which group to use.

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+secret may have been established in a previous handshake. If
+PSK-(EC)DHE modes are used, this session key will also be forward
+secret. The resumption-PSK mode has been designed so that the
+resumption master secret computed by connection N and needed to form
+connection N+1 is separate from the traffic keys used by connection N,
+thus providing forward secrecy between the connections.
+
+For all handshake modes, the Finished MAC (and where present, the
+signature), prevents downgrade attacks. In addition, the random
+tainting mechanism described in {{server-hello}} allows the detection
```
bifurcation: "Random tainting" sounds like you're tainting at random :)  How about: "The use of certain bytes in the client and server random to indicate TLS 1.3 support ..."

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+{:br}
 
-[[TODO]]
+Informally, TLS 1.3 provides these properties by AEAD-protecting the
+plaintext with a strong key. AEAD encryption {{RFC5116}} provides confidentiality
+and integrity for the data. Non-replayability is provided by using
+a separate nonce for each record, with the nonce being derived from
+the record sequence number ({{nonce}}), with the sequence
+number being maintained independently at both sides thus records which
+are delivered out of order result in AEAD deprotection failures.
```
bifurcation: Is "deprotection" a standard term?  I get what you mean, but the word seems unusual.

draft-ietf-tls-tls13.md
ekr: It's also in the intro. Tightened that a bit

draft-ietf-tls-tls13.md
ekr: Not sure.

draft-ietf-tls-tls13.md
ekr: done

draft-ietf-tls-tls13.md
ekr: I use it elsewhere.

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+We assume that the attacker has complete control of the network in
+between the parties {{RFC3552}}. Even under these conditions, the
+handshake should provide the properties listed below. Note that
+these properties are not necessarily independent, but reflect
+the protocol consumers' needs.
+
+Establishing the same session key.
+: The handshake needs to output the same session key on both sides of the
+handshake (See {{CK01}}; defn 1, part 1).
+
```
wbl: If successful! An attacker can always disrupt the handshake, if only by cutting the link midway through. Just because one side completed successfully does not mean the other side did so.

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+type of channel because even a constant time padding removal
+function will then feed the content into data-dependent functions.
+
+Generation N+1 keys are derived from generation N keys via a key
+derivation function {{updating-traffic-keys}}. As long as this function is truly one way, it
+is not possible to compute the previous keys after a key change
+(forward secrecy). However, TLS does not provide backward secrecy
+after key updates; systems which want backward secrecy must do
+a fresh handshake and establish a new session key with an (EC)DHE
+exchange.
```
wbl: Maybe define backwards secrecy similarly to how forward secrecy was defined. It took me a moment to figure out what was meant.

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+: An attacker should not be able to craft a new record which is
+different from an existing record which will be accepted by the receiver
+with more than negligible probability.
+{::comment}Cite INT-CTXT?{:/comment}
+
+Order protection/non-replayability
+: An attacker should not be able to cause the receiver to accept a
+record which it has already accepted or cause the receiver to accept
+record N+1 without having first processed record N.
+
```
wbl: Even for DTLS? If we're not folding it in, then this looks good, but if we are we need a more complex definition.

ekr: I think a TODO is in order for that one :)

ekr:
@wbl you happy?

wbl:
Yup.

draft-ietf-tls-tls13.md
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+: The cryptographic parameters should be the same on both sides and
+should be the same as if the peers had been communicating in the
+absence of an attack (See {{BBFKZG16}}; defns 8 and 9}).
+
+Forward secret
+: If the long-term authentication keys (signature keys in certificate-based
+authentication modes or the PSK in PSK-(EC)DHE modes) are compromised after
+the handshake is complete, this does not compromise the security of the
+session key (See {{DOW92}}).
+
```
tvdmerwe: I suggest rephrasing slightly.
``` If the long-term keying material (in this case the signature keys in certificate-based ... ```

**Comment by ekr** [comment]
Merged separately.

File: draft-ietf-tls-tls13.md [hunk 11]
```diff
@@ -4128,9 +4212,174 @@ Section 7 of [RFC6066], as it is not applicable to AEAD ciphers and has
...
+of a given record. More formally it should not be able to distinguish
+between multiple encryptions of the same versus different plaintexts.
+{::comment}Cite IND-CPA?{:/comment}
+
+Integrity.
+: An attacker should not be able to craft a new record which is
+different from an existing record which will be accepted by the receiver
+with more than negligible probability.
+{::comment}Cite INT-CTXT?{:/comment}
+
```
tvdmerwe: Again, as this is intended to be informal, I suggest removing "with more than negligible probability". Perhaps add 
``` For a more formal definition see (reference).```

PR #536, opened 2016-07-11
Use definition lists more consistently

martinthomson:
The definitions of terms didn't use a definition list.

Most of the other definition lists didn't include any delimiter other than whitespace, which isn't clear in some renderings.

This fixes both of those.


ekr:
Not sure about adding the colons. Something happened to the toolchain here; previous versions out-dented the term and then broke the line. I think rather than adding colons I'd rather go back to that style if I (or someone else) can figure out what we need to feed to cramdown

ekr:
I may deal with this in a later editorial pass but this has bit rotted enough to not be something we could apply, so closing

PR #544, opened 2016-07-12
Remove server missing_extension requirements.

davidben:
(This is orthogonal to whether alerts should be MUST or SHOULD. I've
left them as MUST for now.)

While the client can reasonably send missing_extension in response to a
ServerHello which, say, selected an ECDHE cipher without key_share, the
server is in no such position. Such a requirement is unhelpful as a
natural implementation would have the same behavior (connection failed)
anyway.

Since servers already must account for some ciphers requiring ECDHE and
some not, a server may implement logic such as:

    ecdheOk := tryToSelectAGroup(myList, clientList)
    for cipher in serverCiphers {
        if cipher not in clientCiphers {
            continue
        }
        if isECDHE(cipher) and not ecdheOk {
            continue
        }
        ... other checks, etc. ...
        yay, I found a cipher
    }
    if no cipher was found {
        send handshake_failure and close connection
    }

There is no reason to ask a server to distinguish between the following
cases, all of which will fail anyway:

- The client sent only ECDHE ciphers I know about but did not send
  key_share.
- The client sent only ECDHE ciphers I know about but only key_shares I
  don't implement.
- The client sent only ECDHE ciphers, some of which I don't know about,
  and did not send key_share.
- The client sent ECDHE ciphers I know about and some PSK ciphers I
  don't know about and did not send key_share.
- The client sent only ECDHE ciphers, none of which I know about, and
  did not send key_share.

Notably, the some cases are not even distinguishable, even though the
old text mandates different alerts for them. The only case we can
reasonably send missing_extension for is the first, but this is an
unhelpful special-case to mandate. It only adds complexity (the
connection will fail anyway), thus risking bugs.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1652,8 +1652,6 @@ which signature algorithms may be used in digital signatures.
 
 Clients which offer one or more cipher suites which use certificate authentication
 (i.e., any non-PSK cipher suite) MUST send the "signature_algorithms" extension.
-If this extension is not provided and no alternative cipher suite is available,
-the server MUST close the connection with a fatal "missing_extension" alert.
 (see {{mti-extensions}})
 
 The "extension_data" field of this extension contains a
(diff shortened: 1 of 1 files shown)


davegarrett:
Ditching an alert warrants a mailing list discussion, rather than just filing a PR. Please propose this there.

The purpose of proper error reporting is to actually catch bugs and fix issues, rather than get a potentially useless reply that causes people to just avoid new versions. It is not a complex requirement to have implementations close connections with appropriate error codes rather than a generic one. If you wish to not keep track of why the handshake is failing, just reassess after actually hitting the error: If you'd normally do a handshake_failure and you found any cipher that looks like it might be viable, but the needed extension is not provided, then send the specific alert instead.

davidben:
Sure thing. Let's continue this on the list.

ekr:
This is mostly overtaken by the refactor of negotiation. @davidben please provide another PR for what is left if you want.

davidben:
Nah, I think I'm good with the current state of missing_extensions. Actually the alert is basically gone by now.

PR #549, opened 2016-07-16
Require the CertificateRequest context to be empty in the handshake.

davidben:
It's only useful during the handshake, so avoid forcing the client save
that field for the in-handshake version.


ekr:
I went back and forth on this, but on balance I think you're right.

PR #559, opened 2016-07-25
WIP: Refactor negotiation as discussed in Berlin

ekr:
This is just a WIP and needs review.

@martinthomson @grittygrease @davidben PTAL.

@davidben: I know you don't like the indication from the server but would like your opinion on other aspects.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3687,18 +3687,9 @@ the value is computed as:
 ##  MTI Cipher Suites
 
 In the absence of an application profile standard specifying otherwise, a
-TLS-compliant application MUST implement the following cipher suites:
-
-~~~~
-    TLS_AES_128_GCM_SHA256
-~~~~
-
-A TLS-compliant application SHOULD implement the following cipher suites:
-
-~~~~
-    TLS_AES_256_GCM_SHA384
-    TLS_CHACHA20_POLY1305_SHA256
-~~~~
+TLS-compliant application MUST implement the TLS_AES_128_GCM_SHA256
+cipher suite and SHOULD implement the TLS_AES_256_GCM_SHA384 and 
+TLS_CHACHA20_POLY1305_SHA256 cipher suites.
 
 A TLS-compliant application MUST support digital signatures with
 rsa_pkcs1_sha256 (for certificates), rsa_pss_sha256 (for
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1265,6 +1265,53 @@ The key exchange messages are used to exchange security capabilities
 between the client and server and to establish the traffic keys used to protect
 the handshake and data.
 
+
+### Cryptographic Negotiation
+
+TLS cryptographic negotiation proceeds by the client offering the
+following four sets of options in its ClientHello.
+
+- A list of cipher suites which indicates the AEAD cipher/HKDF hash
+  pairs which the client supports
+- A "supported_group" ({{negotiated-groups}}) extension which indicates the (EC)DHE groups
+  which the client supports and a "key_share" ({{key-share}}) extension which contains
+  (EC)DHE shares for some or all of these groups
+- A "signature_algorithms" ({{signature-algorithms}}) extension which indicates the signature
+  algorithms which the client can accept.
+- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains the identities
+  of symmetric keys known to the client and the key exchange
+  modes which each PSK supports.
+
+If the server does not select a PSK, then the first three of these
+options are entirely orthogonal: the server independently selects a
+cipher suite, an (EC)DHE group and key share for key establishment,
+and a signature algorithm/certificate pair to authenticate itself.
+If there is no overlap on any of these parameters, then the handshake
+will fail. If there is overlap in the "supported_group" extension
+but the client did not offer a compatible "key_share" extension,
+then the server will respond with a HelloRetryRequest ({{hello-retry-request}}) message.
+
+If the server selects a PSK, then the PSK will indicate which key
+establishment modes it can be used with (PSK alone or (EC)DHE-PSK)
+and which authentication modes it can be used with (PSK alone or
+PSK with signatures). The server can then select those parameters
+to be consistent both with the PSK and the other extensions supplied
+by the client. Note that if the PSK can be used without (EC)DHE
+or without signatures, then non-overlap in either of these parameters
+need not be fatal.
+
+The server indicates its selected parameters in the ServerHello
+as follows. If PSK is being used then the server will send a
+"pre_shared_key" extension indicating the selected key. The
+extension will also contain indicators of whether (EC)DHE and/or
+certificate-based authentication are in use. If PSK is not being
+used, then (EC)DHE and certificate-based authentication are always
+used. When (EC)DHE is in use, the server will also provide a "key_share"
+extension. When authenticating via a certificate, the server will
+also send Certificate ({{certificate}}) and CertificateVerify
+({{certificate-verify}}) messages.
+
+
 ###  Client Hello
 
 When this message will be sent:
@@ -1518,41 +1565,6 @@ server's ServerHello/KeyShare, the client MUST verify that
 the selected NamedGroup matches that supplied in
 the HelloRetryRequest. If these values differ, the client
 MUST abort the connection with a fatal "handshake_failure" alert.
-
-## Cryptographic Negotiation
-
-TLS cryptographic negotiation proceeds by the client offering the
-following four sets of options:
-
-- A list of cipher suites which indicates the AEAD cipher/HKDF hash
-  pairs which the client supports
-- A "supported_group" ({{negotiated-groups}}) extension which indicates the (EC)DHE groups
-  which the client supports and a "key_share" ({{key-share}}) extension which contains
-  (EC)DHE shares in some or all of these groups
-- A "signature_algorithms" ({{signature-algorithms}}) extension which indicates the signature
-  algorithms which the client can accept.
-- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contain the identities
-  of symmetric keys known to the client and the key exchange
-  modes which each PSK supports.
-
-If the server does not select a PSK, then the first three of these
-options are entirely orthogonal: the server independently selects a
-cipher suite, an (EC)DHE group and key share for key establishment,
-and a signature algorithm/certificate pair to authenticate itself.
-If there is no overlap on any of these parameters, then the handshake
-will fail. If there is overlap in the "supported_group" extension
-but the client did not offer a compatible "key_share" extension,
-then the server will respond with a HelloRetryRequest message.
-
-If the server selects a PSK, then the PSK will indicate which key
-establishment modes it can be used with (PSK alone or (EC)DHE-PSK)
-and which authentication modes it can be used with (PSK alone or
-PSK with signatures). The server can then select those parameters
-to be consistent both with the PSK and the other extensions supplied
-by the client. Note that if the PSK can be used without (EC)DHE
-or without signatures, then non-overlap in either of these parameters
-need not be fatal.
-
 
 ##  Hello Extensions
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -678,11 +678,6 @@ TLS supports three basic key exchange modes:
 
 - A combination of a symmetric key and Diffie-Hellman
 
-Which mode is used depends on the negotiated cipher suite. Conceptually,
-the handshake establishes three secrets which are used to derive all the
-keys.
-
-
 {{tls-full}} below shows the basic full TLS handshake:
 
 ~~~
@@ -734,21 +729,22 @@ in the diagram above.
 
 In the Key Exchange phase, the client sends the ClientHello
 ({{client-hello}}) message, which contains a random nonce
-(ClientHello.random), its offered protocol version, cipher suite, and
-extensions, and in general either one or more Diffie-Hellman key shares (in the
+(ClientHello.random), its offered protocol version, a list of
+symmetric ciphers, some set of Diffie-Hellman key shares (in the
 "key_share" extension {{key-share}}), one or more pre-shared key labels (in the
-"pre_shared_key" extension {{pre-shared-key-extension}}), or both.
+"pre_shared_key" extension {{pre-shared-key-extension}}), or both,
+and some extensions.
 
 The server processes the ClientHello and determines the appropriate
 cryptographic parameters for the connection. It then responds with its
 own ServerHello which indicates the negotiated connection
-parameters. [{{server-hello}}].  The combination of the ClientHello
-and the ServerHello determines the shared keys. If either a pure
-(EC)DHE or (EC)DHE-PSK cipher suite is in use, then the ServerHello
+parameters. [{{server-hello}}]. The combination of the ClientHello
+and the ServerHello determines the shared keys. If pure
+(EC)DHE or (EC)DHE-PSK key establishment is in use, then the ServerHello
 will contain a "key_share" extension with the server's ephemeral
 Diffie-Hellman share which MUST be in the same group as one of the
-client's shares. If a pure PSK or an (EC)DHE-PSK cipher suite is
-negotiated, then the ServerHello will contain a "pre_shared_key"
+client's shares. If pure PSK or (EC)DHE-PSK key establishment is
+in use, then the ServerHello will contain a "pre_shared_key"
 extension indicating which of the client's offered PSKs was selected.
 
 The server then sends two messages to establish the Server Parameters:
@@ -768,19 +764,16 @@ Specifically:
 
 Certificate.
 : the certificate of the endpoint. This message is omitted if the
-  server is not authenticating with a certificate (i.e.,
-  with PSK or (EC)DHE-PSK cipher suites). Note that if raw public keys
-  {{RFC7250}} or the cached information extension
-  {{?I-D.ietf-tls-cached-info}} are in use, then this message
-  will not contain a certificate but rather some other value
-  corresponding to the server's long-term key.
-  [{{certificate}}]
+  server is not authenticating with a certificate. Note that if raw
+  public keys {{RFC7250}} or the cached information extension
+  {{?I-D.ietf-tls-cached-info}} are in use, then this message will not
+  contain a certificate but rather some other value corresponding to
+  the server's long-term key.  [{{certificate}}]
 
 CertificateVerify.
 : a signature over the entire handshake using the public key
   in the Certificate message. This message is omitted if the
-  server is not authenticating via a certificate (i.e.,
-  with PSK or (EC)DHE-PSK cipher suites). [{{certificate-verify}}]
+  server is not authenticating via a certificate. [{{certificate-verify}}]
 
 Finished.
 : a MAC (Message Authentication Code) over the entire handshake.
@@ -854,10 +847,9 @@ below, this functionality was provided by "session IDs" and
 "session tickets" {{RFC5077}}. Both mechanisms are obsoleted in TLS
 1.3.
 
-PSK cipher suites can either use PSK in combination with
-an (EC)DHE exchange in order to provide forward secrecy in combination
-with shared keys, or can use PSKs alone, at the cost of losing forward
-secrecy.
+PSKs can be used with (EC)DHE exchange in order to provide forward
+secrecy in combination with shared keys, or can be used alone, at the
+cost of losing forward secrecy.
 
 {{tls-resumption-psk}} shows a pair of handshakes in which the first establishes
 a PSK and the second uses it:
@@ -901,9 +893,11 @@ Subsequent Handshake:
 As the server is authenticating via a PSK, it does not send a
 Certificate or a CertificateVerify. When a client offers resumption
 via PSK it SHOULD also supply a "key_share" extension to the server as well
-to allow the server to decline resumption and fall back to a full handshake, if needed.
-A "key_share" extension MUST also be sent if the client is attempting to
-negotiate an (EC)DHE-PSK cipher suite.
+to allow the server to decline resumption and fall back to a full handshake,
+if needed. The server responds with a "pre_shared_key" extension
+to negotiate use of PSK key establishment and can (as shown here)
+respond with a "key_share" extension to do (EC)DHE key
+establishment, thus providing forward secrecy.
 
 
 ## Zero-RTT Data
@@ -958,9 +952,8 @@ server authenticating via a signature
 https://github.com/tlswg/tls13-spec/issues/443]]
 
 
-IMPORTANT NOTE: The security properties for 0-RTT data (regardless of
-the cipher suite) are weaker than those for other kinds of TLS data.
-Specifically:
+IMPORTANT NOTE: The security properties for 0-RTT data are weaker than
+those for other kinds of TLS data.  Specifically:
 
 1. This data is not forward secret, because it is encrypted solely
 with the PSK.
@@ -1345,16 +1338,13 @@ legacy_session_id
   set by a pre-TLS 1.3 server.
 
 cipher_suites
-: This is a list of the cryptographic options supported by the client,
-  with the client's first preference first.  Each cipher suite defines
-  a key exchange algorithm, a record protection algorithm (including
-  secret key length) and a hash to be used with HKDF. The server will
-  select a cipher suite or, if no acceptable choices are presented,
-  return a "handshake_failure" alert and close the connection. If the
-  list contains cipher suites the server does not recognize, support,
-  or wish to use, the server MUST ignore those cipher suites, and
-  process the remaining ones as usual.  Values are defined in
-  {{cipher-suites}}.
+: This is a list of the symmetric cipher options supported by the
+  client, specifically the record protection algorithm (including
+  secret key length) and a hash to be used with HKDF, in descending
+  order of client preference. If the list contains cipher suites
+  the server does not recognize, support, or wish to use, the server
+  MUST ignore those cipher suites, and process the remaining ones as
+  usual. Values are defined in {{cipher-suites}}.
 
 legacy_compression_methods
 : Versions of TLS before 1.3 supported compression with the list of
@@ -1422,9 +1412,7 @@ random
 
 cipher_suite
 : The single cipher suite selected by the server from the list in
-  ClientHello.cipher_suites.  For resumed sessions, this field is
-  the value from the state of the session being resumed.
-  [[TODO: interaction with PSK.]]
+  ClientHello.cipher_suites.
 
 extensions
 : A list of extensions.  Note that only extensions offered by the
@@ -1491,7 +1479,6 @@ Structure of this message:
 
        struct {
            ProtocolVersion server_version;
-           CipherSuite cipher_suite;
            NamedGroup selected_group;
            Extension extensions<0..2^16-1>;
        } HelloRetryRequest;
@@ -1501,11 +1488,9 @@ selected_group
   is requesting a retried ClientHello/KeyShare for.
 {:br }
 
-The server_version, cipher_suite, and extensions fields have the
+The server_version and extensions fields have the
 same meanings as their corresponding values in the ServerHello.
-[[NOTE: cipher_suite may disappear. https://github.com/tlswg/tls13-spec/issues/528]]
-The
-server SHOULD send only the extensions necessary for the client to
+The server SHOULD send only the extensions necessary for the client to
 generate a correct ClientHello pair (currently no such extensions
 exist). As with ServerHello, a
 HelloRetryRequest MUST NOT contain any extensions that were not first
@@ -1530,7 +1515,7 @@ in its original KeyShare.
 
 Upon re-sending the ClientHello and receiving the
 server's ServerHello/KeyShare, the client MUST verify that
-the selected CipherSuite and NamedGroup match that supplied in
+the selected NamedGroup matches that supplied in
 the HelloRetryRequest. If either of these values differ, the client
 MUST abort the connection with a fatal "handshake_failure" alert.
 
@@ -1648,16 +1633,14 @@ Clients MUST NOT use cookies in subsequent connections.
 ###  Signature Algorithms
 
 The client uses the "signature_algorithms" extension to indicate to the server
-which signature algorithms may be used in digital signatures.
-
-Clients which offer one or more cipher suites which use certificate authentication
-(i.e., any non-PSK cipher suite) MUST send the "signature_algorithms" extension.
-If this extension is not provided and no alternative cipher suite is available,
+which signature algorithms may be used in digital signatures. If a server
+is authenticating via a certificate and the client has not sent a
+"signature_algorithms" extension then the server MUST
 the server MUST close the connection with a fatal "missing_extension" alert.
 (see {{mti-extensions}})
 
-The "extension_data" field of this extension contains a
-"supported_signature_algorithms" value:
+The "extension_data" field of this extension contains a "supported_signature_algorithms"
+value:
 
 %%% Signature Algorithm Extension
 
@@ -1737,10 +1720,6 @@ EdDSA algorithms
   the "PureEdDSA" algorithms and not the "prehash" variants.
 {:br }
 
-The semantics of this extension are somewhat complicated because the cipher
-suite adds additional constraints on signature algorithms.
-{{server-certificate-selection}} describes the appropriate rules.
-
 rsa_pkcs1_sha1, dsa_sha1, and ecdsa_sha1 SHOULD NOT be offered. Clients
 offering these values for backwards compatibility MUST list them as the lowest
 priority (listed after all other algorithms in the
@@ -1781,19 +1760,6 @@ Note: In versions of TLS prior to TLS 1.3, this extension was named
 {{I-D.ietf-tls-negotiated-ff-dhe}}. This extension was also used to negotiate
 ECDSA curves. Signature algorithms are now negotiated independently (see
 {{signature-algorithms}}).
-
-Clients which offer one or more (EC)DHE cipher suites MUST send at least one
-supported NamedGroup value and servers MUST NOT negotiate any of these
-cipher suites unless a supported value was provided.
-If this extension is not provided and no alternative cipher suite is available,
-the server MUST close the connection with a fatal "missing_extension" alert.
-(see {{mti-extensions}})
-If the extension is provided, but no compatible group is offered, the
-server MUST NOT negotiate a cipher suite of the relevant type. For
-instance, if a client supplies only ECDHE groups, the server MUST NOT
-negotiate finite field Diffie-Hellman. If no acceptable group can be
-selected across all cipher suites, then the server MUST generate a
-fatal "handshake_failure" alert.
 
 The "extension_data" field of this extension contains a
 "NamedGroupList" value:
@@ -1854,14 +1820,6 @@ subsequent connections.
 The "key_share" extension contains the endpoint's cryptographic parameters
 for non-PSK key establishment methods (currently DHE or ECDHE).
 
-Clients which offer one or more (EC)DHE cipher suites MUST send this
-extension and SHOULD send at least one supported KeyShareEntry value.
-Servers MUST NOT negotiate any of these cipher suites unless a supported
-value was provided.
-If this extension is not provided in a ServerHello or ClientHello,
-and the peer is offering (EC)DHE cipher suites, then the endpoint MUST close
-the connection with a fatal "missing_extension" alert.
-(see {{mti-extensions}})
 Clients MAY send an empty client_shares vector in order to request
 group selection from the server at the cost of an additional round trip.
 (see {{hello-retry-request}})
@@ -1908,7 +1866,8 @@ client_shares
   in the "supported_groups" extension.
 
 server_share
-: A single KeyShareEntry value for the negotiated cipher suite.
+: A single KeyShareEntry value that is in the same group as one of the
+  client's shares.
 {:br }
 
 Clients offer an arbitrary number of KeyShareEntry values, each
@@ -1916,25 +1875,16 @@ representing a single set of key exchange parameters. For instance, a
 client might offer shares for several elliptic curves or multiple
 FFDHE groups.  The key_exchange values for each KeyShareEntry MUST by
 generated independently.  Clients MUST NOT offer multiple
-KeyShareEntry values for the same group.  Clients MUST NOT offer
-any KeyShareEntry values for groups not listed in the client's
-"supported_groups" extension.
+KeyShareEntry values for the same group.  Clients MUST NOT offer any
+KeyShareEntry values for groups not listed in the client's
+"supported_groups" extension.  Servers MAY check for violations of
+these rules and and MAY abort the connection with a fatal
+"illegal_parameter" alert if one is violated.
 
-Servers offer exactly one KeyShareEntry value, which corresponds to
-the key exchange used for the negotiated cipher suite.  Servers MUST
-NOT offer a KeyShareEntry value for a group not offered by the client
-in its corresponding KeyShare or "supported_groups" extension.
-
-Implementations MAY check for violations of these rules and
-and MAY abort the connection with a fatal "illegal_parameter" alert
-if one is violated.
-
-If the server selects an (EC)DHE cipher suite and no mutually
-supported group is available between the two endpoints' KeyShare
-offers, yet there is a mutually supported group that can be found via
-the "supported_groups" extension, then the server MUST reply with a
-HelloRetryRequest.  If there is no mutually supported group at all,
-the server MUST NOT negotiate an (EC)DHE cipher suite.
+Servers offer exactly one KeyShareEntry which corresponds to one of
+the KeyShareEntries offered by the client. Servers MUST NOT send
+a KeyShareEntry for any group not indicated in the "supported_groups"
+extension.
 
 [[TODO: Recommendation about what the client offers.
 Presumably which integer DH groups and which curves.]]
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -734,21 +729,22 @@ in the diagram above.
 
 In the Key Exchange phase, the client sends the ClientHello
 ({{client-hello}}) message, which contains a random nonce
-(ClientHello.random), its offered protocol version, cipher suite, and
-extensions, and in general either one or more Diffie-Hellman key shares (in the
+(ClientHello.random), its offered protocol version, a list of
+symmetric ciphers, some set of Diffie-Hellman key shares (in the
```
martinthomson: PRF hash?

draft-ietf-tls-tls13.md
```diff
@@ -3872,91 +3828,35 @@ and renegotiation_info {{RFC5746}}.
...
-All cipher suites in this section are specified for use with both TLS 1.2
-and TLS 1.3, as well as the corresponding versions of DTLS.
-(see {{backward-compatibility}})
+represents 256-bit AES in the GCM mode of operation. 
+
+|       Cipher Suite Name         |    Value    | Specification |
+|:--------------------------------|:------------|:--------------|
+| TLS13_AES_128_GCM_SHA256        | {0x13,0x01} |   [This RFC]  |
+| TLS13_AES_256_GCM_SHA384        | {0x13,0x02} |   [This RFC]  |
+| TLS13__CHACHA20_POLY1305_SHA256 | {0x13,0x02} |   [This RFC]  |
```
davidben: Nit: extra underscore

draft-ietf-tls-tls13.md
```diff
@@ -3872,91 +3828,35 @@ and renegotiation_info {{RFC5746}}.
...
+   CipherSuite TLS13_CIPHER_HASH = VALUE;
 ~~~
 
 | Component | Contents |
 |:----------|:---------|
-| TLS       | The string "TLS" |
-| KEA       | The key exchange algorithm (e.g. ECDHE, DHE) |
-| AUTH      | The authentication algorithm (e.g. certificates, PSK) |
-| WITH      | The string "WITH" |
+| TLS13     | The string "TLS13" |
```
davidben: Do we expect these to be used for TLS 1.4 as well? Wondering if we should pick some more generic name like `TLS_AEAD_AES_128_GCM_SHA256` or maybe just `TLS_AES_128_GCM_SHA256`.

draft-ietf-tls-tls13.md
```diff
@@ -1854,14 +1820,6 @@ subsequent connections.
 The "key_share" extension contains the endpoint's cryptographic parameters
 for non-PSK key establishment methods (currently DHE or ECDHE).
 
-Clients which offer one or more (EC)DHE cipher suites MUST send this
-extension and SHOULD send at least one supported KeyShareEntry value.
-Servers MUST NOT negotiate any of these cipher suites unless a supported
-value was provided.
-If this extension is not provided in a ServerHello or ClientHello,
```
davidben: I think the ServerHello clause is still valid since the server PSK extension (or lack of it) is the canonical handshake shape in this design.

ekr: Hmm.... How do you know if the other side is offering an (EC)DHE cipher suite

draft-ietf-tls-tls13.md
```diff
@@ -1908,33 +1866,25 @@ client_shares
...
+KeyShareEntry values for the same group.  Clients MUST NOT offer any
+KeyShareEntry values for groups not listed in the client's
+"supported_groups" extension.  Servers MAY check for violations of
+these rules and and MAY abort the connection with a fatal
+"illegal_parameter" alert if one is violated.
+
+Servers offer exactly one KeyShareEntry which corresponds to one of
+the KeyShareEntries offered by the client. Servers MUST NOT send
+a KeyShareEntry for any group not indicated in the "supported_groups"
+extension.
```
davidben: A few things got lost in this version:

* Server sends HelloRetryRequest if key share isn't there.
* If no mutually-supported group at all, server is only allowed to pick a pure PSK mode.

Also, it wasn't in the old one, but "exactly one KeyShareEntry" suggests that pure PSK mode isn't allowed. (Unless that's implied by text elsewhere.)

draft-ietf-tls-tls13.md
davidben: Right, hence just the ServerHello clause. ClientHello clause was always kind of silly. But when receiving a ServerHello, if the server PSK extension says `psk_dhe_ke` or is missing, then we should fail the connection if `key_share` is missing. I.e. there is redundancy in the ServerHello.

draft-ietf-tls-tls13.md
ekr: Good point. That was the nice thing about "TLS_STAR" :)

draft-ietf-tls-tls13.md
ekr: I see what you mean. But it needs a rewrite.

draft-ietf-tls-tls13.md
martinthomson: Do we even need to pretend that there is a structure to the name?

ekr: I do think it's nice to indicate the break between AEAD and HKDF

draft-ietf-tls-tls13.md
```diff
@@ -1530,7 +1515,7 @@ in its original KeyShare.
 
 Upon re-sending the ClientHello and receiving the
 server's ServerHello/KeyShare, the client MUST verify that
-the selected CipherSuite and NamedGroup match that supplied in
+the selected NamedGroup matches that supplied in
 the HelloRetryRequest. If either of these values differ, the client
```
grittygrease: "either of these values differ" -> "this value differs"

draft-ietf-tls-tls13.md
grittygrease: is "cipher suite" officially a symmetric cipher and an HKDF hash now?

draft-ietf-tls-tls13.md
```diff
@@ -2105,7 +2084,7 @@ client's "pre_shared_key" extension. In addition, it MUST verify that
 the following values are consistent with those negotiated in the
 connection during which the ticket was established.
 
-- The TLS version number, symmetric ciphersuite, and the hash for HKDF.
+- The TLS version number, symmetric cipher suite, and the hash for HKDF.
```
grittygrease: HDKF hash is implicit in the cipher suite now, if I'm reading the definition correctly.

draft-ietf-tls-tls13.md
```diff
@@ -2746,13 +2711,18 @@ from the resumption master secret:
...
-The client MAY use this PSK for future handshakes by including
-the ticket value in the "pre_shared_key" extension in its ClientHello
-({{pre-shared-key-extension}}) and supplying a suitable PSK cipher
-suite. Servers may send multiple tickets on a single connection, for
-instance after post-handshake authentication. For handshakes that
-do not use a resumption_psk, the resumption_context is a string of
-Hash.Length zeroes.
+The client MAY use this PSK for future handshakes by including the
+ticket value in the "pre_shared_key" extension in its ClientHello
+({{pre-shared-key-extension}}). Servers may send multiple tickets on a
```
grittygrease: Where are tickets used in post-handshake authentication? Multiple tickets are useful for unlinkability of simultaneous resumption requests, but not post-handshake auth.

ekr: I'm having trouble reading this text. Can you rephrase?

draft-ietf-tls-tls13.md
kaduk: That's my understanding of the plan.

draft-ietf-tls-tls13.md
```diff
@@ -1648,16 +1633,14 @@ Clients MUST NOT use cookies in subsequent connections.
 ###  Signature Algorithms
 
 The client uses the "signature_algorithms" extension to indicate to the server
-which signature algorithms may be used in digital signatures.
-
-Clients which offer one or more cipher suites which use certificate authentication
-(i.e., any non-PSK cipher suite) MUST send the "signature_algorithms" extension.
```
kaduk: Is it a good idea to drop the normative restriction on the client?

kaduk: That is, the condition for when the client MUST send signature_algorithms is different, but if the client wants the server to authenticate with a cert, they had better send it.

draft-ietf-tls-tls13.md
```diff
@@ -1908,33 +1866,25 @@ client_shares
...
-HelloRetryRequest.  If there is no mutually supported group at all,
-the server MUST NOT negotiate an (EC)DHE cipher suite.
+KeyShareEntry values for the same group.  Clients MUST NOT offer any
+KeyShareEntry values for groups not listed in the client's
+"supported_groups" extension.  Servers MAY check for violations of
+these rules and and MAY abort the connection with a fatal
+"illegal_parameter" alert if one is violated.
+
+Servers offer exactly one KeyShareEntry which corresponds to one of
+the KeyShareEntries offered by the client. Servers MUST NOT send
```
kaduk: This particular wording makes it sound unimportant, like the server should pick one at random to reply to.
Maybe "[...] which corresponds to one of the KeyShareEntries offered by the client, that the server has selected for the negotiated key exchange."

draft-ietf-tls-tls13.md
```diff
@@ -2021,25 +1974,51 @@ selected_identity
...
-The client MUST verify that the server's selected_identity
-is within the range supplied by the client. If the server supplies an
-"early_data" extension, the client MUST verify that the server selected the
-first offered identity. If any other value is returned, the client MUST
-generate a fatal "unknown_psk_identity" alert and close the connection.
+Each PSK offered by the client also indicates the authentication and
+key exchange modes with which the server can use it, with each
+list being in the order of the client's preference, with most
+preferred first. 
+PskKeyExchangeModes have the following meanings:
```
kaduk: Blank line before this?

draft-ietf-tls-tls13.md
```diff
@@ -2624,18 +2596,11 @@ for a server CertificateVerify would be:
...
-extension before selecting them. This is somewhat inelegant but is a compromise
-designed to minimize changes to the original cipher suite design.
-
-If sent by a client, the signature algorithm used in the
-signature MUST be one of those present in the
-supported_signature_algorithms field of the CertificateRequest
-message.
+produced without unsupported algorithms (see {{signature-algorithms}}).
+
+If sent by a client, the signature algorithm used in the signature
```
kaduk: I thought if the client didn't send a signature_algorithms, the server was supposed to abort.

draft-ietf-tls-tls13.md
kaduk: @grittygrease the text about "after post-handshake auth" is referring to how the state the server wants to keep in the ticket changes when the client does post-handshake authentication, so the server probably wants to issue new tickets with that information.  The tickets are not used during post-handshake auth, but there is a reason to regenerate tickets after post-handshake auth.

draft-ietf-tls-tls13.md
```diff
@@ -2746,13 +2711,18 @@ from the resumption master secret:
...
-the ticket value in the "pre_shared_key" extension in its ClientHello
-({{pre-shared-key-extension}}) and supplying a suitable PSK cipher
-suite. Servers may send multiple tickets on a single connection, for
-instance after post-handshake authentication. For handshakes that
-do not use a resumption_psk, the resumption_context is a string of
-Hash.Length zeroes.
+The client MAY use this PSK for future handshakes by including the
+ticket value in the "pre_shared_key" extension in its ClientHello
+({{pre-shared-key-extension}}). Servers may send multiple tickets on a
+single connection, for instance after post-handshake
```
kaduk: It's unclear if this example is trying to count the "multiple tickets" as "I sent one at the main handshake, and I sent one more after the post-handshake client auth", or sending multiple tickets right next to each other (or both).

draft-ietf-tls-tls13.md
```diff
@@ -2746,13 +2711,18 @@ from the resumption master secret:
...
+authentication. For handshakes that do not use a resumption_psk, the
+resumption_context is a string of [[TBD: This is not safe now
+that we allow additional server signatures with PSK: 
+OPEN ISSUE https://github.com/tlswg/tls13-spec/issues/558]]
+
+Any ticket MUST only be resumed with a cipher suite that is identical
```
kaduk: Just to double-check, we can get away with this now because the PSK state is not part of the cipher suite, and we don't have to worry about taking the original cipher suite and converting it to its PSK analogue?

draft-ietf-tls-tls13.md
davidben: Not if both ends are configured to do everything via PSKs rather than certificates. Unfortunately, the semantics of these cases are pretty fuzzy. This needs to be spelled out much more clearly about what the presence and absence of the extension means and the consequences on the "default PSK-less handshake shape" since the text goes with the more complex server indication scheme.

draft-ietf-tls-tls13.md
```diff
@@ -2793,43 +2760,36 @@ ticket_lifetime
...
-
 ticket
 : The value of the ticket to be used as the PSK identifier.
 The ticket itself is an opaque label. It MAY either be a database
 lookup key or a self-encrypted and self-authenticated value. Section
 4 of {{RFC5077}} describes a recommended ticket construction mechanism.
+
+
+ticket_extensions
+: A placeholder for extensions in the ticket. Clients MUST ignore
```
kaduk: Is it still a placeholder now that an extension is defined?

draft-ietf-tls-tls13.md
```diff
@@ -2793,43 +2760,36 @@ ticket_lifetime
...
-  suite.
-{:br }
 
-In all cases, the PSK or (EC)DHE-PSK cipher suites that the client
-offers/uses MUST have the same symmetric parameters (cipher/hash) as
-the cipher suite negotiated for this connection. If no flags are set
-that the client recognizes, it MUST ignore the ticket.
+ticket_age_add
+: A randomly generated 32-bit value that is used to obscure the age of the ticket that the
+  client includes in the "early_data" extension.  The actual ticket age is
```
kaduk: Do we want to clarify that "actual" means "as measured by the client"?

draft-ietf-tls-tls13.md
kaduk: Right.  But everything is PSK, there's no CertificateVerify.  So, I still think the conditional is not needed.  (Maybe the text needs more of a rewrite than just that.)

draft-ietf-tls-tls13.md
kaduk: Nit: chacha/poly != aes256/gcm

draft-ietf-tls-tls13.md
martinthomson: Can we instead use "AEAD algorithm"?

draft-ietf-tls-tls13.md
ekr: I change TLS13_ to TLS_!

draft-ietf-tls-tls13.md
martinthomson: Is this supposed to be the same paragraph or a different one?

draft-ietf-tls-tls13.md
ekr: This text is about the client's cert, not te sig-alg extension

draft-ietf-tls-tls13.md
ekr: Right. That's why "identical"

ekr:
@martinthomson @davidben ptal

draft-ietf-tls-tls13.md
```diff
@@ -1272,6 +1265,53 @@ The key exchange messages are used to exchange security capabilities
...
+- A "signature_algorithms" ({{signature-algorithms}}) extension which indicates the signature
+  algorithms which the client can accept.
+- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains the identities
+  of symmetric keys known to the client and the key exchange
+  modes which each PSK supports.
+
+If the server does not select a PSK, then the first three of these
+options are entirely orthogonal: the server independently selects a
+cipher suite, an (EC)DHE group and key share for key establishment,
+and a signature algorithm/certificate pair to authenticate itself.
```
kaduk: "to use to authenticate itself to the client"?

draft-ietf-tls-tls13.md
```diff
@@ -1272,6 +1265,53 @@ The key exchange messages are used to exchange security capabilities
...
+If there is no overlap on any of these parameters, then the handshake
```
kaduk: "If any of these parameters has no overlap between the values supported by the client and server" (the current text could be misread as requiring all parameters to have no overlap before handshake failure).

draft-ietf-tls-tls13.md
```diff
@@ -1272,6 +1265,53 @@ The key exchange messages are used to exchange security capabilities
...
+will fail. If there is overlap in the "supported_group" extension
+but the client did not offer a compatible "key_share" extension,
+then the server will respond with a HelloRetryRequest ({{hello-retry-request}}) message.
+
+If the server selects a PSK, then the PSK will indicate which key
+establishment modes it can be used with (PSK alone or (EC)DHE-PSK)
+and which authentication modes it can be used with (PSK alone or
+PSK with signatures). The server can then select those parameters
```
kaduk: Maybe expand "those" to "key exchange and authentication"?

draft-ietf-tls-tls13.md
```diff
@@ -1648,16 +1680,16 @@ Clients MUST NOT use cookies in subsequent connections.
 ###  Signature Algorithms
 
 The client uses the "signature_algorithms" extension to indicate to the server
-which signature algorithms may be used in digital signatures.
-
-Clients which offer one or more cipher suites which use certificate authentication
-(i.e., any non-PSK cipher suite) MUST send the "signature_algorithms" extension.
-If this extension is not provided and no alternative cipher suite is available,
+which signature algorithms may be used in digital signatures. Clients which
+desire the server to authenticate via a certificate MUST send this algorithm.
```
kaduk: s/algorithm/extension/?

draft-ietf-tls-tls13.md
```diff
@@ -1345,16 +1385,13 @@ legacy_session_id
...
-  a key exchange algorithm, a record protection algorithm (including
-  secret key length) and a hash to be used with HKDF. The server will
-  select a cipher suite or, if no acceptable choices are presented,
-  return a "handshake_failure" alert and close the connection. If the
-  list contains cipher suites the server does not recognize, support,
-  or wish to use, the server MUST ignore those cipher suites, and
-  process the remaining ones as usual.  Values are defined in
-  {{cipher-suites}}.
+: This is a list of the symmetric cipher options supported by the
+  client, specifically the record protection algorithm (including
```
kaduk: Hmm, maybe we don't need to explicitly say "AEAD" here.

draft-ietf-tls-tls13.md
```diff
@@ -2746,13 +2761,24 @@ from the resumption master secret:
...
-the ticket value in the "pre_shared_key" extension in its ClientHello
-({{pre-shared-key-extension}}) and supplying a suitable PSK cipher
-suite. Servers may send multiple tickets on a single connection, for
-instance after post-handshake authentication. For handshakes that
-do not use a resumption_psk, the resumption_context is a string of
-Hash.Length zeroes.
+The client MAY use this PSK for future handshakes by including the
+ticket value in the "pre_shared_key" extension in its ClientHello
+({{pre-shared-key-extension}}). Servers MAY send multiple tickets on a
+single connection, either immediately after each other or
```
kaduk: Something about the either/or here is bothering me, though I don't quite have a concrete suggestion. (The potential wrong implication being that only one ticket may be sent after any given specific even.)

draft-ietf-tls-tls13.md
ekr: I didn't

draft-ietf-tls-tls13.md
ekr: Dealt with this by adding a whole section on how to negotiate.

draft-ietf-tls-tls13.md
```diff
@@ -1782,19 +1811,6 @@ Note: In versions of TLS prior to TLS 1.3, this extension was named
...
-cipher suites unless a supported value was provided.
-If this extension is not provided and no alternative cipher suite is available,
-the server MUST close the connection with a fatal "missing_extension" alert.
-(see {{mti-extensions}})
-If the extension is provided, but no compatible group is offered, the
-server MUST NOT negotiate a cipher suite of the relevant type. For
-instance, if a client supplies only ECDHE groups, the server MUST NOT
-negotiate finite field Diffie-Hellman. If no acceptable group can be
-selected across all cipher suites, then the server MUST generate a
-fatal "handshake_failure" alert.
```
davidben: It's probably worth being clear on what missing this extension means (I've seen TLS 1.2 implementation which send no curves and expect to do ECDHE, though hopefully this will be noticed by servers not tolerating it anymore). Maybe something like:

Clients which offer (EC)DHE or (EC)DHE-PSK key establishment MUST send at least one supported NamedGroup value. Servers MUST NOT negotiation an (EC)DHE group unless a supported value was provided. Note that, in that case, pure PSK key establishment is still possible.

draft-ietf-tls-tls13.md
```diff
@@ -1272,6 +1265,54 @@ The key exchange messages are used to exchange security capabilities
...
+as follows. If PSK is being used then the server will send a
+"pre_shared_key" extension indicating the selected key. The
+extension will also contain indicators of whether (EC)DHE and/or
+certificate-based authentication are in use. If PSK is not being
+used, then (EC)DHE and certificate-based authentication are always
+used. When (EC)DHE is in use, the server will also provide a "key_share"
+extension. When authenticating via a certificate, the server will
+also send Certificate ({{certificate}}) and CertificateVerify
+({{certificate-verify}}) messages.
+
```
davidben: Perhaps move most of the various handshake_failure texts scattered elsewhere out and move them up here, since failure to negotiate parameters is really a property of the overall negotiation algorithm. Something like:

If the server is unable to negotiate a supported set of parameters, it MUST return a "handshake_failure" alert and close the connection.

draft-ietf-tls-tls13.md
```diff
@@ -2247,8 +2277,8 @@ extensions
 
 When this message will be sent:
 
-> A non-anonymous server can optionally request a certificate from the client,
-if appropriate for the selected cipher suite. This message, if sent, will
+> A server which is authenticating with a certificate can optionally
```
davidben: There's another mention of anonymous at the end of the extension that probably should be correspondingly tweaked.

draft-ietf-tls-tls13.md
```diff
@@ -1908,33 +1915,26 @@ client_shares
...
-the "supported_groups" extension, then the server MUST reply with a
-HelloRetryRequest.  If there is no mutually supported group at all,
-the server MUST NOT negotiate an (EC)DHE cipher suite.
+KeyShareEntry values for the same group.  Clients MUST NOT offer any
+KeyShareEntry values for groups not listed in the client's
+"supported_groups" extension.  Servers MAY check for violations of
+these rules and and MAY abort the connection with a fatal
+"illegal_parameter" alert if one is violated.
+
+If using (EC)DHE key establishment, servers offer exactly one
```
davidben: Nit: Elsewhere the text tends to refer to "(EC)DHE key establishment" as a distinct pre-composed thing from "(EC)DHE-PSK key establishment" and lists both. (Though that is a bit wordy so maybe we want to change that? Anyway, better be consistent one way or another.)

draft-ietf-tls-tls13.md
```diff
@@ -1908,33 +1915,26 @@ client_shares
...
+KeyShareEntry which corresponds to one of the KeyShareEntries offered
```
davidben: Super nit-picky nit: there is no such defined type KeyShareEntries. Maybe "KeyShareEntry values".

draft-ietf-tls-tls13.md
davidben: Oh, I see this is largely already covered in the section below.

draft-ietf-tls-tls13.md
```diff
@@ -2021,25 +2024,52 @@ selected_identity
...
+preferred first.
+
+PskKeyExchangeModes have the following meanings:
+
+psk_ke
+: PSK-only key establishment. In this mode, the server MUST not
+supply a "key_share" value.
+
+psk_dhe_ke
+: PSK key establishment with (EC)DHE key establishment. In this mode,
```
davidben: Nit: Elsewhere this is just referred to as (EC)DHE-PSK key establishment.

draft-ietf-tls-tls13.md
```diff
@@ -1985,27 +1985,30 @@ The "pre_shared_key" extension is used to indicate the identity of the
 pre-shared key to be used with a given handshake in association
 with a PSK or (EC)DHE-PSK cipher suite (see {{RFC4279}} for background).
```
davidben: Nit: Another cipher suite -> key establishment

draft-ietf-tls-tls13.md
davidben: Reading more, perhaps we should indeed switch terminology to the decomposed form. Though, if decomposed, "(EC)DHE key establishment" and "PSK key establishment" sound mutually exclusive when they're not. Maybe rather than giving it a shared noun, talk about whether PSK was used/negotiated and whether (EC)DHE was used/negotiated? Dunno.

draft-ietf-tls-tls13.md
```diff
@@ -2021,25 +2024,52 @@ selected_identity
...
+PskAuthenticationModes have the following meanings:
+
+psk_auth
+: PSK-only authentication. In this mode, the server MUST NOT supply
+either a Certificate or CertificateVerify message.
+
+psk_sign_auth
+: PSK authentication plus a digital signature from the server. In this
+mode, the server MUST send Certificate ({{certificate}}) and CertificateVerify
+({{certificate-verify}}) messages.
```
davidben: I think this wants a TON more text and clarification as to semantics and client/server policies. I'd probably even suggest we add this in as a separate change. This PR is already quite large. Right now it's primarily editorial stuff. If we also throw in PSK+Sign, then it becomes both a large bunch of editorial changes throughout the document plus thinking about some stuff like:

* What does it mean if the certificate changes?
* When should the client offer one more or another?
* How should the server decide which it prefers?
* Does this give us the overall behavior we want?

The awkward thing about PSK vs PSK+Sign compared to all our other negotiations is this varies over time. PSK+Sign's only purpose in life is to arrange for NewSessionTicket's lifetime to be uncapped. Which means that if the lifetime isn't close to expiring, both sides would prefer normal PSK just so they have to do less work (particularly servers). Only when it gets close to the client (and only the client; the server doesn't gain any direct benefit from signing again) wanting a renewal, do we want to do a PSK+Sign and switch back.

draft-ietf-tls-tls13.md
davidben: I think it's still missing client behavior around:

* Server signaled it wanted to do ECDHE but we didn't get a key_share.
* Server signaled it didn't want to do ECDHE but we did get a key_share.

(This is the cost of doing the more redundant PSK-based server indication rather than using ServerHello extensions.)

draft-ietf-tls-tls13.md
davidben: Actually, didn't we decide in Berlin not to allow PSK + CertificateRequest? (Or maybe that was 0-RTT + CertificateRequest. One of these.) Or am I misremembering? By adding PSK + Sign in this PR, this text now allows it.

siyengar:
could we change NamedGroup to be called KeyExchangeAlgorithm instead. It seems like that's what it technically is after this change and will allow us to support other kinds of exchanges in the future like Post quantum algorithms in a more semantically friendly way than calling them a "named group"

draft-ietf-tls-tls13.md
```diff
@@ -1272,6 +1265,54 @@ The key exchange messages are used to exchange security capabilities
...
+establishment modes it can be used with (PSK alone or (EC)DHE-PSK)
+and which authentication modes it can be used with (PSK alone or
+PSK with signatures). The server can then select those parameters
+to be consistent both with the PSK and the other extensions supplied
+by the client. Note that if the PSK can be used without (EC)DHE
+or without signatures, then non-overlap in either of these parameters
+need not be fatal.
+
+The server indicates its selected parameters in the ServerHello
+as follows. If PSK is being used then the server will send a
```
davegarrett: nit: should be a colon instead of a period after "as follows"

draft-ietf-tls-tls13.md
```diff
@@ -1530,8 +1563,8 @@ in its original KeyShare.
 
 Upon re-sending the ClientHello and receiving the
 server's ServerHello/KeyShare, the client MUST verify that
-the selected CipherSuite and NamedGroup match that supplied in
-the HelloRetryRequest. If either of these values differ, the client
+the selected NamedGroup matches that supplied in
+the HelloRetryRequest. If these values differ, the client
 MUST abort the connection with a fatal "handshake_failure" alert.
```
davegarrett: nit: doesn't need a separate sentence anymore; can just be:
"
Upon re-sending the ClientHello and receiving the
server's ServerHello/KeyShare, the client MUST verify that
the selected NamedGroup matches the one supplied in
the HelloRetryRequest, and MUST abort the connection
with a fatal "handshake_failure" alert if it does not.
"

draft-ietf-tls-tls13.md
davegarrett: Yes, this needs to be corrected to "extension" here.

draft-ietf-tls-tls13.md
```diff
@@ -1648,16 +1681,16 @@ Clients MUST NOT use cookies in subsequent connections.
...
-
-Clients which offer one or more cipher suites which use certificate authentication
-(i.e., any non-PSK cipher suite) MUST send the "signature_algorithms" extension.
-If this extension is not provided and no alternative cipher suite is available,
+which signature algorithms may be used in digital signatures. Clients which
+desire the server to authenticate via a certificate MUST send this algorithm.
+If a server
+is authenticating via a certificate and the client has not sent a
+"signature_algorithms" extension then the server MUST
 the server MUST close the connection with a fatal "missing_extension" alert.
```
davegarrett: duplicate "the server MUST" here

draft-ietf-tls-tls13.md
```diff
@@ -3872,91 +3885,35 @@ and renegotiation_info {{RFC5746}}.
...
+   CipherSuite TLS13_CIPHER_HASH = VALUE;
 ~~~
 
 | Component | Contents |
 |:----------|:---------|
-| TLS       | The string "TLS" |
-| KEA       | The key exchange algorithm (e.g. ECDHE, DHE) |
-| AUTH      | The authentication algorithm (e.g. certificates, PSK) |
-| WITH      | The string "WITH" |
+| TLS     | The string "TLS13" |
```
davegarrett: need to change the "TLS" here to "TLS13"

draft-ietf-tls-tls13.md
```diff
@@ -3872,91 +3885,35 @@ and renegotiation_info {{RFC5746}}.
...
-All cipher suites in this section are specified for use with both TLS 1.2
-and TLS 1.3, as well as the corresponding versions of DTLS.
-(see {{backward-compatibility}})
+represents 256-bit AES in the GCM mode of operation. 
+
+|       Cipher Suite Name         |    Value    | Specification |
+|:--------------------------------|:------------|:--------------|
+| TLS_AES_128_GCM_SHA256        | {0x13,0x01} |   [This RFC]  |
+| TLS_AES_256_GCM_SHA384        | {0x13,0x02} |   [This RFC]  |
+| TLS_CHACHA20_POLY1305_SHA256  | {0x13,0x03} |   [This RFC]  |
```
davegarrett: what about AES CCM? it is standards track

draft-ietf-tls-tls13.md
```diff
@@ -3872,91 +3885,35 @@ and renegotiation_info {{RFC5746}}.
...
+
+Although TLS 1.3 uses the same cipher suite space as previous versions
+of TLS, TLS 1.3 cipher suites are defined differently, only specifying
+the symmetric ciphers, and cannot it be used for TLS 1.2. Similarly,
+TLS 1.2 and lower cipher suites cannot be used with TLS 1.3
```
davegarrett: missing period

draft-ietf-tls-tls13.md
davegarrett: it's not a nit-pick; searching the doc for specific names is a normal use-case, which this breaks; the correct label needs to be used here

draft-ietf-tls-tls13.md
davegarrett: or, rather, change the "TLS13" here and above to simply "TLS" to match the naming below; either way, one needs to be used in all places in this section

draft-ietf-tls-tls13.md
ekr: Right. @martinthomson given the way that your post-handshake server auth draft is heading, I'd like to go back to the idea of using SignatureAlgorithms to indicate what the server did. Objections?

draft-ietf-tls-tls13.md
ekr: Fixed, I believe.

draft-ietf-tls-tls13.md
ekr: Added here. I'll make another pass to remove the redundant ones.

draft-ietf-tls-tls13.md
ekr: No, you're right. Any thoughts on how to phrase it less clunky than "is authenticating with a certificate and has not accepted a PSK"?

draft-ietf-tls-tls13.md
ekr: My plan is to get rid of (EC)DHE-PSK

draft-ietf-tls-tls13.md
ekr: I am fixing

draft-ietf-tls-tls13.md
ekr: I can split this into a separate change with a TODO, especially if you help write the text for the policy :)

draft-ietf-tls-tls13.md
ekr: I'm willing to have someone make a case for it.

draft-ietf-tls-tls13.md
hannestschofenig: Why isn't CCM included in this list? 
Could you add CCM_8 as well?

ekr:
@davidben @martinthomson I changed the server's indicator to be SignatureAlgorithms. PTAL.

ekr:
P.S. Also removed the server signing with PSK mode. Will move that to a separate PR.

draft-ietf-tls-tls13.md
```diff
@@ -1272,6 +1267,56 @@ The key exchange messages are used to exchange security capabilities
...
+establishment modes it can be used with (PSK alone or with (EC)DHE) and
+which authentication modes it can be used with (PSK alone or PSK with
+signatures). The server can then select those key establishment and
+authentication parameters to be consistent both with the PSK and the
+other extensions supplied by the client. Note that if the PSK can be
+used without (EC)DHE or without signatures, then non-overlap in either
+of these parameters need not be fatal.
+
+The server indicates its selected parameters in the ServerHello as
+follow: If PSK is being used then the server will send a
```
davegarrett: should be "follows"; lost an 's' here

draft-ietf-tls-tls13.md
```diff
@@ -1345,16 +1390,13 @@ legacy_session_id
...
-  or wish to use, the server MUST ignore those cipher suites, and
-  process the remaining ones as usual.  Values are defined in
-  {{cipher-suites}}.
+: This is a list of the symmetric cipher options supported by the
+  client, specifically the record protection algorithm (including
+  secret key length) and a hash to be used with HKDF, in descending
+  order of client preference. If the list contains cipher suites
+  the server does not recognize, support, or wish to use, the server
+  MUST ignore those cipher suites, and process the remaining ones as
+o  usual. Values are defined in {{cipher-suites}}.
```
davegarrett: a 'o' snuck in here somehow at the beginning of the line

draft-ietf-tls-tls13.md
```diff
@@ -1648,16 +1686,19 @@ Clients MUST NOT use cookies in subsequent connections.
 ###  Signature Algorithms
 
 The client uses the "signature_algorithms" extension to indicate to the server
-which signature algorithms may be used in digital signatures.
+which signature algorithms may be used in digital signatures. Clients which
+desire the server to authenticate via a certificate MUST send this extension.
+If a server
+is authenticating via a certificate and the client has not sent a
+"signature_algorithms" extension then the server MUST
+close the connection with a fatal "missing_extension" alert.
```
davegarrett: extra period here; already one after the parenthetical

draft-ietf-tls-tls13.md
```diff
@@ -1691,7 +1732,7 @@ The "extension_data" field of this extension contains a
            obsolete_RESERVED (0x0000..0x0200),
            obsolete_RESERVED (0x0204..0x0400),
            obsolete_RESERVED (0x0404..0x0500),
-           obsolete_RESERVED (0x0504..0x0600),
+cd            obsolete_RESERVED (0x0504..0x0600),
```
davegarrett: "cd" added to beginning of line for some reason

draft-ietf-tls-tls13.md
davegarrett: Well, using these new cipher suite codepoints is required to use these ciphers with TLS 1.3+, so not having one for a cipher that was compatible before this rewrite is equivalent to banning said cipher. I am stating no opinion on whether CCM should be used, in general, however it is listed as standards track and was previously in the list here, so I don't see a reason it should be dropped in the new system. If we are to change support, I would recommend it be done in a separate PR, as this is supposed to be a negotiation change here. I just think not having a codepoint for it here requires a justification, not lacking one. Probably warrants a mailing list discussion.

ekr: On Thu, Aug 11, 2016 at 5:56 PM, Dave Garrett <notifications@github.com>
wrote:

> In draft-ietf-tls-tls13.md
> <https://github.com/tlswg/tls13-spec/pull/559#discussion_r74529138>:
>
> > -are being or will be used for interop testing and therefore are likely to be
> > -assigned.
> > -
> > -Note: [RFC6655] is inconsistent with respect to the ordering of components
> > -within PSK AES CCM cipher suite names. The names above are as defined.
> > -
> > -All cipher suites in this section are specified for use with both TLS 1.2
> > -and TLS 1.3, as well as the corresponding versions of DTLS.
> > -(see {{backward-compatibility}})
> > +represents 256-bit AES in the GCM mode of operation.
> > +
> > +|       Cipher Suite Name         |    Value    | Specification |
> > +|:--------------------------------|:------------|:--------------|
> > +| TLS_AES_128_GCM_SHA256        | {0x13,0x01} |   [This RFC]  |
> > +| TLS_AES_256_GCM_SHA384        | {0x13,0x02} |   [This RFC]  |
> > +| TLS_CHACHA20_POLY1305_SHA256  | {0x13,0x03} |   [This RFC]  |
>
> Well, using these new cipher suite codepoints is required to use these
> ciphers with TLS 1.3+, so not having one for a cipher that was compatible
> before this rewrite is equivalent to banning said cipher.
>
We're no more banning it than we are banning CAMELLIA, which also can in
principle be used with TLS 1.3 but which we're not defining a code point
for. The relevant question is whether it's desirable to define a code point
for CCM in *this* document.

I am stating no opinion on whether CCM should be used, in general, however

ekr: Regardless, I'll add CCM and CCM_8 for now.

draft-ietf-tls-tls13.md
```diff
@@ -3690,23 +3690,22 @@ In the absence of an application profile standard specifying otherwise, a
...
 A TLS-compliant application SHOULD implement the following cipher suites:
 
 ~~~~
-    TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
-    TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
-    TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
-    TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
+    TLS_AES_256_GCM_SHA384
+    TLS_CHACHA20_POLY1305_SHA256
 ~~~~
```
davegarrett: Now that things are simpler, this is overly complex formatting. I suggest simply having things inline (as was done in previous TLS docs). e.g.:

"
TLS-compliant application MUST implement TLS_AES_128_GCM_SHA256
cipher suite. A TLS-compliant application SHOULD also implement the
TLS_AES_256_GCM_SHA384 and TLS_CHACHA20_POLY1305_SHA256
cipher suites.
"

If this change isn't made, there's a nit that the first list is now just one entry, so "suites" should be "suite".

draft-ietf-tls-tls13.md
martinthomson: How about this:

```
struct {
  select (role) {
    case server:
      bool signed;
    case client:
  };
  SignatureScheme supported_schemes<0..2^16-2>;
} SignatureAlgorithms;
```

I don't think that it makes sense to put the actual scheme in here.  I would like to have the supported schemes available for those doing client auth; it's neatly symmetrical that way and lets us use this properly in the spontaneous case without having to invent a parallel structure.  The only downside is that you have to save this stuff in case you have a CertificateVerify.

If you care about that last point, you could make the `bool` a mode switch:

```
enum {
  no_signature, server_signature, both_signature
} AuthenticationMode;
```

Which would let you signal that there is a CertificateVerify on the way.

ekr: On Thu, Aug 11, 2016 at 11:23 PM, Martin Thomson <notifications@github.com>
wrote:

> In draft-ietf-tls-tls13.md
> <https://github.com/tlswg/tls13-spec/pull/559#discussion_r74546042>:
>
> > @@ -1854,14 +1820,6 @@ subsequent connections.
> >  The "key_share" extension contains the endpoint's cryptographic parameters
> >  for non-PSK key establishment methods (currently DHE or ECDHE).
> >
> > -Clients which offer one or more (EC)DHE cipher suites MUST send this
> > -extension and SHOULD send at least one supported KeyShareEntry value.
> > -Servers MUST NOT negotiate any of these cipher suites unless a supported
> > -value was provided.
> > -If this extension is not provided in a ServerHello or ClientHello,
>
> How about this:
>
> struct {
>   select (role) {
>     case server:
>       bool signed;
>     case client:
>   };
>   SignatureScheme supported_schemes<0..2^16-2>;
> } SignatureAlgorithms;
>
>
I'm not sure I understand the semantics of this proposal. I assume
supported_schemes is
"this is what I would accept".  "signed" is "I, server, am going to sign"?
But then I don't understand
"AuthenticationMode" below.

>  I don't think that it makes sense to put the actual scheme in here. I
> would like to have the supported schemes available for those doing client
> auth; it's neatly symmetrical that way and lets us use this properly in the
> spontaneous case without having to invent a parallel structure.
>
Well, in both cases you have to say "I would accept spontaneous auth",
right?


The only downside is that you have to save this stuff in case you have a
> CertificateVerify.
>
> If you care about that last point, you could make the bool a mode switch:
>
> enum {
>   no_signature, server_signature, both_signature
> } AuthenticationMode;
>
> Which would let you signal that there is a CertificateVerify on the way.
>

You lost me here.

draft-ietf-tls-tls13.md
KATOAkihiro: >We're no more banning it than we are banning CAMELLIA, which also can in
 >principle be used with TLS 1.3 but which we're not defining a code point
for.
Then I 'm going to ask you for a challenge. Can you merge two code point about Camellia followings:

TLS_CAMELLIA_128_GCM_SHA256
TLS_CAMELLIA_256_GCM_SHA384

beurdouche: You definitely should ask that on the TLS mailing list. Having Camellia in the main RFC will probably be controversial.

ekr: Right. Camellia is not standards track. Feel free to ask for the code point assignment separately.

PR #561, opened 2016-07-25
Rename various version fields.

davidben:
From the past three times we've bumped the version, reality has
unequivocally told us that the TLS version negotiation, as described in
previous specifications, has been a collossal failure.

The working group decided not to replace the mechanism with a more
sustainable one, so do the next best thing in trying to save TLS from
itself. If the fields are renamed to what they actually are
('server_version' is the true version and 'client_version' is a
maximum), then perhaps we'll have some hope of future implementors
getting it right, if we ever clear through the prior iterations of the
bug.

While I'm here, remove the mention of 'The protocol version the current
record is compatible with' for record_version. record_version is purely
vestigial.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2936,7 +2936,7 @@ Alert messages ({{alert-protocol}}) MUST NOT be fragmented across records.
 
        struct {
            ContentType type;
-           ProtocolVersion record_version = { 3, 1 };    /* TLS v1.x */
+           ProtocolVersion legacy_record_version = { 3, 1 };    /* TLS v1.x */
            uint16 length;
            opaque fragment[TLSPlaintext.length];
        } TLSPlaintext;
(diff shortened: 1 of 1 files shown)


kaduk:
+1

davegarrett:
This sounds like a good idea, however for the server version I'd go with "negotiated_version" to be more precise. For "record_version", we could also extend the "legacy" prefix convention we've started using and change it to "legacy_record_version" to remind that it's a deprecated field.

davidben:
> negotiated_version

Hrm. We don't name other fields like ServerHello.negotiated_cipher_suite and such. Though if we wanted to be consistent with those fields, it should be ClientHello.max_version rather than ClientHello.max_supported_version. Dunno, maybe that would be better... max_supported_version is a bit wordy.

(Though I do think breaking pattern on the ClientHello version is justified because that field is already inconsistent with everything else in TLS. The version is the only place where the client doesn't offer a list, so everything else already self-describes as a "pick one of these" type of field.)

> legacy_record_version.

Done.

Issue #543, opened 2016-07-12
Ilari ]


ekr:

ekr:
Copying here so they don't get lost.

On Mon, Jul 11, 2016 at 12:08:00PM -0700, Eric Rescorla wrote:
> Folks,
>
> I've just submitted draft-ietf-tls-tls13-14.txt and it should
> show up on the draft repository shortly. In the meantime you
> can find the editor's copy in the usual location at:

> As usual, comments welcome.
> -Ekr

Did a readthrough, here's a bunch of comments (didn't check the
issues list):

-----------------------------------------------------------------------
> ## Zero-RTT Data
>
>
> [[OPEN ISSUE: Should it be possible to combine 0-RTT with the
> server authenticating via a signature
> https://github.com/tlswg/tls13-spec/issues/443]]

One should note that it is not clear what sort of continuity properties
should be provoded:

- Certificate valid for hostname
- The same key as in last time
- The same certificate as in last time

The last choice obviously limits lifetime of derived PSK to
the parent certificate.

Also, extension matching is quite different between those three (and the
last two are rather different from the current rules, tuned for 0-RTT
only being used with PSK.


> ##  Numbers
>
> Note that in some cases (e.g., DH parameters) it is necessary to represent
> integers as opaque vectors. In such cases, they are represented as unsigned
> integers (i.e., additional leading zero octets are not used even if the most
> significant bit is set).

Well, I think the DH parameters are gone, but DH public keys are still there,
and those AFAIK are padded with zeroes.

> ###  Client Hello
>
> When this message will be sent:
>
> > When a client first connects to a server, it is required to send the
> ClientHello as its first message. The client will also send a
> ClientHello when the server has responded to its ClientHello with a
> ServerHello that selects cryptographic parameters that don't match the
> client's "key_share" extension. In that case, the client MUST send the same
> ClientHello (without modification) except:

Should the "ServerHello" be "HelloRetryRequest"?

> - Including a new KeyShareEntry as the lowest priority share
>   (i.e., appended to the list of shares in the "key_share" extension).
>
> - Removing the EarlyDataIndication {{early-data-indication}} extension
>   if one was present. Early data is not permitted after HelloRetryRequest.

Also, adding cookie extension if that was present in HRR?

BTW, with the early_data context and cross-connection cookies gone, I
have figured out how to delta-compress the first client_hello and its
rejection against to-be-sent second client hello plus connection
parameters server selected into 6 bytes(!) of space (not including
[...]

PR #562, opened 2016-07-27
Don't talk about rolling your own PRNG

kaduk:
Instead, talk about not rolling your own PRNG.  This document is
not the place to discuss how to do so, and RFC 40986 is a fine reference.


davegarrett:
Cleaning up the very old info here is in the list in issue #543.

This PR looks good to me. Recommending implementers to not reinvent the wheel is better than trying to cram anything useful in here, and we already reference RFC4086.

PR #565, opened 2016-07-28
Forbid post-handshake handshake/application interleave.

davidben:
No sender could possibly need to send half of a message interleaved with
application data in TLS, so forbid this. This cuts down on the
state-space a receiver needs to worry about.

This also means one cannot force the peer to buffer all but one byte of
a giant post-handshake message, and then never complete it for the
lifetime of the connection. (Although it does not remove all
steady state buffering DoS-type concerns.)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2729,9 +2729,9 @@ TLS also allows other messages to be sent after the main handshake.
 These messages use a handshake content type and are encrypted under the application
 traffic key.
 
-Handshake messages sent after the handshake MUST NOT be interleaved with
-application data. That is, if a message is split over two or more handshake
-records, there MUST NOT be any application data records between them.
+Handshake messages sent after the handshake MUST NOT be interleaved with other
+record types. That is, if a message is split over two or more handshake
+records, there MUST NOT be any other records between them.
 
 ### New Session Ticket Message {#NewSessionTicket}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2729,6 +2729,10 @@ TLS also allows other messages to be sent after the main handshake.
 These messages use a handshake content type and are encrypted under the application
 traffic key.
 
+Handshake messages sent after the handshake MUST NOT be interleaved with
+application data. That is, if a message is split over two or more handshake
+records, there MUST NOT be any application data records between them.
```
kaduk: This sort of implicitly assumes that there are only application data or handshake records and no other types ... which is basically true for this purpose, but why not just say "MUST NOT be any other records between them"?

davidben: Good point. What about alerts? I could imagine wanting to allow sending:

```
[hs:NewSessio-]
[alert:internal_error]
--- connection closed ---
```

Maybe it should say non-alert records? I can't imagine any use for sticking an alert in the middle other than the server spontaneously combusting mid-send and wanting to tell you about it. Though I'm also happy just saying all records because this case is silly.

kaduk: I think we're in the "all alerts are fatal alerts" world, now, so isn't it illegal to send the other part of the handshake message after the alert?

davidben: Oh, good point. This specifically says "interleave" which is distinct from "interrupt". Okay, I'll switch it to any alert.

davidben: Done

PR #566, opened 2016-07-28
Codify receiver obligations on epoch changes.

davidben:
TLS is stuck with the legacy of a layer of indirection between messages
and records. This means that message boundaries and record boundaries
need not be correlated at all, which has historically been the cause of
bugs.

Among other problems, correlating messages with key changes is subtle.
Receivers MUST ensure that there is no buffered handshake data cross
record boundaries. Otherwise one can inject prefixes authenticated under
one epoch into another. Write down these rules.

Note: the spelling of these rules is very TLS-centric. DTLS will need a
very different spelling of the rules. (It's not as straight-forward as
record boundaries. One must make sure there is no partial message data
beyond the current message.) Since DTLS 1.3 does not exist yet, I've
left that alone. We should extract a common "transport" or "dispatch"
layer interface between DTLS/TLS for the protocol to sit on top of for
that.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1387,6 +1387,10 @@ formats; if not, then it MUST send a fatal "decode_error" alert.
 After sending the ClientHello message, the client waits for a ServerHello
 or HelloRetryRequest message.
 
+The next message the client sends will be under a different key, so the server
+MUST check that the end of the ClientHello message aligns with a record
+boundary.
+
 ###  Server Hello
 
 When this message will be sent:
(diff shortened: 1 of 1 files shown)


kaduk:
What should the receiver do if the message does not end on a record boundary?

davidben:
Shut down the connection. Will add a sentence. (Not sure if we should say which alert. There's a lot of different implementation strategies for this sort of thing, particularly once DTLS is involved. Some cases naturally hit one set of alerts and some others.)

kaduk:
Thanks.

ekr:
I agree with these requirements, but I think I would prefer to have a single requirement that any time a key changes you must align rather than have boilerplate. Would you like to take a stab at that or should I?

PR #567, opened 2016-07-28
Supply guidance on the magnitude of replay window

kaduk:
Just saying "small allowance" is not very clear, so supply some
concrete text with an arbitrarily selected value for discussion.

Some (but not much) previous discussion in the vicinity of https://www.ietf.org/mail-archive/web/tls/current/msg20566.html

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2244,8 +2244,10 @@ hidden by TCP.
 
 A small allowance for errors in clocks and variations in measurements
 is advisable.  However, any allowance also increases the opportunity
-for replay.  In this case, it is better to reject early data than to
-risk greater exposure to replay attacks.
+for replay.  In this case, it is better to reject early data and fall back
+to a full 1-RTT handshake than to risk greater exposure to replay attacks,
+so in common network topologies, a small allowance on the order of ten
+seconds is reasonable.
 
 ### OCSP Status Extensions
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2245,9 +2245,10 @@ hidden by TCP.
 A small allowance for errors in clocks and variations in measurements
 is advisable.  However, any allowance also increases the opportunity
 for replay.  In this case, it is better to reject early data and fall back
-to a full 1-RTT handshake than to risk greater exposure to replay attacks,
-so in common network topologies, a small allowance on the order of ten
-seconds is reasonable.
+to a full 1-RTT handshake than to risk greater exposure to replay attacks.
+In common network topologies for browser clients, small allowances on the
+order of ten seconds are reasonable.  Clock skew distributions are not
+symmetric, so the optimal tradeoff may involve an asymmetric replay window.
 
 ### OCSP Status Extensions
 


kaduk:
See also issue #543

martinthomson:
Not sure that tens of seconds is good advice.  It's certainly larger than I had imagined for a general purpose server.

If we accept that clock drift is in the order of seconds per day, then single-digit seconds is probably OK.

You don't need to have a high rate of 0-RTT success toward the end of the ticket validity period (1 week), you only need to ensure that most users get 0-RTT if they come back a short time (measured in hours) after the initial session.

kaduk:
The reference point was the 15 seconds that Ilari had used (but still seemed large to me).
It is still unclear to me what metric we actually want to use to bound the replay window (i.e., whether it is guided by clock rate skew, RTT, absolute time, or something else), so this text is mostly just a discussion starter.  Is one second too small?

draft-ietf-tls-tls13.md
```diff
@@ -2183,8 +2183,10 @@ hidden by TCP.
 
 A small allowance for errors in clocks and variations in measurements
 is advisable.  However, any allowance also increases the opportunity
-for replay.  In this case, it is better to reject early data than to
-risk greater exposure to replay attacks.
+for replay.  In this case, it is better to reject early data and fall back
+to a full 1-RTT handshake than to risk greater exposure to replay attacks,
+so in common network topologies, a small allowance on the order of ten
+seconds is reasonable.
```
ekr: This seems very high. The difference here is causes by errors in RTT estimation and clock speed differentials, which should not be ten seconds over a week.

@siyengar do you have any estimates of how fast client clocks change?

siyengar: Clock skew distributions are skewed more heavily negative than positive, so we should recommend people to set 2 different bounds. This guidance could probably be written as "a small allowance is reasonable, for example a lower bound of X and and an upper bound of Y". 

@ekr Don't have numbers on how frequently clocks change, but probably more interestingly, enforcing a window of 120s (-120 to +120), cut off 0.1% of 0-RTT connections (which would then fallback to 1rtt)

If we were to enforce a window of 10s (-10 to +10), that would additionally cut off another 0.3% 0rtt connections.

ekr: How many of those hosts would reject certs anyway? I.e., how many clocks slew 120s over a week but aren't many days off in absolute time?

ekr: @siyengar

siyengar: At the present moment since we've started enforcing limits, so I can't look that up presently since any numbers over > 120s are biased. 

The numbers from 120s to 10s though are unbiased. Even after excluding clients whose absolute time [-10, 10] days we still get about 0.3% of 0rtt lost over a week. 

Also android is worse than ios for drift.

martinthomson: @siyengar, over what period are your numbers taken?  I wouldn't mind losing 0.3% of 0-RTT given that we have a seamless fallback.  If this is over a day, that suggests we could have a narrower tolerance for drift than even 10s.  If this is just based on attempts at 0-RTT and is uncorrelated to the time between ticket issuance and use, then I'm not sure how to interpret the data in a way that would help us make a better decision.

siyengar: @martinthomson The data over a period of a week. The results hold even over a larger time window. 

The assumption of the ticket age method is that the clock skew remains the same over time. I can't tell you exactly how often clocks drift. However I can say how much 0rtt we would lose if we set the client skew to 10s, or any X. So when we receive a 0RTT attempt, if we enforced 10s as the ticket age skew, then how much 0rtt would we end up rejecting from client that were within a reasonable absolute client time. Does that clear things up?

The time window will be server specific. Any number we write down is not enforceable in the protocol, so we're just writing down a recommended number. Servers can tune this as they want based on their tolerance and different servers have different amounts of 0rtt traffic they can lose. I'm just noting at what level we've chosen to draw that line which is 0.1%. You might be fine with 0.3%. We're also re-evaluating that line with some other experiments.

About to get on a flight so will be AWOL for a few hours.

martinthomson: That's helpful in giving us numbers, but I'm not 100% clear still.  Maybe I'll ask differently.

I'm interested in the drift rate, which sounds like it might in the range of 0-200 milliseconds per day for the bulk of your clients.  You keep using absolute times, which leads me to assume that you are looking at the window you are enforcing on your servers.  That doesn't tell me much about the statistics of actual drift rates, since many clients will return within much shorter times than the week.

Am I right in saying that you have tickets that are valid for a week, and you have the time set to 120s, which gives you a 0.1% 0-RTT rejection rate (similarly for 10s and 0.3%)?

I understand that you have your own constraints when choosing what numbers to pick, but I would suggest that we advise servers to pick a lower number (which results in a higher reject rate), simply because anything we say is an endorsement of something.  Endorsing a replay window of 4 minutes is probably not quite conservative enough for general advice.  If Facebook has a much better handle on their management of replays than in the general case, a larger window is reasonable for them.

knekritz: Among mobile clients resuming with a ticket that was issued within the past day, we see about .1% outside a window of +/- 30s, .4% outside +/- 10s, and 2% outside +/- 2s (this is among clients that can successfully validate our certificate).

That said, I don't think it makes sense to set a very small time limit without also preventing clients from slow trickling data in after the client hello. Setting a very small limit as is will prevent legitimate clients from doing 0-RTT while leaving a loophole open for malicious clients to extend the replay window. "Tens of seconds" sounds like reasonable guidance to me.

ekr: @knekritz: thanks. @kaduk can you update this PR to where you're happy if needed and I will merge

siyengar: it would also be useful to mention that they should consider and lower and upper bound separately. This will help implementations provide the right api.

kaduk:
I updated the text some (and rebased); hopefully it is an improvement.

draft-ietf-tls-tls13.md
```diff
@@ -2244,8 +2244,12 @@ hidden by TCP.
...
 A small allowance for errors in clocks and variations in measurements
 is advisable.  However, any allowance also increases the opportunity
-for replay.  In this case, it is better to reject early data than to
-risk greater exposure to replay attacks.
+for replay.  In this case, it is better to reject early data and fall back
+to a full 1-RTT handshake than to risk greater exposure to replay attacks.
+In common network topologies for browser clients, small allowances on the
+order of ten seconds are reasonable.  Clock skew distributions are not
+symmetric, so the optimal tradeoff may involve an asymmetric replay window
+as well.
```
martinthomson: s/as well//

PR #568, opened 2016-07-29
draft-ietf-tls-cached-info is now RFC 7924

xiaoyinl:
draft-ietf-tls-cached-info is now [RFC 7924](https://tools.ietf.org/html/rfc7924). Update the reference.


xiaoyinl:
Failed due to `404 Not Found while fetching http://unicorn-wg.github.io/idrefs/bibxml/reference.RFC.7924.xml`

ekr:
Merged by hand

PR #569, opened 2016-07-29
Remove references to the disallowed status_request

FiloSottile:


FiloSottile:
I assumed the behavior from the "TLS 1.3: No" entry in the extensions table. However this means that browsers will have to send both status_request and status_request_v2 in all CH for a while. Do we want that?

draft-ietf-tls-tls13.md
```diff
@@ -2194,9 +2194,8 @@ server sends an empty extension to indicate negotiation of this
 extension and the OCSP information is carried in a CertificateStatus
 message. In TLS 1.3, the server's OCSP information is
```
kaduk: Might be worth noting that the v1 status_request is forbidden inline here, e.g., "In TLS 1.3, the RFC 6066 status_request extension is disallowed, so only the RFC 6961 status_request_v2 may be used, and the server's OCSP information is carried in an extension in EncryptedExtensions".

kaduk:
I think we want browsers sending both for a while, for maximal compatibility, yes.

I a little bit wonder if it's worth having text about "when TLS 1.3 is negotiated", for servers that support both 1.2 and 1.3, but I think we generally don't do that and it's generally the right thing to (not) do.

ekr:
This was an error on my part. Both are still allowed.

PR #579, opened 2016-08-15
status_request is Encrypted

martinthomson:
I think that this was just an error.  `status_request` is mentioned elsewhere in the text and seemingly permitted.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2809,7 +2809,7 @@ ticket no more than once, with more recent tickets being used
 first.
 For handshakes that do not use a resumption_psk, the
 resumption_context is a string of [[TBD: This is not safe now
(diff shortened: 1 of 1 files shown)


kaduk:
Somehow I had the sense that you can send status_request if you are willing to negotiate a lower version, but if 1.3 is negotiated status_request cannot be returned by the server (only status_request_v2).

But, if you encrypt it in your 1.3 ClientHello, the 1.2-only server you might be talking to wouldn't be able to handle it ... so, contingent on the above, this doesn't seem to make sense.

davidben:
Encrypted means encrypted by the server, not the client. There are no client-encrypted extensions.

No one currently implements status_request_v2 that I know of, and I don't see much reason to tie v2 and TLS 1.3 together, so +1 from me on this change.

kaduk:
Oops, sorry for being confused.

PR #426, opened 2016-02-26
KeyUpdate message: add receive_generation field

JudsonWilson:
A contribution from:

Judson Wilson, Henry Corrigan-Gibbs, Riad S. Wahby,
Keith Winstein, Philip Levis, and Dan Boneh

Stanford University

PR #580, opened 2016-08-16
KeyUpdate message: add receive_generation field

keithw:
Rebased version of #426


ekr:
@seanturner can the chairs bring this discussion to a close

PR #582, opened 2016-08-18
Update reference

xiaoyinl:
draft-ietf-tls-negotiated-ff-dhe is now RFC 7919.
draft-mattsson-tls-ecdhe-psk-aead is now draft-ietf-tls-ecdhe-psk-aead.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -47,7 +47,6 @@ normative:
   RFC7251:
   RFC7748:
   RFC7905:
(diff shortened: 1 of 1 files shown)


martinthomson:
Do we even need a reference to draft-ietf-tls-ecdhe-psk-aead with the new negotiation syntax?

xiaoyinl:
@martinthomson I removed it from the reference section. It was used in previous versions. But it is no longer referenced to by any text in the spec since version 15.

Issue #560, opened 2016-07-25
Cookie-only HelloRetryRequest doesn't quite work right


davidben:
This was mentioned at the hackathon, but I forgot to file a bug. The server may send a HelloRetryRequest if it needs a key_share AND/OR if it wants a cookie to prove reachability. If only the cookie, this part of the spec can't be followed:

> Upon receipt of a HelloRetryRequest, the client MUST first verify that the selected_group field corresponds to a group which was provided in the “supported_groups” extension in the original ClientHello. It MUST then verify that the selected_group field does not correspond to a group which was provided in the “key_share” extension in the original ClientHello.

Relatedly, if the server selected a pure PSK cipher but still needed to assert a cookie, it can't send a HelloRetryRequest without requesting a key share.

Simplest fix would probably be to move key_share to the extensions field, since it otherwise is an extension anyway. Though the `select (role)` needs now three cases, so it's a little screwy. Then the text can say something like "if there is a key_share extension present, the client MUST first verify, etc.". It may or may not also be worth saying something like "servers MUST NOT send HelloRetryRequest without any extensions".

ekr:
Yeah, I noticed this the other day too. 

Not sure about your fix. Why not just say that if you sent a share in the group you take the cookie and move on

davidben:
How would you handle pure PSK ciphers where you don't pick a group at all?

ekr:
Yeah, fair point. I'll have to think about that.

hannestschofenig:
I would change the structure of the HelloRetryRequest message 

FROM:

struct {
       ProtocolVersion server_version;
       CipherSuite cipher_suite;
       NamedGroup selected_group;
       Extension extensions<0..2^16-1>;
   } HelloRetryRequest;

TO:

struct {
       ProtocolVersion server_version;
       CipherSuite cipher_suite;
       Extension extensions<0..2^16-1>;
   } HelloRetryRequest;

and put the selected_group in the extension part. Then, if the client only used a PSK cipher without any ECDHE then it will not provide a selected_group extension in the ClientHello and the server does not need to include it in the HelloRetryRequest either. 

On the other hand, you are not likely going to send a HelloRetryRequest when you are using a pure PSK-based ciphersuite since the error condition of the server not finding an acceptable set of algorithms and groups that are mutually supported is not going to occur. The use of the  HelloRetryRequest for DoS protection is only mentioned as a side remark and I believe it was introduced for unreliable transports rather than for TLS over TCP.

[...]

PR #583, opened 2016-08-18
Move HelloRetryRequest.selected_group to an extension.

davidben:
Gave writing this up a go. It might need further refining though. (Not sure if the lack of definition for random ServerHello extensions is sufficient to imply, e.g., thou shalt not send sigalgs in HRR. Although arguably this was a problem in the original version too.)

---

This avoids the problems around the interaction with other
HelloRetryRequest reasons (see issue #560). Instead, all problems are
extensions and the rules are simply:

- A HelloRetryRequest must contain at least one extension.
- If a selected_group extension is present, it must not be a no-op.
- Future HRR xtns can define their own rules and modifications to the
  ClientHello.

Also clarify the intent that the old ClientHello be computable from the
new one.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2075,11 +2075,11 @@ The "extension_data" field of this extension contains a
        } PskIdentity;
 
        struct {
-           select (Role) {
-               case client:
+           select (Handshake.msg_type) {
+               case client_hello:
                    psk_identity identities<2..2^16-1>;
 
-               case server:
+               case server_hello:
                    uint16 selected_identity;
            }
        } PreSharedKeyExtension;
@@ -2158,11 +2158,11 @@ The "extension_data" field of this extension contains an
 %%% Key Exchange Messages
 
        struct {
-           select (Role) {
-               case client:
+           select (Handshake.msg_type) {
+               case client_hello:
                    uint32 obfuscated_ticket_age;
 
-               case server:
+               case server_hello:
                   struct {};
            }
        } EarlyDataIndication;
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1558,9 +1558,10 @@ When this message will be sent:
 > Servers send this message in response to a ClientHello
 message if they were able to find an acceptable set of algorithms and
 groups that are mutually supported, but
-the client's KeyShare did not contain an acceptable
-offer. If it cannot find such a match, it will respond with a
-fatal "handshake_failure" alert.
+the client's ClientHello did not contain sufficient information to
+proceed with the handshake.
+If a server cannot successfully select algorithms, it will respond with a
+fatal "handshake_failure" alert instead.
 
 Structure of this message:
 
@@ -1568,13 +1569,9 @@ Structure of this message:
 
        struct {
            ProtocolVersion server_version;
-           NamedGroup selected_group;
-           Extension extensions<0..2^16-1>;
+           Extension extensions<2..2^16-1>;
        } HelloRetryRequest;
 
-selected_group
-: The mutually supported group the server intends to negotiate and
-  is requesting a retried ClientHello/KeyShare for.
 {:br }
 
 The version and extensions fields have the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1953,11 +1947,14 @@ The "extension_data" field of this extension contains a
 %%% Key Exchange Messages
 
        struct {
-           select (role) {
-               case client:
+           select (Handshake.msg_type) {
```
davidben: If this syntax is acceptable, we can make the other extensions match for consistency. I wasn't sure how to spell it.

draft-ietf-tls-tls13.md
```diff
@@ -1585,28 +1582,25 @@ exist). As with ServerHello, a
...
+block is not empty and otherwise abort the handshake with a fatal
+"decode_error" alert.  Clients SHOULD also abort with "unexpected_message" in
+response to any second HelloRetryRequest which was sent in the same connection
+(i.e., where the ClientHello was itself in response to a HelloRetryRequest).
+
+Otherwise, the client MUST process all extensions in the HelloRetryRequest and
+send a second updated ClientHello. The HelloRetryRequest extensions defined in
+this specification are:
+
+- cookie [{{cookie}}]
```
martinthomson: Nit: Skip the square brackets (that looks like a citation reference), and instead do "(see {{cookie}})".

draft-ietf-tls-tls13.md
```diff
@@ -1585,28 +1582,25 @@ exist). As with ServerHello, a
...
+
+- key_share [{{key-share}}]
```
martinthomson: Nit 2: I think that the markdown processor likes "*" for its lists.  Or at least we should be consistent with other lists in the spec.

draft-ietf-tls-tls13.md
martinthomson: I think that it is good.

draft-ietf-tls-tls13.md
```diff
@@ -1984,11 +1985,25 @@ KeyShareEntry values for groups not listed in the client's
...
+KeyShareEntry values MUST be preserved.
+
 If using (EC)DHE key establishment, servers offer exactly one
-KeyShareEntry. This value MUST correspond to the KeyShareEntry value offered
+KeyShareEntry in the ServerHello. This value MUST correspond to the KeyShareEntry value offered
 by the client that the server has selected for the negotiated key exchange.
 Servers MUST NOT send a KeyShareEntry for any group not
 indicated in the "supported_groups" extension.
+If a HelloRetryRequest was received, the client MUST verify that the
+selected NamedGroup matches that supplied in the selected_group field and MUST
```
martinthomson: Maybe:

> If a HelloRetryRequest included the supported_groups extension, the client MUST verify that the key_share value in the ServerHello is for the same group and abort [...]

Note that HRR might not include supported_groups if the server is only sending a cookie.

martinthomson:
I think that we also need to change the IANA considerations for extensions to include a new column: HRR.  This will indicate if the extension is defined as being valid for inclusion in HelloRetryRequest.

draft-ietf-tls-tls13.md
```diff
@@ -1585,28 +1582,25 @@ exist). As with ServerHello, a
...
+
+- cookie [{{cookie}}]
+
+- key_share [{{key-share}}]
+
+Note that HelloRetryRequest extensions are defined such that the original
+ClientHello may be computed from the new one, given minimal state about which
+HelloRetryRequest extensions were sent. For example, the key_share extension
+causes the new KeyShareEntry to be appended to the client_shares field, rather
+than replacing it.
```
martinthomson: I think that it would help to point out here that extensions in HelloRetryRequest can be sent by the server even if the client did not include them in their ClientHello.  Or maybe we just need a special rule for `cookie`.

draft-ietf-tls-tls13.md
davidben: It seems to like either. We're gloriously inconsistent, but hyphens are winning.
```
$ git grep '^- ' | wc -l
     175 # Most of this is the changes list
$ git grep '^* ' | wc -l
      62 # Most of this is the contributors list
```
 (Happy to use either if you particularly care though. I don't have views.)

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md
davidben: Added special rule for `cookie` earlier where it says unexpected extensions are illegal, just to leave the semantics the same for now.

[I'm not a huge fan of `cookie` being MTI like this, to be honest. Making it negotiated doesn't cost much, though even that's not quite right. It's a function of whether the transport provides what you need here or not. TLS doesn't need it. DTLS does. I've heard competing ideas for whether QUIC should use this or do something at its lower layer, but either way it'd be "all clients MUST support" or "no need to have cookie at all". Nice thing about lower layers is you don't need to worry about ClientHello exceeding a packet, as you do in DTLS.

Anyway, we can sort all that out separately.]

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md
davidben: Switched the other two instances for consistency.

draft-ietf-tls-tls13.md
```diff
@@ -1953,11 +1947,14 @@ The "extension_data" field of this extension contains a
...
                    KeyShareEntry client_shares<0..2^16-1>;
 
-               case server:
+               case hello_retry_request:
+                   NamedGroup selected_group;
+
+               case server_hello:
                    KeyShareEntry server_share;
            }
        } KeyShare;
```
davegarrett: I really don't like using the key_share extension for this; it's not actually sharing a key here. This really should be done with the supported_groups extension instead.

davidben: Honestly, the key_share / supported_groups split is kinda silly. If we didn't have the TLS 1.2 constraint, I imagine we would have made it one extension and just left the missing key shares blank (or maybe did something tighter).

I used key_share because it makes it nicely symmetric with the ServerHello one (if the server sends key_share in either message, it means the server wants to do (EC)DHE for *this* connection). Whereas the server might send supported_groups in either PSK-only or (EC)DHE-ful handshakes.

This also lets the description for KeyShare processing stay in one place. The full story is told in the description of the key_share extension.

davegarrett: There's no disputing that it's a compatibility kludge to use two extensions for negotiation, but adding this kludge onto that kludge feels iffy. Using key_share to exchange a different structure from normal does not feel like the way to go here. I would much prefer less special-casing.

Using supported_groups makes more sense because the point of an HRR is to inform the client that the server doesn't support anything in the client's key_share, but does support something in its supported_groups. Just sending its full supported_groups extension could also work, if the client wishes to pick itself.

davidben: Ugh, definitely not. supported_groups vs key_share is just syntax, but it must be a single group, not a list. The properties of the negotiation are complex enough as it is with the half-client half-server pick. We should at least ensure the server gets the final say.

For BoringSSL we've intentionally had the server just pick a group without paying attention to the key share list. Then we'll decide whether we HelloRetryRequest based on that decision. It's possible, in the future, there will be more variation in curve preferences but, right now, the only sensible curves are X25519 and P-256 anyway.

We already have extensions which are different between ServerHello and HelloRetryRequest. In fact, we have no extensions which are the same between the two. Every other extension is sent in at most one of the two, so implementations must already treat them different. Likewise, we already have extensions which are different between ClientHello and ServerHello. Also just fine.

I think the handful of cases where the same structure was used for client and server despite the two having very different roles (ALPN) were mistakes, and I'm glad we're not repeating them for 1.3's other extensions. No one is going to use the same code for client and server extension parsing because what you do with the fields is very different. It's purely an artificial constraint.

martinthomson: > For BoringSSL we've intentionally had the server just pick a group without paying attention to the key share list. Then we'll decide whether we HelloRetryRequest based on that decision. It's possible, in the future, there will be more variation in curve preferences but, right now, the only sensible curves are X25519 and P-256 anyway.

We've done the same in NSS, though in future we will pick the curve and permit "equal" curves to be picked if the first choice doesn't have a key share.

I generally agree on the position that the encoding doesn't need to be the same for client and server (unless it needs to be).  Extensions aren't fixed structures, they are containers for a protocol.

ekr: I agree with David that sending a full list from the server would be bad.

ekr:
Merged by hand

PR #585, opened 2016-08-21
Downgrade a SHA-1 ordering requirement from MUST to SHOULD.

davidben:
Alas, it turns out that not only are there plenty of servers incapable
of signing anything but SHA-1, but there are bugs in signature algorithm
processing. Specifically, older versions of NSS have the following bug:

   https://bugzilla.mozilla.org/show_bug.cgi?id=1119983

Affected versions will:

a) Refuse to sign anything but SHA-1. (Type mismatch between
   HashAlgorithm and internal OID identifier.)

b) Require that the SHA-1-based algorithm appear in the first N sigalg
   entries, where N is the number of sigalg entries that begin with a
   HashAlgorithm that NSS knows how to parse. (Mistake in appending to
   an array.)

The upshot is, with SHA-1 at the end, advertising new algorithms like
rsa_pss_sha256 or ed25519 will break those servers. This bug does not
appear very common (I found 35 hosts in a probe of the Alexa top 1m
list), but it is difficult to predict these things ahead of time. At
least one affected piece of software targets enterprise customers (it
bundles a copy of NSS), where outdated software is prevalent and
visibility nil.

To avoid tying our hands should we need ugly workarounds, downgrade MUST
to SHOULD. RSASSA-PKCS1-v1_5 is forbidden in TLS 1.3 anyway, so while
advertising rsa_pkcs1_* at the front would be poor, it shouldn't have
serious adverse effects.

(Note that version-intolerance is many orders magnitude higher, so this
will not save us from the fallback on its own. Nonetheless, even with a
fallback, it is simpler if the signature algorithms list could
eventually be kept constant between TLS 1.2 and TLS 1.3.)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1817,8 +1817,8 @@ EdDSA algorithms
 {:br }
 
 rsa_pkcs1_sha1, dsa_sha1, and ecdsa_sha1 SHOULD NOT be offered. Clients
-offering these values for backwards compatibility MUST list them as the lowest
-priority (listed after all other algorithms in the
+offering these values for backwards compatibility SHOULD list them as the
+lowest priority (listed after all other algorithms in the
 supported_signature_algorithms vector). TLS 1.3 servers MUST NOT offer a SHA-1
 signed certificate unless no valid certificate chain can be produced without it
 (see {{server-certificate-selection}}).


briansmith:
I say we just let those servers be broken and avoid this change.

davidben:
Yeah, I was intending to see if we could break them first, but I didn't want to find out too late that we can't. I'm mostly worried about some over-eager server deciding actually enforce this MUST on clients.

If the WG is willing to consider the RSA-PSS numbers fixed, we can possibly try to ship them in Chrome quickly with an emergency field trial off switch. But even then I won't get information back on problematic enterprise stuff for a while. They by and large only run the stable channel.

davegarrett:
On Sunday, August 21, 2016 12:53:32 pm David Benjamin wrote:
> Alas, it turns out that not only are there plenty of servers incapable
> of signing anything but SHA-1, but there are bugs in signature algorithm
> processing. Specifically, older versions of NSS have the following bug:
> 
>    https://bugzilla.mozilla.org/show_bug.cgi?id=1119983
> 
> Affected versions will:
> 
> a) Refuse to sign anything but SHA-1. (Type mismatch between
>    HashAlgorithm and internal OID identifier.)
> 
> b) Require that the SHA-1-based algorithm appear in the first N sigalg
>    entries, where N is the number of sigalg entries that begin with a
>    HashAlgorithm that NSS knows how to parse. (Mistake in appending to
>    an array.)

Even allowing SHA-1 in the list at all was a compromise for backwards compatibility with old broken junk. Attempting to bend over backwards to work around something this broken is a bad idea. At some point, a bug has to break things.

> The upshot is, with SHA-1 at the end, advertising new algorithms like
> rsa_pss_sha256 or ed25519 will break those servers. This bug does not
> appear very common (I found 35 hosts in a probe of the Alexa top 1m
> list), but it is difficult to predict these things ahead of time. At
> least one affected piece of software targets enterprise customers (it
> bundles a copy of NSS), where outdated software is prevalent and
> visibility nil.

That's extraordinarily rare. If they're broken, then they break. It's not the job of a security protocol spec to account for rare brokeness not being ever patched due to incompetence. If we have to kludge stuff, let's please at least focus on the things affecting measurably large numbers of people. :(

davidben:
(To be clear, that original compromise is completely forced. I had hopes of getting SHA-1 SKX signatures out of TLS 1.2 at one point---you can't implement TLS 1.2 without SHA-2! But it turns out SHA-1-only servers are more common than TLS 1.0 and TLS 1.1 combined. I'm amazed we managed that one, but so it goes. Will revisit it in a couple years.)

It does seem rare and indeed I had no intention of invoking the SHOULD. But I don't have much confidence in how accurate these top site probes are. In my experience, even client-side metrics have problems. These bugs tend to be concentrated at enterprise deployments which are invisible to top site lists and underrepresented in client-side metrics.

I may well be being unnecessarily paranoid here. Perhaps it's a low enough zeroth-order estimate the spec can just take that risk. I figured there was no need in doing so, especially since placing `rsa_pkcs1_*` at the front didn't actually cost anything. *shrug*

(The risk here is the probe is totally inaccurate and we're left in a bind. I don't want to put `rsa_pkcs1_*` in front, but I would prefer that to fallbacks.)

davegarrett:
On Sunday, August 21, 2016 11:12:33 pm David Benjamin wrote:
> I may well be being unnecessarily paranoid here. Perhaps it's a low enough zeroth-order estimate the spec can just take that risk. I figured there was no need in doing so, especially since placing `rsa_pkcs1_*` at the front didn't actually cost anything. *shrug*

Huh? A client placing it at the front will negotiate it when connecting to a TLS 1.2 server that supports it. In a vacuum where a client is crafting a specific request for a specific server, yeah, this might be safe in some instance, but that's not what the spec can assume. If the value is offered with a high priority, it will be used when it shouldn't.

davidben:
(To clarify in case I was unclear, the workaround, should we need it, would be to put all rsa_pkcs_* values at the front, not just rsa_pkcs1_sha1, so preferences relative to that will still be correct.)

In TLS 1.3, all those entries are no-ops, so it doesn't matter what their order.

For TLS 1.2, it's not quite a no-op because we decided to backport RSA-PSS, but nearly all cases where it matters will negotiated 1.3 anyway. There is no chance we will be able to get rid of TLS 1.2 + PKCS#1 before we get rid of TLS 1.2. The PSS switch for 1.3 is good, but I feel the backport was basically a no-op. (Note if PKCS#1 is totally bust, whether the server preferentially signs PSS or not won't matter if the client still accepts PKCS#1.)

Or perhaps am I misunderstanding the concern or missing something?

There's also ordering between, say, RSA and ECDSA. But poking at SSL Labs, it seems every major browser already orders RSA > ECDSA in the sigalgs list anyway. (Amusingly, Chrome is the only browser which actually puts SHA-1 things at the end! BoringSSL inherited OpenSSL's behavior of sorting by hash first. Everyone else does all the RSAs and then all the ECDSAs. Except for Edge which does that but punts all the SHA512s to the end.)

davidben:
Current plan is to just ship the RSA-PSS signature algorithms in Chrome early so we can clear through this mess before it's time to deploy TLS 1.3. Hopefully that'll work out and we won't need anything messy! :-) Closing.

PR #590, opened 2016-08-25
Fixing some nits in s2.

seanturner:
Some suggested edits for s2.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -794,9 +794,9 @@ EncryptedExtensions.
   determine the cryptographic parameters. [{{encrypted-extensions}}]
 
 CertificateRequest.
-: if certificate-based client authentication is requested, the
+: if certificate-based client authentication is desired, the
   desired parameters for that certificate. This message is
-  omitted if client authentication is not requested. [{{certificate-request}}]
+  omitted if client authentication is not desired. [{{certificate-request}}]
 
 Finally, the client and server exchange Authentication messages. TLS
 uses the same set of messages every time that authentication is needed.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -752,8 +752,8 @@ Auth | {CertificateVerify*}
 ~~~
 {: #tls-full title="Message flow for full TLS Handshake"}
 
-The handshake can be thought of as having three phases, indicated
-in the diagram above:
+The handshake can be thought of as having three phases (indicated
+in the diagram above):
 
 - Key Exchange: Establish shared keying material and select the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2335,7 +2335,7 @@ extensions
 : A list of extensions.
 {:br }
 
-###  Certificate Request
+###  Certificate Request {#certificate-request}
```
davegarrett: You don't need to add this anchor here; every section gets an autogenerated anchor named in this exact form, so this is a no-op. Manually specifying these is only useful if you've got a long section name or otherwise wish to use an abbreviation, instead.

draft-ietf-tls-tls13.md
```diff
@@ -709,12 +709,12 @@ use the established keys to protect application layer traffic.
...
 
-- Diffie-Hellman (of both the finite field and elliptic curve
+- Diffie-Hellman (both the finite field and elliptic curve
   varieties).
 
-- A pre-shared symmetric key (PSK)
+- A pre-shared symmetric key (PSK).
 
-- A combination of a symmetric key and Diffie-Hellman
+- A combination of a symmetric key and Diffie-Hellman.
```
davegarrett: nit of a nit: none of these bullet points are sentences, thus none really should have periods. I'd suggest not adding the latter two and dropping the first instead.

draft-ietf-tls-tls13.md
seanturner: Whoops my bad.

draft-ietf-tls-tls13.md
```diff
@@ -753,14 +753,14 @@ Auth | {CertificateVerify*}
 {: #tls-full title="Message flow for full TLS Handshake"}
 
 The handshake can be thought of as having three phases, indicated
-in the diagram above.
+in the diagram above:
```
kaduk: I might put "indicated in the diagram above" in parentheses, as otherwise the colon right after "above" could be a touch jarring.

draft-ietf-tls-tls13.md
```diff
@@ -775,28 +775,28 @@ and potentially some other extensions.
...
 EncryptedExtensions.
-: responses to any extensions which are not required in order to
+: responses to any extensions that are not required to
   determine the cryptographic parameters. [{{encrypted-extensions}}]
 
 CertificateRequest.
-: if certificate-based client authentication is desired, the
-  desired parameters for that certificate. This message will
-  be omitted if client authentication is not desired.
+: if certificate-based client authentication is requested, the
```
ekr: I think "desired" is the right word here, because it's this message which tells you to do it.

PR #594, opened 2016-08-25
Nits in s2.2.

seanturner:


draft-ietf-tls-tls13.md
```diff
@@ -931,8 +931,8 @@ Subsequent Handshake:
 {: #tls-resumption-psk title="Message flow for resumption and PSK"}
 
 As the server is authenticating via a PSK, it does not send a
-Certificate or a CertificateVerify. When a client offers resumption
-via PSK it SHOULD also supply a "key_share" extension to the server as well
+Certificate or a CertificateVerify message. When a client offers resumption
+via PSK, it SHOULD also supply a "key_share" extension to the server as well
```
kaduk: I think this sentence is fine with or without this comma.  (I might add one before "to allow", though.)  But I have no real opinion and leave it to the editor's discretion.

draft-ietf-tls-tls13.md
```diff
@@ -877,9 +877,9 @@ described in the following sections.
 
 Although TLS PSKs can be established out of band,
 PSKs can also be established in a previous session and
-then reused ("session resumption"). Once a handshake has completed, the server can
-send the client a PSK identity which corresponds to a key derived from
-the initial handshake (See {{NewSessionTicket}}). The client
+then reused (i.e., "session resumption"). Once a handshake has completed, the server can
```
ekr: This i.e., is wrong, we are defining the term here.

draft-ietf-tls-tls13.md
ekr: I think the comma improves it.

PR #596, opened 2016-08-25
Fis PSK-related types

Lekensteyn:
The PskKeMode and PskAuthMode types were not defined, substitute them by
the defined enums. Use non-plural enum names for consistency with other
enumeration types.

Remove the old opaque psk_identity type in favor of PskIdentity which
was otherwise unused. Increase minimum length from 2 to 4 to account for
the PskIdentity.identity field length.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2043,21 +2043,19 @@ The "extension_data" field of this extension contains a
 
 %%% Key Exchange Messages
 
-       enum { psk_ke(0), psk_dhe_ke(1), (255) } PskKeyExchangeModes;
-       enum { psk_auth(0), psk_sign_auth(1), (255) } PskAuthenticationModes;
-
-       opaque psk_identity<0..2^16-1>;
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2043,21 +2043,19 @@ The "extension_data" field of this extension contains a
...
+           PskKeyExchangeMode ke_modes<1..255>;
+           PskAuthenticationMode auth_modes<1..255>;
            opaque identity<0..2^16-1>;
        } PskIdentity;
 
        struct {
            select (Role) {
                case client:
-                   psk_identity identities<2..2^16-1>;
+                   PskIdentity identities<4..2^16-1>;
```
kaduk: Hmm, looks like the minimum length is 6, though (two nonempty 1-byte-length fields at 2 bytes each, plus two bytes of length for the optional identity field).

Lekensteyn: Correct, I missed the minimum length requirement. Will amend patch.

PR #604, opened 2016-08-29
Change "supported_groups" to "supported_kems" - supported key exchang…

zhenfeizhang:
I would like to suggest that we change the terminology "NamedGroup" to "KeyExchangeMethod". 

In [1], it is suggested that we redefine the syntax, which leads to the separation of public key crypto 
and symmetric crypto during a handshake. Because of this separation, new terminology was defined
for key exchange algorithms and authentication algorithms for public key crypto in the key exchange
extension. "NamedGroup" was used to refer the underlying key exchange parameters, which comes 
from the "Supported Elliptic Curves" in previous versions. 

The use of "NamedGroup" implicitly requests the key exchange algorithm to be Deffie-Hellman type. 
While it is safe for now, it would be nice to have some crypto agility, and future proof. It would make 
the transition to other key exchange primitives (such as lattice based key exchange) or methods 
(such as key encapsulation mechanism) easier in the future, if we do not restrict the key exchange
by certain "Group".

Knowing that NIST has planned to standardize quantum-safe cryptography within 7 years of time
(which can and should be accelerated), and those algorithms cannot be described in terms of "group",
the current terminology will due for a redesign by then. So I would suggest to change the "NamedGroup"
now rather than latter.

Overall, this will have the following impact 

1. HelloRetryRequest

Change HelloRetryRequest structure to

struct {
       ProtocolVersion server_version;
       KeyExchangeMethod selected_kem;
       Extension extensions<0..2^16-1>;
   } HelloRetryRequest;

2. Negotiated Groups

Throughout, change "supported_groups" to
"supported_kems"; change "NamedGroupList" to
"KeyExchangeMethodList"; change "named_group_list" to
"kem_list"; change NamedGroup to KeyExchangeMethod

3. Key Share:

Change KeyShareEntry structure to

 struct {
       KeyExchangeMethod kem;
       opaque key_exchange<1..2^16-1>;
   } KeyShareEntry;

[1] https://github.com/ekr/tls13-spec/blob/15126cf5a08c445aeed97c0c25c4f10c2c1b8f26/draft-ietf-tls-tls13.md

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -781,7 +781,8 @@ parameters. [{{server-hello}}]. The combination of the ClientHello
 and the ServerHello determines the shared keys. If (EC)DHE
 key establishment is in use, then the ServerHello
 will contain a "key_share" extension with the server's ephemeral
-Diffie-Hellman share which MUST be in the same group as one of the
+Diffie-Hellman share which MUST be using a same key exchange method (in the
+same group) as one of the
 client's shares. If PSK key establishment is
 in use, then the ServerHello will contain a "pre_shared_key"
 extension indicating which of the client's offered PSKs was selected.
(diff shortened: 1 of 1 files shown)


ekr:
Closed per chair instruction

PR #611, opened 2016-08-31
Dual keying ladders, plus limit number of KeyUpdates.

ekr:
The idea here is that you can be silently receiving, receive a pile
of KeyUpdates, and then respond with just one. Because there are
dual ladders, this doesn't cause PFS problem.

@davidben PTAL

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2956,7 +2956,7 @@ the responses).
 
 
 ### Key and IV Update {#key-update}
-     enum { update_not_requested(0), update_requested(2), (255)
+     enum { update_not_requested(0), update_requested(1), (255)
      } KeyUpdateRequest;
 
      struct {
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3625,7 +3625,9 @@ operation.
                  |
                  v
               Handshake
-               Secret -----> Derive-Secret(., "client handshake traffic secret",
+               Secret
+                 |
+                 +---------> Derive-Secret(., "client handshake traffic secret",
                  |                         ClientHello...ServerHello)
                  |                         = client_handshake_traffic_secret
                  |
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2957,7 +2957,16 @@ the responses).
 
 ### Key and IV Update {#key-update}
 
-     struct {} KeyUpdate;
+     enum { true(1), false(2) (255) } Boolean;
+
+     struct {
+         Boolean requestUpdate;
+     } KeyUpdate;
+
+requestUpdate
+: Indicates that the recipient of the KeyUpdate should respond with its
+own KeyUpdate.
+{:br }
 
 The KeyUpdate handshake message is used to indicate that the sender is
 updating its sending cryptographic keys. This message can be sent by
@@ -2970,31 +2979,25 @@ next generation of keys, computed as described in
 {{updating-traffic-keys}}. Upon receiving a KeyUpdate, the receiver
 MUST update its receiving keys.
 
-Each implementation MUST maintain a min_send_generation counter that
-is initialized to 0 at the time the traffic keys are installed.
-Upon receiving a KeyUpdate, if the receiver has sent any data records since receiving the
-last KeyUpdate it MUST also increment its min_send_generation value
-by 1. Otherwise, the min_send_generation MUST remain unchanged. Prior
-to sending a record, if min_send_generation is greater than the
-current sending generation, the sender MUST first send a KeyUpdate.
-
-This mechanism allows either side to force an update to the
+If the requestUpdate field is set to "true" then the receiver MUST
+send a KeyUpdate of its own with requestUpdate set to "false" prior
+to sending its next application data record. This mechanism allows either side to force an update to the
 entire connection, but causes an implementation which
 receives multiple KeyUpdates while it is silent to respond with
 a single update. Note that implementations may receive an arbitrary
 number of messages between sending a KeyUpdate and receiving the
 peer's KeyUpdate because those messages may already be in flight.
-If implementations independently send their own
-KeyUpdates and they cross in flight, then both sides simply
-notes that it has already updated without sending an additional
-update.
-
-Note that the side which sends its KeyUpdate first needs to retain
+The side which sends its KeyUpdate first needs to retain
 its receive traffic keys (though not the traffic secret) for the previous
 generation of keys until it receives the KeyUpdate from the other
-side. Because the send and receive keys are derived from independent
+side. Because send and receive keys are derived from independent
 traffic secrets, retaining the receive traffic secret does not threaten
 the forward secrecy of data sent before the sender changed keys.
+
+If implementations independently send their own KeyUpdates with
+requestUpdate set to true, and they cross in flight, then each side
+will also send a response, with the result that each side increments
+by two generations.
 
 Both sender and receiver MUST encrypt their KeyUpdate
 messages with the old keys. Additionally, both sides MUST enforce that
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2494,7 +2494,7 @@ for each scenario:
 
 | Mode | Handshake Context | Base Key |
 |------|-------------------|----------|
-| 0-RTT | ClientHello | early_traffic_secret|
+| 0-RTT | ClientHello | client_early_traffic_secret|
 | 1-RTT (Server) | ClientHello ... later of EncryptedExtensions/CertificateRequest | [sender]_handshake_traffic_secret |
 | 1-RTT (Client) | ClientHello ... ServerFinished     | [sender]_handshake_traffic_secret |
 | Post-Handshake | ClientHello ... ClientFinished + CertificateRequest | [sender]_traffic_secret_N |
@@ -3605,7 +3605,7 @@ In this diagram, the following formatting conventions apply:
   from the left.
 - Derive-Secret's Secret argument is indicated by the arrow coming in
   from the left. For instance, the Early Secret is the Secret for
-  generating the early_traffic_secret.
+  generating the client_early_traffic_secret.
 
 Note that the 0-RTT Finished message is not included in the Derive-Secret
 operation.
@@ -3617,9 +3617,9 @@ operation.
    PSK ->  HKDF-Extract
                  |
                  v
-           Early Secret ---> Derive-Secret(., "early traffic secret",
+           Early Secret ---> Derive-Secret(., "client early traffic secret",
                  |                         ClientHello)
-                 |                         = early_traffic_secret
+                 |                         = client_early_traffic_secret
                  v
 (EC)DHE -> HKDF-Extract
                  |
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2969,24 +2968,33 @@ MUST generate a fatal "unexpected_message" alert.  After sending a
...
 next generation of keys, computed as described in
 {{updating-traffic-keys}}. Upon receiving a KeyUpdate, the receiver
-MUST update their receiving keys and if they have not already updated
-their sending state up to or past the then current receiving
-generation MUST send their own KeyUpdate prior to sending any other
-messages.  This mechanism allows either side to force an update to the
-entire connection. Note that implementations may receive an arbitrary
+MUST update its receiving keys.
+
+Each implementation MUST receive a min_send_generation counter that
```
kaduk: s/receive/maintain/?

draft-ietf-tls-tls13.md
```diff
@@ -2969,24 +2968,33 @@ MUST generate a fatal "unexpected_message" alert.  After sending a
...
+is initialized to 0 at the time the traffic keys are installed.
+Upon receiving a KeyUpdate, if the receiver has sent any data records since receiving the
```
davidben: Should it be since sending the last KeyUpdate?

draft-ietf-tls-tls13.md
```diff
@@ -2969,24 +2968,33 @@ MUST generate a fatal "unexpected_message" alert.  After sending a
...
+last KeyUpdate it MUST also increment its min_send_generation value
+by 1. Otherwise, the min_send_generation MUST remain unchanged. Prior
+to sending a record, if min_send_generation is greater than the
+current sending generation, the sender MUST first send a KeyUpdate.
```
davidben: I think this is the same as a flag here since it's always current sending generation or current sending generation + 1 and never compared against anything.

Bigger issue is I don't think this prevents ping-pong. If both sides are continually sending data, then once one side decides to KeyUpdate, both sides will just keep doing it forever. Or perhaps I'm misunderstanding?

draft-ietf-tls-tls13.md
```diff
@@ -2969,24 +2968,33 @@ MUST generate a fatal "unexpected_message" alert.  After sending a
...
-
-Note that if implementations independently send their own
-KeyUpdates and they cross in flight, this only results in an
-update of one generation; when each side receives the other
-side's update it just updates its receive keys and notes that
-the generations match and thus no send update is needed.
+If implementations independently send their own
+KeyUpdates and they cross in flight, then both sides simply
+notes that it has already updated without sending an additional
+update.
```
davidben: How do you tell? Say we both do KeyUpdate immediately followed by data.

draft-ietf-tls-tls13.md
```diff
@@ -3617,18 +3625,27 @@ operation.
                  |
```
davidben: Should we rename early_traffic_secret => client_early_traffic_secret?

draft-ietf-tls-tls13.md
```diff
@@ -3617,18 +3625,27 @@ operation.
...
                  v
               Handshake
-               Secret -----> Derive-Secret(., "handshake traffic secret",
+               Secret -----> Derive-Secret(., "client handshake traffic secret",
```
davidben: Nit: to be less messy and consistent with Master Secret, how about:

```
Handshake Secret
       |
       +--------->  Derive-Secret(...
       |
      ...

File: draft-ietf-tls-tls13.md (hunk 5 above)
Comment (ekr) [inline draft-ietf-tls-tls13.md]: Oh, sure

File: draft-ietf-tls-tls13.md (hunk 3 above)
Comment (davidben) [inline draft-ietf-tls-tls13.md]: After chatting out-of-band, I misunderstood this scheme and it doesn't cause ping-pong. But instead it doesn't let you ask the peer to KeyUpdate if, say, the peer sent 1,000 KeyUpdates in a row two hours ago because they thought it was funny. min_send_generation will be far behind send_generation.

Comment (kaduk) [inline draft-ietf-tls-tls13.md]: Wait, so is it or is it not always [current sending generation] or [current sending generation + 1]?  (I read that it is.)
Which would make it a flag "rekey before sending", and receiving a KeyUpdate sets the flag; the flag gets cleared on rekey.  Which does seem susceptible to ping-pong, so I guess I would benefit from the same clarifications you got.

Comment (davidben) [inline draft-ietf-tls-tls13.md]: I believe the intended interpretation was that min_send_generation *only* gets bumped on incoming KeyUpdate and not when you initiate KeyUpdate, so it's not quite a flag. If you initiate KeyUpdate, then min_send = send - 1 and the peer's KeyUpdate will catch you up.

The failure mode is then the inverse because sending 1,000 KeyUpdates for fun will get min_send too far behind send to catch up easily, especially with coallescing.

**Comment by ekr** [comment]
DavidBen and I talked offline and concluded that a boolean would be better (though still not ideal). This PR embodies that.

@davidben PTAL

File: draft-ietf-tls-tls13.md [hunk 7]
```diff
@@ -2969,24 +2977,27 @@ MUST generate a fatal "unexpected_message" alert.  After sending a
...
 peer's KeyUpdate because those messages may already be in flight.
-
-Note that if implementations independently send their own
-KeyUpdates and they cross in flight, this only results in an
-update of one generation; when each side receives the other
-side's update it just updates its receive keys and notes that
-the generations match and thus no send update is needed.
-
-Note that the side which sends its KeyUpdate first needs to retain
+The side which sends its KeyUpdate first needs to retain
```
kaduk: This is still a bit awkward, since neither side can always be sure which side was "first sender"

draft-ietf-tls-tls13.md
kaduk: Ah, got it; thanks.

draft-ietf-tls-tls13.md
davidben: Do we even need this section with the ladders split? It seems just needless mess to recommend people drop the traffic secret and not the receive traffic keys like that. I think I would drop the sentence beginning with "The side which sends its [...]" and then maybe stick a "However" in front of "Because send and receive [...]" to transition from "Note that implementations may [...]"

draft-ietf-tls-tls13.md
```diff
@@ -2958,7 +2957,16 @@ the responses).
 
 ### Key and IV Update {#key-update}
 
-     struct {} KeyUpdate;
+     enum { true(1), false(2) (255) } Boolean;
```
davidben: Ahahahahaha.

Nit: If we're defining this here and not generically, maybe it should be something like:
```
enum { do_not_request_update(0), request_update(1) } KeyUpdateRequest
```
(It doesn't look like you actually need the `(255)`. It implicitly uses enough space for the maximally-defined value otherwise.)

Other nit, I think usually we write `request_update` rather than `requestUpdate`?

draft-ietf-tls-tls13.md
```diff
@@ -3609,26 +3620,37 @@ operation.
...
                  |                         ClientHello)
-                 |                         = early_traffic_secret
+                 |                         = client_early_traffic_secret
                  v
 (EC)DHE -> HKDF-Extract
                  |
                  v
               Handshake
-               Secret -----> Derive-Secret(., "handshake traffic secret",
+               Secret
```
davidben: Unbelievably nitpicky formatting nit: I think we can fit "Handshake Secret" comfortably on one line.

draft-ietf-tls-tls13.md
```diff
@@ -3699,19 +3721,18 @@ each class of traffic keys:
...
+following table indicates the purpose values for each type of key:
 
 | Key Type         | Purpose            |
 |:-----------------|:-------------------|
-| client_write_key | "client write key" |
-| server_write_key | "server write key" |
-| client_write_iv  | "client write iv"  |
-| server_write_iv  | "server write iv"  |
+| write_key | "write key" |
+| write_iv  | "write iv"  |
```
davidben: Nit: Are the "write" prefixes still meaningful? It seemed they were just needed to clarify "client" and "server".

(Oh hrmf. client_write_traffic_secret_N is a bit of a mouthful, but is it said anywhere "client" and "server" refer to the writer than than the reader?)

draft-ietf-tls-tls13.md
kaduk: You don't need the (255), yes, but it is at least somewhat conventional to have it just to remind readers of the width of the length field.

draft-ietf-tls-tls13.md
ekr: I changed them

draft-ietf-tls-tls13.md
```diff
@@ -2969,24 +2977,24 @@ MUST generate a fatal "unexpected_message" alert.  After sending a
...
 next generation of keys, computed as described in
 {{updating-traffic-keys}}. Upon receiving a KeyUpdate, the receiver
-MUST update their receiving keys and if they have not already updated
-their sending state up to or past the then current receiving
-generation MUST send their own KeyUpdate prior to sending any other
-messages.  This mechanism allows either side to force an update to the
-entire connection. Note that implementations may receive an arbitrary
+MUST update its receiving keys.
+
+If the requestUpdate field is set to "update_requested" then the receiver MUST
```
davidben: A few more requestUpdates to fix.

keithw:
I think we have to oppose a change to KeyUpdate that adds P4 (bounded write obligations) but not P3 (ability to learn that a KeyUpdate was read by other side). These are orthogonal and achievable simultaneously with little disruption.  Here are four options that would each be perfectly fine with us:

1. Forbid implementations from sending an `update_not_requested` KeyUpdate unless it is triggered by an `update_requested` KeyUpdate.

2. add a uint8 field to KeyUpdate representing the sender's current receive generation (lower 8 bits) 

3. keep request_update an 8-bit field (same width as it is already), and define the high bit as representing the update_requested boolean, and the remaining bits represent the lower 7 bits of the sender's current receive generation

4. keep request_update an 8-bit field (same width as it is already), define the high bit as representing the update_requested boolean, and the remaining bits represent a 7-bit nonce that must be echoed in the response. (If multiple update_requested KeyUpdates arrive, any of their nonces can be the one that gets echoed.)

Approach 1 seems like it might be acceptable, even understanding the concerns of the list. I am happy to send sample text if helpful.

ekr:
You should raise these concerns on the mailing list, not here on Github.

With that said, I did not perceive much WG enthusiasm for P3, and there was explicit opposition, which is why I did not include it. I will of course defer to the judgement of the chairs.

keithw:
Will reply on-list.

ekr:
@seanturner Please advise on how you want this resolved.

draft-ietf-tls-tls13.md
```diff
@@ -2957,8 +2956,17 @@ the responses).
 
 
 ### Key and IV Update {#key-update}
+     enum { update_not_requested(0), update_requested(1), (255)
```
davidben: What is the intended behavior if an implementation sees a value other than the ones listed? (I think we should say that's fatal and we need an extension if we ever wanted to make that extensible... hopefully this doesn't happen.)

For completeness, an alternate spelling that saves a byte would be to define two messages, KeyUpdate and KeyUpdateMutual that both trigger a key update but one also implicitly requests. (I don't care between the spellings. One has this weird Boolean enum and the other makes two distinct messages have overlapping behaviors.)

ekr: I agree. And since I'm on the topic, I think that in any case where you're not supposed to explicitly receive extensible code points, an unknown enum is "decode_error"

davidben: Do enums ever note, syntactically, whether they're complete or not? NamedGroup certainly isn't, but it doesn't say anything. Interestingly, CipherSuite isn't even an enum. It's a uint8[2].

ekr: No, they don't. We could do illegal_parameter :)

davidben: Oh I meant that more suggesting that we should find a way to denote it in the enum.

ekr: Sure, I would be onboard with that. Maybe "final enum" by analogy to final
classes?

ekr: @davidben we can do this in a separate PR.

ekr:
Separately merged

PR #612, opened 2016-08-31
TLS 1.3 -> TLS 2.0

davegarrett:
There appears to be significant support on the mailing list for rebranding TLS 1.3 as TLS 2.0. See thread for rationale, specific proposed changes, and discussion:
https://www.ietf.org/mail-archive/web/tls/current/msg20938.html

Here is a PR for consideration pending discussion and evaluation of WG consensus.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1,5 +1,5 @@
 ---
-title: The Transport Layer Security (TLS) Protocol Version 1.3
+title: The Transport Layer Security (TLS) Protocol Version 2.0
 abbrev: TLS
 docname: draft-ietf-tls-tls13-latest
 category: std
@@ -326,7 +326,7 @@ informative:
        date: 2015-09
 --- abstract
 
-This document specifies version 1.3 of the Transport Layer Security
+This document specifies version 2.0 of the Transport Layer Security
 (TLS) protocol.  TLS allows client/server applications to
 communicate over the Internet in a way that is designed to prevent eavesdropping,
 tampering, and message forgery.
@@ -335,7 +335,7 @@ tampering, and message forgery.
 
 #  Introduction
 
-DISCLAIMER: This is a WIP draft of TLS 1.3 and has not yet seen significant security analysis.
+DISCLAIMER: This is a WIP draft of TLS 2.0 and has not yet seen significant security analysis.
 
 RFC EDITOR: PLEASE REMOVE THE FOLLOWING PARAGRAPH
 The source for this draft is maintained in GitHub. Suggested changes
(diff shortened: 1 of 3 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3941,17 +3944,38 @@ by IANA
   "Recommended": ecdsa_secp256r1_sha256, ecdsa_secp384r1_sha384,
   rsa_pss_sha256, rsa_pss_sha384, rsa_pss_sha512, ed25519.
 
+-  TLS ProtocolVersionID Registry: This registry formalizes the definition
+  of values previously labled ProtocolVersion values in {{RFC5246}}. The
```
xiaoyinl: Here is a typo "labled" -> "labeled"

davegarrett: Re-pushed with typo fix. Thanks.

davegarrett:
As to this repository itself, apparently you can just rename it and the old name will continue to be an alias of the new name:
https://help.github.com/articles/renaming-a-repository/

The associated repository description would also need a minor update.

After scouring things for more places renaming would be needed, I ~think~ that's the last spot (both literally the last spot that should be updated ,and figuratively, the last spot to worry about). If we do this change, however, updating everything will be best to avoid confusion, and just as a matter of proper record.

Just posting this here for completeness and as a reminder, in the event that we get consensus to make the change.

xiaoyinl:
index.html in gh-pages branch also needs to be updated.

davegarrett:
After the rename is done, yeah, we should probably have the chair spam contact emails for a variety of other drafts to inform them of the update, though, they should also catch it if TLS 2.0 goes to RFC first.

I don't think we necessarily need to update the "draft-ietf-tls-tls13" token, though.

davegarrett:
Version ID registry split off to PR #634. This PR is now based on PR #634 which is rebased onto current master.

davegarrett:
There was too much bitrot to rebase easily, so I've redone the needed changeset. Should be able to merge to current master if accepted, now.

@xiaoyinl Another quick review to make sure I didn't miss anything on this pass would be appreciated.

xiaoyinl:
@davegarrett It looks good to me.

jrchamp:
Did this ever get resolved? I heard "TLS 2017" or "TLS 4" so as to be a larger, different number when mentioned alongside SSL2 and SSL3. The 2017 one is nice because you get a sense for the actual time period that the standard was finalized. Thank you for the work being done to make this all possible.

kaduk:
> Did this ever get resolved?

There was a long thread on the mailing list in a comment period that just ended, so the chairs are still reviewing the thread and attempting to determine consensus.

davegarrett:
At this point, the only possibilities I see happening are TLS 4 or TLS 1.3, with the latter being more likely at this point. One way or another, this PR is obsolete and will either be closed and/or redone. I'm just leaving it open until we come to a final conclusion.

ekr:
Closing this to keep my TODO list clear.

PR #614, opened 2016-09-01
prohibit use of renegotiation to negotiate TLSv1.3

tomato42:
the renegotiation mechanism from TLSv1.2 and earlier is
deprecated, that deprecation extends to the ability to
renegotiate existing TLSv1.2 (or earlier) session to TLSv1.3

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1370,6 +1370,12 @@ ClientHello (without modification) except:
 If a server receives a ClientHello at any other time, it MUST send
 a fatal "unexpected_message" alert and close the connection.
 
+If a server established a TLS session at a previous version of the protocol
+and receives a TLS 1.3 ClientHello in a renegotiation, it MUST retain the
+previous protocol version. In particular, it MUST NOT negotiate TLS 1.3.
+A client that receives a TLS 1.3 ServerHello in an established session MUST
+send a fatal "protocol_version" alert and close the connection.
+
 Structure of this message:
 
 %%% Key Exchange Messages


draft-ietf-tls-tls13.md
```diff
@@ -1370,6 +1370,11 @@ ClientHello (without modification) except:
 If a server receives a ClientHello at any other time, it MUST send
 a fatal "unexpected_message" alert and close the connection.
 
+If a server receives a TLSv1.3 ClientHello inside already established TLS
+session (as would be the case of the renegotiation mechanism from previous
+versions of TLS protocol), it MUST send a fatal "illegal_parameter" alert and
+close the connection.
```
ekr: I tend to think that this should be unexpected_handshake

ekr:
Sorry, but I just realized you can't do this.  You need to prohibit negotiating back *up* to 1.3, but the client has to be able to offer 1.3.

Here's the Bugzilla thread on this
https://bugzilla.mozilla.org/show_bug.cgi?id=1292898

tomato42:
so the client MAY send a TLSv1.3 Client Hello in a TLSv1.2 session, but server MUST NOT negotiate TLSv1.3 then?

draft-ietf-tls-tls13.md
tomato42: unexpected_message indicates that the message can't be sent at this point in time, since it's TLSv1.2 session, sending ClientHello messages is valid, the problem are the contents of the message.

tomato42:
How about this?

draft-ietf-tls-tls13.md
```diff
@@ -1370,6 +1370,13 @@ ClientHello (without modification) except:
 If a server receives a ClientHello at any other time, it MUST send
 a fatal "unexpected_message" alert and close the connection.
 
+If a server receives a TLS v1.3 ClientHello inside already established TLS
```
davidben: Nit: It seems elsewhere we mostly say "TLS 1.3" rather than "TLS v1.3"

draft-ietf-tls-tls13.md
```diff
@@ -1370,6 +1370,13 @@ ClientHello (without modification) except:
...
+session (as would be the case of the renegotiation mechanism from previous
+versions of TLS protocol), it MUST NOT negotiate TLS v1.3 but continue
```
davidben: "inside already established" sounds like it might apply to tunneling a ClientHello over the application stream. "would" is also a bit string. I don't think the spec says you have to keep ClientHello.version fixed on renego. It's just that everyone does in practice due to past interop issues.

Perhaps:
If a server has established a TLS session at a previous version of the protocol and receives a TLS 1.3 ClientHello in a renegotiation, it MUST retain the previous protocol version. In particular, it MUST NOT negotiate TLS 1.3.

draft-ietf-tls-tls13.md
```diff
@@ -1370,6 +1370,13 @@ ClientHello (without modification) except:
...
+negotiation of TLS v1.2 or lower session. Client that receives a TLS v1.3
```
davidben: Nit: A client that receives [...]

davidben:
OpenSSL derived implementations don't even allow you to switch versions on renego at all because that would be insane. We do send the higher ClientHello version for the reasons EKR pointed out, but the ServerHello version must match.

I would actually argue that it is fair game to assume version-related things are implicitly forked and locked (a 1.1 session implicitly cannot be resumed at 1.2, etc.), but the spec should be much clearer on this, so saying it explicitly now is a good idea.

ekr:
Yes.

richsalz:
I agree, resuming sessions must be *at the same version* and text saying so is a good idea.

davidben:
@richsalz: This is actually about negotiating at TLS 1.2 and then renegotiating to TLS 1.3. It's not clear how you would even manage to resume a TLS 1.2 session at TLS 1.3 or vice versa, but explicit text to this effect would be good. It has certainly historically been a mess.

tomato42:
OK, that should fix all the comments.

ekr:
Merged. @tomato42 please send me a PR adding yourself to Contributors

tomato42:
@ekr here it is: #617

PR #615, opened 2016-09-01
Attempt to add Finished to the ClientHello.

ekr:
1. Only allow one PSK
2. Add a Finished extension
3. Remove the explicit 0-RTT Finished.

The idea here is:
- Eliminate a key change because you don't need to encrypt
  this Finished. Also remove a state machine entry.
- Prove that the client knows the PSK even without 0-RTT
- Avoid the somewhat clunky resumption_context mechanism
  by adding a function of the PSK directly into the handshake transcript.

There are a bunch of open questions here:

1. We are now computing the early_traffic_secret as a function of
   a transcript which was already partly derived from PSK, which is
   creepy. OTOH, we were already doing that for a number of other
   secrets.

2. Can we in fact get rid of resumption_context here (Antoine
   argued yes, but I believe there were some concerns about the
   use of HKDF this way)

3. This is clunky to implement.

4. Should we move ticket_age into a separate extension so it can
   be used without 0-RTT.

Probably some other stuff I have forgotten.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -783,7 +783,7 @@ contains a "key_share" extension with the server's ephemeral
 Diffie-Hellman share which MUST be in the same group as one of the
 client's shares. If PSK key establishment is
 in use, then the ServerHello contains a "pre_shared_key"
-extension indicating which of the client's offered PSKs was selected.
+extension.
 Note that implementations can use (EC)DHE and PSK together, in which
 case both extensions will be supplied.
 
@@ -918,7 +918,9 @@ Initial Handshake:
 Subsequent Handshake:
        ClientHello
          + pre_shared_key
-         + key_share*            -------->
+         + key_share*
+         + hello_finished*
+                                 -------->
                                                        ServerHello
                                                   + pre_shared_key
                                                       + key_share*
@@ -957,7 +959,7 @@ as with a 1-RTT handshake with PSK resumption.
            + early_data
            + pre_shared_key
            + key_share*
-         (Finished)
+           + hello_finished*
          (Application Data*)
          (end_of_early_data)       -------->
                                                          ServerHello
@@ -1311,11 +1313,10 @@ following four sets of options in its ClientHello:
   (EC)DHE shares for some or all of these groups.
 - A "signature_algorithms" ({{signature-algorithms}}) extension which indicates the signature
   algorithms which the client can accept.
-- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains the identities
-  of symmetric keys known to the client and the key exchange
-  modes which each PSK supports.
+- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains a symmetric
+  key known to the client and the key exchange  modes which it supports.
 
-If the server does not select a PSK, then the first three of these
+If the server does not select PSK, then the first three of these
 options are entirely orthogonal: the server independently selects a
 cipher suite, an (EC)DHE group and key share for key establishment,
 and a signature algorithm/certificate pair to authenticate itself to
@@ -1336,7 +1337,7 @@ The server indicates its selected parameters in the ServerHello as
 follows:
 
 - If PSK is being used then the server will send a
-"pre_shared_key" extension indicating the selected key.
+"pre_shared_key" extension.
 - If PSK is not being used, then (EC)DHE and certificate-based
 authentication are always used.
 - When (EC)DHE is in use, the server will also provide a
@@ -1617,6 +1618,7 @@ The extension format is:
            pre_shared_key(41),
            early_data(42),
            cookie(44),
+           hello_finished(45),
            (65535)
        } ExtensionType;
 
@@ -2063,34 +2065,35 @@ The "extension_data" field of this extension contains a
        enum { psk_auth(0), psk_sign_auth(1), (255) } PskAuthenticationMode;
 
        struct {
-           PskKeyExchangeMode ke_modes<1..255>;
-           PskAuthenticationMode auth_modes<1..255>;
-           opaque identity<0..2^16-1>;
-       } PskIdentity;
-
-       struct {
            select (Handshake.msg_type) {
                case client_hello:
-                   PskIdentity identities<6..2^16-1>;
+                  opaque identity<0..2^16-1>;
+                   PskKeyExchangeMode ke_modes<1..255>;
+                   PskAuthenticationMode auth_modes<1..255>;
 
                case server_hello:
-                   uint16 selected_identity;
+                   // empty.
            }
        } PreSharedKeyExtension;
 
-identities
-: A list of the identities (labels for keys) that the client is willing
-  to negotiate with the server. If sent alongside the "early_data"
-  extension (see {{early-data-indication}}), the first identity is the
-  one used for 0-RTT data.
+identity
+: An identity (labels for a key) that the client is willing
+  to negotiate with the server.
+
+ke_modes
+: The key exchange modes that this key can be used with, in order
+of client preference, with most preferred first.
+
+auth_modes
+: The authentication modes that this key can be used with, in order
+of client preference, with most preferred first.
 
 selected_identity
 : The server's chosen identity expressed as a (0-based) index into
   the identities in the client's list.
 {: br}
 
-
-Each PSK offered by the client also indicates the authentication and
+The PSK offered by the client also indicates the authentication and
 key exchange modes with which the server can use it, with each
 list being in the order of the client's preference, with most
 preferred first. Any PSK MUST only be used with a single HKDF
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2063,34 +2065,35 @@ The "extension_data" field of this extension contains a
...
-           PskKeyExchangeMode ke_modes<1..255>;
-           PskAuthenticationMode auth_modes<1..255>;
-           opaque identity<0..2^16-1>;
-       } PskIdentity;
-
-       struct {
            select (Handshake.msg_type) {
                case client_hello:
-                   PskIdentity identities<6..2^16-1>;
+                  opaque identity<0..2^16-1>;
```
davidben: Nit: space

draft-ietf-tls-tls13.md
```diff
@@ -2063,34 +2065,35 @@ The "extension_data" field of this extension contains a
...
            }
        } PreSharedKeyExtension;
 
-identities
-: A list of the identities (labels for keys) that the client is willing
-  to negotiate with the server. If sent alongside the "early_data"
-  extension (see {{early-data-indication}}), the first identity is the
-  one used for 0-RTT data.
+identity
+: An identity (labels for a key) that the client is willing
```
davidben: labels => label?

draft-ietf-tls-tls13.md
```diff
@@ -2124,24 +2127,31 @@ mode, the server MUST send Certificate ({{certificate}}) and CertificateVerify
 {:/comment}
 {:br}
 
-In order to accept PSK key establishment, the server sends a
-"pre_shared_key" extension with the selected identity.
+Each PSK MUST be associated with a single Hash algorithm. For PSKs established
+via the ticket mechanism ({{NewSessionTicket}}), this is the Hash used for
+the KDF. For externally established PSKs, the Hash algorithm is set when the
+PSK is established.
```
davidben: The 'New Session Ticket Message' session says this, but probably worth being clear here that NST PSKs not only are constrained by hash but also the full cipher suite.

davidben: Actually, I'm not sure I understand what "the Hash algorithm is set when the PSK is established" means. Does that mean when we negotiate PSK or that it is set ahead of time at some external time?

draft-ietf-tls-tls13.md
```diff
@@ -2205,8 +2214,7 @@ can behave in one of two ways:
   its followup ClientHello.
 
 In order to accept early data, the server server MUST have accepted a
-PSK cipher suite and selected the the first key offered in the
-client's "pre_shared_key" extension. In addition, it MUST verify that
+PSK cipher suite. In addition, it MUST verify that
```
davidben: What's a PSK cipher suite? "MUST have accepted the PSK" perhaps?

draft-ietf-tls-tls13.md
```diff
@@ -2312,7 +2320,47 @@ the existing behavior for SignedCertificateTimestamps {{?RFC6962}},
...
+This means that the "hello_finished" is computed over a semantically
+valid ClientHello message where the length of the message is consistent
+with the number of bytes hashed.
+
+The actual ClientHello, complete with the correct verify_data, is included
+in all other handshake hash computations.
+
+[[OPEN ISSUE: Should we instead hash the first length - Hash.length() bytes?]]
+
+## Server Parameters
```
davidben: Nit: This seems to have lost the 'Messages' suffix which all the other sections have. Although "parameters messages" sounds kinda weird. "Server Parameter Messages" maybe??

draft-ietf-tls-tls13.md
```diff
@@ -2312,7 +2320,47 @@ the existing behavior for SignedCertificateTimestamps {{?RFC6962}},
...
-## Server Parameters Messages
+### Hello Finished
+
+The "hello_finished" extension is used to carry a Finished.verify_data
+({{finished}}) in the ClientHello. The reason for this structure rather
+than sending the Finished as a separate message is for backward
+compatibility: the client can send a "hello_finished" extension
+to servers which may not support TLS 1.3 without causing them to
+fail the handshake. This allows for mixed server configurations which
+concurrently support TLS 1.3 (including PSK resumption but not 0-RTT)
```
davidben: Nit: Maybe worth a link to the 0-RTT backwards compat section here.

draft-ietf-tls-tls13.md
```diff
@@ -2312,7 +2320,47 @@ the existing behavior for SignedCertificateTimestamps {{?RFC6962}},
...
+alert.
+
+The contents of the extension are simply a Finished value, with the
+Handshake Context being the ClientHello message with the verify_data
+field of the extension being set to a string of Hash.length zeroes.
+One way to implement this is:
+
+1. Marshall the entire ClientHello with zeroes in the verify_data
+   location.
+2. Compute the "hello_finished" extension
```
davidben: Nit: missing a period

draft-ietf-tls-tls13.md
```diff
@@ -2494,7 +2541,7 @@ for each scenario:
 
 | Mode | Handshake Context | Base Key |
 |------|-------------------|----------|
-| 0-RTT | ClientHello | early_traffic_secret|
+| PSK | ClientHello prefix with 0s for extension body | pre-shared key |
```
davidben: Nit: Perhaps extension body -> hello_finished body, so it's clearer which extension?

draft-ietf-tls-tls13.md
```diff
@@ -2715,18 +2762,16 @@ The context string for a server signature is
 and for a client signature is "TLS 1.3, client
 CertificateVerify".
 
-For example, if Hash(Handshake Context + Certificate) was 32 bytes
-of 01 and Hash(resumption_context) was 32 bytes of 02 (these lengths
-would make sense for SHA-256, the input to the final signing process
-for a server CertificateVerify would be:
+For example, if Hash(Handshake Context + Certificate) was 32 bytes of
+01 (this length would make sense for SHA-256, the input to the final
```
davidben: This was in the original too, but 'SHA-256' needs a ')' after it.

draft-ietf-tls-tls13.md
```diff
@@ -2843,12 +2887,10 @@ from the resumption master secret:
    resumption_psk = HKDF-Expand-Label(
```
davidben: Above here, "the following two values" => "the following value"

draft-ietf-tls-tls13.md
```diff
@@ -2843,12 +2887,10 @@ from the resumption master secret:
...
                         resumption_secret,
                         "resumption psk", "", Hash.Length)
-
-   resumption_context = HKDF-Expand-Label(
-                            resumption_secret,
-                            "resumption context", "", Hash.Length)
 ~~~~
 
+[[TODO: Remove resumption context if we come to consensus on Hello Finished]]
```
davidben: (Already removed in this PR?)

davidben:
Various editorial nitpicks aside, I think this looks good. No opinions on whether we hash a ClientHello prefix or stick zeros in there. If we pick one and it's more natural for an implementation to do the other, they can just subtract 32 or call HashUpdate("\x00" x 32).

draft-ietf-tls-tls13.md
```diff
@@ -918,7 +918,9 @@ Initial Handshake:
 Subsequent Handshake:
        ClientHello
          + pre_shared_key
-         + key_share*            -------->
+         + key_share*
+         + hello_finished*
```
martinthomson: Bike shed (hence here and not on the list):

If we are to avoid confusion about this, could we call this a MAC rather than Finished.  Maybe `client_hello_mac`.  This isn't the end, so Finished is only a product of the (convoluted) process we used to reach this point.

draft-ietf-tls-tls13.md
```diff
@@ -2063,34 +2065,35 @@ The "extension_data" field of this extension contains a
...
-  to negotiate with the server. If sent alongside the "early_data"
-  extension (see {{early-data-indication}}), the first identity is the
-  one used for 0-RTT data.
+identity
+: An identity (labels for a key) that the client is willing
+  to negotiate with the server.
+
+ke_modes
+: The key exchange modes that this key can be used with, in order
+of client preference, with most preferred first.
```
martinthomson: Nit: alignment

draft-ietf-tls-tls13.md
kaduk: Yes, let's use a better name.

draft-ietf-tls-tls13.md
```diff
@@ -1311,11 +1313,10 @@ following four sets of options in its ClientHello:
   (EC)DHE shares for some or all of these groups.
 - A "signature_algorithms" ({{signature-algorithms}}) extension which indicates the signature
   algorithms which the client can accept.
-- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains the identities
-  of symmetric keys known to the client and the key exchange
-  modes which each PSK supports.
+- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains a symmetric
```
kaduk: s/contains/indicates/?  "Don't send the key itself in the extension."

draft-ietf-tls-tls13.md
```diff
@@ -1311,11 +1313,10 @@ following four sets of options in its ClientHello:
...
+  key known to the client and the key exchange  modes which it supports.
```
kaduk: nit: two spaces between exchange and modes.

draft-ietf-tls-tls13.md
kaduk: My impiression was that the entity creating the PSK would dictate a single hash algorithm to use with it.  (For pure-PSK, that is.)

draft-ietf-tls-tls13.md
```diff
@@ -2124,24 +2127,31 @@ mode, the server MUST send Certificate ({{certificate}}) and CertificateVerify
...
+via the ticket mechanism ({{NewSessionTicket}}), this is the Hash used for
+the KDF. For externally established PSKs, the Hash algorithm is set when the
+PSK is established.
+
+Prior to accepting PSK key establishment, the server MUST validate the
+"hello_finished" extension. If this extension is not present, the server
+MUST send a fatal "missing_extension" alert. If it does not validate, the
+server MUST send a fatal "decrypt_error" alert.
+
+In order to accept PSK key establishment, the server sends an empty
```
kaduk: Remind me why there is no value from the server sending a MAC as well for key confirmation?

draft-ietf-tls-tls13.md
```diff
@@ -2124,24 +2127,31 @@ mode, the server MUST send Certificate ({{certificate}}) and CertificateVerify
...
+"pre_shared_key" extension.
+
 Clients MUST verify that the server's selected_identity is within the
```
kaduk: selected_identity is gone (there's also a straggler up around L2091)

draft-ietf-tls-tls13.md
```diff
@@ -2312,7 +2320,47 @@ the existing behavior for SignedCertificateTimestamps {{?RFC6962}},
...
+compatibility: the client can send a "hello_finished" extension
+to servers which may not support TLS 1.3 without causing them to
+fail the handshake. This allows for mixed server configurations which
+concurrently support TLS 1.3 (including PSK resumption but not 0-RTT)
+and TLS 1.2 without creating interoperability problems.
+
+The "hello_finished" extension MUST be sent if the client offers a
+"pre_shared_key" extension and MUST NOT be sent otherwise. The Hash
+used in the computation is that associated with the indicated PSK.
+Servers MUST NOT send this extension. This extension MUST be the
```
kaduk: In general I'm not super-keen on ordering requirements within the extensions block, "but this one has a good reason so it's okay".

PR #616, opened 2016-09-02
add SHA-3 hashes

tomato42:
since SHA-3 is now finished, we can define signatures that use it


ekr:
@tomato42 I'm not seeing WG consensus for this. @seanturner can you advise?

seanturner:
@ekr Based on this [email](https://mailarchive.ietf.org/arch/msg/tls/OSzZcLuSyO7n87iKti4DZI-oMKo) you can close this out.

PR #621, opened 2016-09-02
Error handling specification

tomato42:
While the specification already details a lot of situations that "MUST NOT" happen or that are errors, it doesn't specify how those situations should be handled by peers.

Add those details.


ekr:
@tomato42. Thanks for this PR, but I think I want to go in a different direction. In general, I'd like the descriptions of the alerts to be clear enough that you can infer which alert to send without us having to tell you explicitly. That won't always be true, in which case we will need to specify, but I'd like to avoid calling out each error condition. This applies in particular to "illegal_parameter" and "decode_error". If you'd like to suggest clarifications to those alert descriptions and then reduce this PR to just the cases which seem to be underspecified, I can take a look

tomato42:
Reading the current draft I had a feeling as if error handling for 70 or 80% of cases was already specified, the patches just put the description of an error with the way to handle it. I can't guarantee that they don't duplicate the error handling description (I have read the draft from cover to cover just once...).

And actually, I found some places where on first read you'd think that you need to send alerts where you don't, so I think that being explicit in all places is clearer for an implementer.

In case that didn't convince you: I guess you'd like to start with 1cfa45e and go from there?

draft-ietf-tls-tls13.md
```diff
@@ -1503,7 +1527,8 @@ random
 
 cipher_suite
 : The single cipher suite selected by the server from the list in
-  ClientHello.cipher_suites.
+  ClientHello.cipher_suites. If the value was not advertised in that list
+  client MUST send a fatal "illegal_parameter" alert and close the connection.
```
davegarrett: should be "the client"

draft-ietf-tls-tls13.md
```diff
@@ -1827,6 +1866,10 @@ trust anchors are not validated since they begin a certification path (see
 path MAY use a signature algorithm that is not advertised as being supported
 in the "signature_algorithms" extension.
 
+In case no algorithms in the "signature_algorithms" extension sent by client
```
davegarrett: should be "In the case that no" and "the client"

draft-ietf-tls-tls13.md
```diff
@@ -1905,6 +1948,10 @@ Finite Field Groups (DHE)
 Items in named_group_list are ordered according to the client's
 preferences (most preferred choice first).
 
+If server receives a list that does not include any group acceptable
```
davegarrett: "the server"

draft-ietf-tls-tls13.md
```diff
@@ -1905,6 +1948,10 @@ Finite Field Groups (DHE)
...
+to it, it MUST abort the connection by sending "handshake_failure"
```
davegarrett: I'd rearrange this to avoid the "it, it". Instead: "does not include any acceptable group, it MUST"

draft-ietf-tls-tls13.md
```diff
@@ -2155,6 +2226,10 @@ in its first flight of messages. If the client opts to do so, it MUST
 supply an "early_data" extension as well as the "pre_shared_key"
 extension.
 
+If a server receives a ClientHello that does not include a "pre_shared_key"
+but includes "early_data" extension MUST send a fatal "illegal_parameter"
```
davegarrett: "it MUST"

draft-ietf-tls-tls13.md
```diff
@@ -2452,8 +2535,8 @@ certificate_extensions
   extensions.
 {:br }
 
-Note: It is a fatal "unexpected_message" alert for an anonymous server to
-request client authentication.
+Clients that receive CertificateRequest message from an anonymous server
```
davegarrett: "receive a"

draft-ietf-tls-tls13.md
```diff
@@ -2521,7 +2604,9 @@ requested client authentication via a CertificateRequest message
 ({{certificate-request}}). If the server requests client authentication
 but no suitable certificate is available, the client
 MUST send a Certificate message containing no certificates (i.e., with
-the "certificate_list" field having length 0).
+the "certificate_list" field having length 0). Server that receives a
+Certificate message when it haven't sent a CertificateRequest MUST
```
davegarrett: typo: "haven't" should be "hasn't"

draft-ietf-tls-tls13.md
```diff
@@ -2779,6 +2867,11 @@ correct. Once a side has sent its Finished message and received and
 validated the Finished message from its peer, it may begin to send and
 receive application data over the connection.
 
+If the contents of the message are not correct recipient MUST send a
+fatal "bad_record_mac".
```
davegarrett: missing word "alert" at the end

davegarrett:
I decided to give this a quick review in case it was pruned down to the point where we'd want to take it. (I'm generally in favor of more explicitness in error specifying than Eric, but I do agree that there's probably too much here.) Unfortunately, there are significant grammatical issues here. I counted about 14 more cases where a sentence was missing a "the" or "a" word, in addition to those I already noted. I'm going to guess that Czech does not have counterparts for those words (at least in the given cases). Poke me in the future if/when you do a revision and I'll do a recheck.

draft-ietf-tls-tls13.md
```diff
@@ -1355,8 +1355,9 @@ MUST return a "handshake_failure" alert and close the connection.
 When this message will be sent:
 
 > When a client first connects to a server, it is REQUIRED to send the
-ClientHello as its first message. The client will also send a
-ClientHello when the server has responded to its ClientHello with a
+ClientHello as its first message. Server which receives a different first
+message MUST send a fatal "unexpected_message" alert. The client will also send
```
briansmith: Neither the server nor the client is required to send any alert in any situation. if they send an alert, they are required to send a particular one, in particular situations, but they can always opt to not send one and instead, for example, just close the connection without sending anything.

ekr:
@tomato42 I've put up a new PR  #625 that moves in the direction I want to go here, which is to provide general rules and only specific alerts where it's unobvious or where I already have to say that you should abort. So I'm going to close this one. Please feel free to call out any cases where you believe that that PR still isn't specific enough so we can evaluate case-by-case

PR #625, opened 2016-09-05
Clean up alert requirements. Fixes #254. This is a long-delayed chang…

### LINKED ISSUE #254: Change alert requirements to be consistent
Consensus at the interim was:

- All alerts shall be SHOULD to send (or may in some cases)
- In cases where alerts SHOULD be sent, there will be MUSTs for which one.

ekr:
…e from the

9/2015 interim. Specifically:

- Use common language for failures
- Don't require implementations to send alerts, but do require them
  to send specific alerts if they send them.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -837,10 +837,9 @@ If the client has not provided a sufficient "key_share" extension (e.g., it
 includes only DHE or ECDHE groups unacceptable or unsupported by the
 server), the server corrects the mismatch with a HelloRetryRequest and
 the client needs to restart the handshake with an appropriate
-"key_share" extension, as shown in Figure 2.
+"key_share" extension, as shown in Figure 2. 
 If no common cryptographic parameters can be negotiated,
-the server sends a "handshake_failure" or "insufficient_security"
-fatal alert (see {{alert-protocol}}).
+the server MUST abort the handshake with an appropriate alert.
 
 ~~~
          Client                                               Server
@@ -1346,8 +1345,11 @@ authentication are always used.
 subsequently send Certificate ({{certificate}}) and
 CertificateVerify ({{certificate-verify}}) messages.
 
-If the server is unable to negotiate a supported set of parameters (i.e., there is no overlap between the client and server parameters), it
-MUST return a "handshake_failure" alert and close the connection.
+If the server is unable to negotiate a supported set of parameters
+(i.e., there is no overlap between the client and server parameters),
+it MUST abort the handshake and and SHOULD send either
+a "handshake_failure" or "insufficient_security" fatal alert
+(see {{alert-protocol}}).
 
 
 ###  Client Hello
@@ -1370,14 +1372,14 @@ ClientHello (without modification) except:
 - Including a "cookie" extension if one was provided in the
   HelloRetryRequest.
   
-If a server receives a ClientHello at any other time, it MUST send
-a fatal "unexpected_message" alert and close the connection.
+If a server receives a ClientHello at any other time, it MUST terminate
+the connection.
 
 If a server established a TLS connection with a previous version of TLS
 and receives a TLS 1.3 ClientHello in a renegotiation, it MUST retain the
 previous protocol version. In particular, it MUST NOT negotiate TLS 1.3.
 A client that receives a TLS 1.3 ServerHello during renegotiation
-MUST send a fatal "protocol_version" alert and close the connection.
+MUST abort the handshake with a "protocol_version" alert.
 
 Structure of this message:
 
@@ -1447,7 +1449,7 @@ legacy_compression_methods
   zero, which corresponds to the "null" compression method in
   prior versions of TLS. If a TLS 1.3 ClientHello is
   received with any other value in this field, the server MUST
-  generate a fatal "illegal_parameter" alert. Note that TLS 1.3
+  abort the handshake with an "illegal_parameter" alert. Note that TLS 1.3
   servers might receive TLS 1.2 or prior ClientHellos which contain
   other compression methods and MUST follow the procedures for
   the appropriate prior version of TLS.
@@ -1462,11 +1464,13 @@ In the event that a client requests additional functionality using
 extensions, and this functionality is not supplied by the server, the
 client MAY abort the handshake. Note that TLS 1.3 ClientHello messages
 MUST always contain extensions, and a TLS 1.3 server MUST respond to
-any TLS 1.3 ClientHello without extensions with a fatal "decode_error"
+any TLS 1.3 ClientHello without extensions or with data following
+the extensions block with a "decode_error"
 alert. TLS 1.3 servers may receive TLS 1.2 ClientHello messages
 without extensions. If negotiating TLS 1.2, a server MUST check that
-the amount of data in the message precisely matches one of these
-formats; if not, then it MUST send a fatal "decode_error" alert.
+the message either contains no data after legacy_compression_methods
+or that it contains a valid extensions block with no data following.
+If not, then it MUST abort the handshake with a "decode_error" alert.
 
 After sending the ClientHello message, the client waits for a ServerHello
 or HelloRetryRequest message.
@@ -1544,7 +1548,7 @@ TLS 1.3 clients receiving a TLS 1.2 or below ServerHello MUST check
 that the last eight octets are not equal to either of these values. TLS
 1.2 clients SHOULD also perform this check if the ServerHello
 indicates TLS 1.1 or below. If a match is found, the client MUST abort
-the handshake with a fatal "illegal_parameter" alert. This mechanism
+the handshake with a "illegal_parameter" alert. This mechanism
 provides limited protection against downgrade attacks over and above
 that provided by the Finished exchange: because the ServerKeyExchange
 includes a signature over both random values, it is not possible for
@@ -1564,9 +1568,9 @@ When this message will be sent:
 message if they were able to find an acceptable set of algorithms and
 groups that are mutually supported, but
 the client's ClientHello did not contain sufficient information to
-proceed with the handshake.
-If a server cannot successfully select algorithms, it will respond with a
-fatal "handshake_failure" alert instead.
+proceed with the handshake. 
+If a server cannot successfully select algorithms, MUST abort
+the handshake with a "handshake_failure" alert.
 
 Structure of this message:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1544,7 +1548,7 @@ TLS 1.3 clients receiving a TLS 1.2 or below ServerHello MUST check
 that the last eight octets are not equal to either of these values. TLS
 1.2 clients SHOULD also perform this check if the ServerHello
 indicates TLS 1.1 or below. If a match is found, the client MUST abort
-the handshake with a fatal "illegal_parameter" alert. This mechanism
+the handshake with a "illegal_parameter" alert. This mechanism
```
davegarrett: an

draft-ietf-tls-tls13.md
```diff
@@ -1564,9 +1568,9 @@ When this message will be sent:
 message if they were able to find an acceptable set of algorithms and
 groups that are mutually supported, but
 the client's ClientHello did not contain sufficient information to
-proceed with the handshake.
-If a server cannot successfully select algorithms, it will respond with a
-fatal "handshake_failure" alert instead.
+proceed with the handshake. 
+If a server cannot successfully select algorithms, MUST abort
```
davegarrett: "it MUST" or "then it MUST"

draft-ietf-tls-tls13.md
```diff
@@ -1988,15 +1993,15 @@ generated independently.  Clients MUST NOT offer multiple
 KeyShareEntry values for the same group.  Clients MUST NOT offer any
 KeyShareEntry values for groups not listed in the client's
 "supported_groups" extension.  Servers MAY check for violations of
-these rules and and MAY abort the connection with a fatal
+these rules and and MAY abort the handshake with a
```
davegarrett: an

draft-ietf-tls-tls13.md
```diff
@@ -1988,15 +1993,15 @@ generated independently.  Clients MUST NOT offer multiple
...
 "illegal_parameter" alert if one is violated.
 
 Upon receipt of this extension in a HelloRetryRequest, the client MUST first
 verify that the selected_group field corresponds to a group which was provided
 in the "supported_groups" extension in the original ClientHello. It MUST then
 verify that the selected_group field does not correspond to a group which was
 provided in the "key_share" extension in the original ClientHello. If either of
-these checks fails, then the client MUST abort the handshake with a fatal
+these checks fails, then the client MUST abort the handshake with a
```
davegarrett: an

draft-ietf-tls-tls13.md
```diff
@@ -2012,7 +2017,7 @@ Servers MUST NOT send a KeyShareEntry for any group not
 indicated in the "supported_groups" extension.
 If a HelloRetryRequest was received, the client MUST verify that the
 selected NamedGroup matches that supplied in the selected_group field and MUST
-abort the connection with a fatal "illegal_parameter" alert if it does not.
+abort the connection with a "illegal_parameter" alert if it does not.
```
davegarrett: an

draft-ietf-tls-tls13.md
```diff
@@ -3009,7 +3012,7 @@ Handshake messages MUST NOT span key changes. Because
 the ServerHello, Finished, and KeyUpdate messages signal a key change,
 upon receiving these messages a receiver MUST verify that the end
 of these messages aligns with a record boundary; if not, then it MUST
-send a fatal "unexpected_message" alert.
+terminate the connection with a "unexpected_message" alert.
```
ekr: an

draft-ietf-tls-tls13.md
```diff
@@ -3023,8 +3026,8 @@ be multiplexed over the same record layer. This document specifies
 three content types: handshake, application data, and alert.
 Implementations MUST NOT send record types not defined in this
 document unless negotiated by some extension. If a TLS implementation
-receives an unexpected record type, it MUST send an
-"unexpected_message" alert.  New record content type values are
+receives an unexpected record type, it MUST terminate the connection
+with a "unexpected_message" alert.  New record content type values are
```
ekr: an

draft-ietf-tls-tls13.md
```diff
@@ -2139,13 +2144,13 @@ Clients MUST verify that the server's selected_identity is within the
 range supplied by the client and that the "key_share" and
 "signature_algorithms" extensions are consistent with the
 indicated ke_modes and auth_modes values. If these values
-are not consistent, the client MUST generate an "illegal_parameter"
-alert and close the connection.
+are not consistent, the client MUST abort the handshake with
+a "illegal_parameter" alert.
```
davegarrett: an

ekr: an

davegarrett: Yeah, you used "an" on a few and "a" for the rest. I'll stop spamming the comments with "an" now if you're going to take over on that. :p

ekr: Actually, what I believe happened here is that I did s/fatal "/"/ and so there were a pile of 'a fatal "unsupported...' or whatever, and they need to be 'an "unsupported...'

draft-ietf-tls-tls13.md
```diff
@@ -3422,16 +3425,25 @@ before destroying the transport.
...
+parties immediately close the connection.
+
+Whenever an implementation encounters a fatal error condition, it
+SHOULD send an appropriate fatal alert and MUST close the connection
+without sending or receiving any additional data. In the rest of this
+specification, the phrase "{terminate the connection, abort the
+handshake}" is used without a specific alert means that the
+implementation SHOULD send the alert indicated by the descriptions
+below. The phrase "{terminate the connection, abort the handshake}
+with a X alert" MUST send alert X if it sends any alert. All
```
davegarrett: I'd add a sentence here stating that any language defining an error response that doesn't use this language is still a MUST in all cases, not a SHOULD, to be clear.

draft-ietf-tls-tls13.md
```diff
@@ -3488,6 +3503,9 @@ access_denied
 decode_error
 : A message could not be decoded because some field was out of the
   specified range or the length of the message was incorrect.
+  This alert is used for errors where the message does not conform
+  to the formal protocol syntax (e.g., a vector is specified as
+  having the range <2..255> but is encoded as having 0 length.)
```
davegarrett: nit: period should be after the closing parenthesis here

davegarrett:
Should we have a blurb somewhere stating a recommendation for what sort of use cases we would expect a silent fail? The main one that comes to mind is a server expecting PSK for client auth which could just reject invalid connections with no further response.

(side-note: I wish GitHub had a "submit all comments" button to allow reviewing and sending all at once in with a single notification. Instead, I just comment as I go, but that can get spammy quickly. :/ )

ekr:
My sense is that it's not any particular use case that has a silent fail but rather implementations that do.

draft-ietf-tls-tls13.md
```diff
@@ -1370,14 +1372,14 @@ ClientHello (without modification) except:
 - Including a "cookie" extension if one was provided in the
   HelloRetryRequest.
   
-If a server receives a ClientHello at any other time, it MUST send
-a fatal "unexpected_message" alert and close the connection.
+If a server receives a ClientHello at any other time, it MUST terminate
```
wbl: Why this requirement?

ekr: Can you clarify your question? this is the same requirement as before.

The reason is that TLS 1.3 forbids renegotiation.

draft-ietf-tls-tls13.md
tomato42: not insisting on it, but maybe also add "or a struct specified as having 4 fields is received with only 3 fields present"?

ekr: It seems like the general form of this is "a too short encoding"?

tomato42: no, since also encoding with a vector length of 3 when the values in vector are uint16 is malformed, and so is encoding of a struct that has additional fields after the ones that are defined

ekr: I'm not following. There are effectively two options here:
- You have a buffer which should coterminate with the production and therefore you can detect both too short and too long
- You have a buffer which contains the production and then some other stuff afterwards in which case you can only detect the buffer being too short while processing that production.

I agree that an odd-length vector with even-sized elements is a distinct kind of error. My point is that "4 fields is received with only 3 fields present"  is a general case of too short encoding.

tomato42: yes, that would be a general case of too short encoding

I was proposing adding an example that clearly shows that the alert applies both to vectors and structs

kaduk: I think it's confusing to try to think about "only received 3 fields when 4 fields expected"; it's more like "only received enough data for 3 fields of the 4-field struct expected, which may indicate the sender sent a different type than was expected", since the bits on the wire are not tagged with an identifier for what structure they represent.

draft-ietf-tls-tls13.md
```diff
@@ -837,10 +837,9 @@ If the client has not provided a sufficient "key_share" extension (e.g., it
 includes only DHE or ECDHE groups unacceptable or unsupported by the
 server), the server corrects the mismatch with a HelloRetryRequest and
 the client needs to restart the handshake with an appropriate
-"key_share" extension, as shown in Figure 2.
+"key_share" extension, as shown in Figure 2.
```
kaduk: Trailing whitespace

kaduk:
The first time alerts come up in the current document is in the "Incorrect DHE Share" section.
I would consider adding a note at the end of the first paragraph in section 2 like "A failure of the handshake or other protocol error triggers the termination of the connection, optionally preceded by an alert message that may or may not be authenticated depending on how far the handshake progressed."

draft-ietf-tls-tls13.md
```diff
@@ -1589,8 +1593,9 @@ offered by the client in its ClientHello, with the exception of optionally the
 "cookie" (see {{cookie}}) extension.
 
 Upon receipt of a HelloRetryRequest, the client MUST verify that the extensions
-block is not empty and otherwise abort the handshake with a fatal
-"decode_error" alert.  Clients SHOULD also abort with "unexpected_message" in
+block is not empty and MUST abort the handshake with a
```
kaduk: This text says that a client that receives a HRR MUST abort the handshake.  (That should actually only happen if the extensions block is empty.)
Probably just "is not empty and abort the handshake with a "decode_error" alert if it is empty." would suffice.

PR #629, opened 2016-09-07
Cleanup

kazu-yamamoto:
Cleaning up the meanings of structure members.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2880,20 +2880,20 @@ to that negotiated connection where the ticket was established.
 
 %%% Ticket Establishment
 
-     enum { ticket_early_data_info(1), (65535) } TicketExtensionType;
+       enum { ticket_early_data_info(1), (65535) } TicketExtensionType;
 
-     struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3059,7 +3059,7 @@ Alert messages ({{alert-protocol}}) MUST NOT be fragmented across records.
            ContentType type;
            ProtocolVersion legacy_record_version = { 3, 1 };    /* TLS v1.x */
            uint16 length;
-           opaque fragment[TLSPlaintext.length];
+           opaque fragment[length];
```
davegarrett: While technically true, the "length" here is defined one line above. Adding in the object explicitly within the object itself just makes it a little more complicated than needed. I'd leave this one as-is.

kazu-yamamoto: As commented in f4004cf8dc10f362a8e2657f31559b95ebf84b0d, "opaque encrypted_record[length]" is already used in TLSCiphertext. I'd suggest to fix one of them.

Issue #534, opened 2016-07-10
cookie extension ID


davegarrett:
The cookie hello extension ID is currently assigned as 44, however the ticket age extension that was at 43 has been dropped, so there's a gap in the list of new IDs now. If we have no new extension to retake 43, cookie should probably be reassigned it to keep a contiguous set of IDs. Filing as an issue as this would be a notable wire-format change.

ekr:
This doesn't seem like it makes life much better. Please take to the list if you want to make this change.

ekr:
Oops. Didn't mean to close just yet.

ekr:
I closed the gap with a different extension

PR #632, opened 2016-09-08
Versions in extensions

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -1412,11 +1412,16 @@ before extensions were defined.
 As of TLS 1.3, all clients and servers will send at least
 one extension (at least "key_share" or "pre_shared_key").
 
-max_supported_version
-: The latest (highest valued) version of the TLS protocol offered by the
-  client. This SHOULD be the same as the latest version supported. For this
-  version of the specification, the version will be { 3, 4 }. (See
-  {{backward-compatibility}} for details about backward compatibility.)
```
davegarrett: The back compat ref should probably stay on here.

draft-ietf-tls-tls13.md
```diff
@@ -1626,6 +1631,7 @@ The extension format is:
            pre_shared_key(41),
            early_data(42),
            cookie(44),
+           supported_versions(46),
```
davegarrett: 43 is free; can we just stick it there to appease nit issue #534?

draft-ietf-tls-tls13.md
```diff
@@ -4223,8 +4262,9 @@ whenever TLS 1.3 is used.
 ## Negotiating with an older server
 
 A TLS 1.3 client who wishes to negotiate with such older servers will send a
-normal TLS 1.3 ClientHello containing { 3, 4 } (TLS 1.3) in
-ClientHello.max_supported_version. If the server does not support this version it
+normal TLS 1.3 ClientHello containing { 3, 3 } (TLS 1.2) in
+ClientHello.max_supported_version but with the correct version in the
```
davegarrett: need to rename all use of max_supported_version to legacy_version

draft-ietf-tls-tls13.md
```diff
@@ -1692,6 +1698,35 @@ be taken into account when designing new extensions:
...
+prepared to negotiate (for this specification, that means minimally
+{3, 4}, but if previous versions of TLS are supported, they MUST
+be present as well).
+
+Servers which are compliant with this specification MUST use
+only the "supported_versions" extension to determine client preferences
+and MUST only select a version of TLS present in that extension.
+They MUST ignore any unknown versions. If the extension is not present,
+they MUST negotiate TLS 1.2 or prior as specified in {{RFC5246}},
+even if ClientHello.version is {3, 4} or later.
```
xiaoyinl: Shouldn't that be "ClientHello.legacy_version"?

draft-ietf-tls-tls13.md
```diff
@@ -1395,7 +1395,7 @@ Structure of this message:
        uint8 CipherSuite[2];    /* Cryptographic suite selector */
 
        struct {
-           ProtocolVersion max_supported_version = { 3, 4 };    /* TLS v1.3 */
+           ProtocolVersion legacy_version = { 3, 3 };    /* TLS v1.2 */
```
martinthomson: Why do we persist in using a struct for the version, rather than just changing to a uint16-based enum?

davegarrett: I'd vote for going uint16, as I did in my version of this proposal. One way or another, we really should do a formal registry, as was suggested on-list in the TLS 1.3 vs 2.0 discussion. I'll submit a PR with just that cherry-picked out of the rebranding PR, as it's a good idea to make this easier to follow, regardless of what we call all this.

davidben: Registry and uint16 SGTM. A registry will be useful anyway to burn things like private use values for drafts and GREASE values if the WG is so inclined.

davegarrett: I've pushed PR #634 with just the changeset for a registry and explicit naming of the values as IDs.

richsalz: And while we’re out it, yes, let’s just make it an integer so it maps to whatever x.y or x or /x we use in the future.

draft-ietf-tls-tls13.md
ekr: Done.

draft-ietf-tls-tls13.md
ekr: Done.

ekr:
Separately merged

PR #634, opened 2016-09-09
define protocol version ID registry

davegarrett:
This pulls a couple cleanup commits out of PR #612 that explicitly renames the protocol version fields as IDs and defines a registry for all values, as they're really just arbitrary codepoints at this point. With PR #632 now doing an updated version of my proposed list-based version negotiation extension, we've got these IDs in entirely too many places and need to just define a formal list to make things less messy.

IANA registry idea initially proposed on-list by @dkg.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -82,6 +82,7 @@ normative:
 
 
 informative:
+  RFC2246:
   RFC4086:
   RFC4279:
   RFC4346:
@@ -1466,7 +1467,13 @@ Structure of this message:
 
 %%% Key Exchange Messages
 
-       uint16 ProtocolVersion;
+       enum {
+           ssl2(0x0200), ssl3(0x0300),
+           tls1(0x0301), tls1_1(0x0302), tls1_2(0x0303), tls1_3(0x0304),
+           experimental_use (0xfeff..0xffff),
+           (0xffff)
+       } ProtocolVersion;
+
        opaque Random[32];
 
        uint8 CipherSuite[2];    /* Cryptographic suite selector */
@@ -1491,14 +1498,14 @@ one extension (at least "key_share" or "pre_shared_key").
 
 legacy_version
 : In previous versions of TLS, this field was used for version negotiation
-  and represented the highest version number supported by the client.
+  and represented the highest version ID supported by the client.
   Experience has shown that many servers do not properly implement
   version negotiation, leading to "version intolerance" in which
   the server rejects an otherwise acceptable ClientHello with a version
   number higher than it supports.  
   In TLS 1.3, the client indicates its version preferences in the
   "supported_versions" extension ({{supported-versions}}) and this field MUST
-  be set to 0x0303, which was the version number for TLS 1.2.
+  be set to 0x0303, which is the version ID for TLS 1.2.
   (See {{backward-compatibility}} for details about backward compatibility.)  
 
 random
@@ -1541,6 +1548,10 @@ extensions
   defined in {{hello-extensions}}.
 {:br }
 
+This document describes TLS Version 1.3, which uses the version ID 0x0304.
+This value is historical, deriving from the use of the values 0x0301
+for TLS 1.0 and 0x0300 for SSL 3.0.
+
 In the event that a client requests additional functionality using
 extensions, and this functionality is not supplied by the server, the
 client MAY abort the handshake. Note that TLS 1.3 ClientHello messages
@@ -1578,9 +1589,9 @@ Structure of this message:
 
 version
 : This field contains the version of TLS negotiated for this session.  Servers
-  MUST select a version from the list in ClientHello.supported_versions extension.
-  A client which receives a version that was not offered MUST abort the handshake.
-  For this version of the specification, the version is 0x0304.  (See
+  MUST select a version ID from the list in ClientHello.supported_versions extension.
+  A client which receives a version ID that was not offered MUST abort the handshake.
+  For this version of the specification, this value is 0x0304.  (See
   {{backward-compatibility}} for details about backward compatibility.)
 
 random
@@ -1795,7 +1806,7 @@ be taken into account when designing new extensions:
 
 The "supported_versions" extension is used by the client to indicate
 which versions of TLS it supports. The extension contains a list of
-supported versions in preference order, with the most preferred
+supported version IDs in preference order, with the most preferred
 version first. Implementations of this specification MUST send this
 extension containing all versions of TLS which they are
 prepared to negotiate (for this specification, that means minimally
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1383,10 +1383,7 @@ Structure of this message:
 
 %%% Key Exchange Messages
 
-       struct {
-           uint8 major;
-           uint8 minor;
-       } ProtocolVersion;
+       uint8 ProtocolVersionID[2];
```
martinthomson:
```
enum {
  ssl3_RESERVED (0x300),
  tls10 (0x301),
  tls11 (0x302),
  tls12 (0x303),
  tls13 (0x304),
  (65535)
} ProtocolVersion;
```

I don't see the point of this being `ProtocolVersionID` over just `ProtocolVersion`.

davegarrett: On Friday, September 09, 2016 01:39:30 am Martin Thomson wrote:
> I don't see the point of this being `ProtocolVersionID` over just `ProtocolVersion`.

It's not a version number, but rather an arbitrary ID for that version number. "1.3" is the version number and "{3,4}" is the ID.

davidben: 0x0700 is also not a signature algorithm, but we don't call them SignatureSchemeIDs. We'd have to stick "ID" after every enum in the entire document.

davegarrett: A signature algorithm is not a number. A version number is, however. The point here is that these are arbitrary identifiers, not the actual version number. Note that we do actually use the version number itself in the context strings. Calling one the version number and the other the version ID makes them explicitly separate.

martinthomson: The point of the change is to formally recognize that these are not version *numbers*, but simply *versions*.  That is, incompatible manifestations of a protocol that serve the same approximate function.  Or, in other words, version numbers failed us, so we're trying this instead.

draft-ietf-tls-tls13.md
```diff
@@ -3951,6 +3949,22 @@ by IANA
   "Recommended": ecdsa_secp256r1_sha256, ecdsa_secp384r1_sha384,
   rsa_pss_sha256, rsa_pss_sha384, rsa_pss_sha512, ed25519.
 
+-  TLS ProtocolVersionID Registry: This registry formalizes the definition
+  of values previously labeled ProtocolVersion values in {{RFC5246}}. The
+  registry [shall be/ has been] initially populated with the values defined
+  as each ProtocolVersion value applicable to past SSL and TLS specifications.
+  All values with the first byte in the range 0-2 SHALL be marked as reserved
+  for SSL 2.0, in addition to { 2, 0 }, and are not available for use.
```
martinthomson: Please reserve a portion of this space for experimentation, in the same way that we reserved space in cipher suites and extensions, etc...

draft-ietf-tls-tls13.md
davidben: +1
That's a good way to phrase it. I should have used that in the email. :-)

ekr:
I don't believe that the source of negotiation failures has in any way been that people somehow thought that the code points were in fact the versions, and this change doesn't seem to me to aid comprehension.

ekr:
Re the IANA registry: that seems like explicitly the wrong idea. Only new versions of TLS are going to be able to register here, not random new RFCs defining "blah"

davidben:
Hrm. Is there no way to add a registry that only TLS gets to extend? It seems we do want an experimental range for all the draft_version games we play. (And of course I'd like to burn some GREASE values.) Or do we just make values up since there won't be that many and 2^16 is a big number?

ekr:
Well, a registry won't help with draft_version, right?

I think we just need to specify the appropriate use of the ranges.

richsalz:
Make it an expert review registry and make the TLSWG chairs be the experts?

kaduk:
You can even supply guidance to the expert as to what criteria are to be used for approval.

davegarrett:
Ok, I've pushed a revised version (multiple commits squashed to one, here). This reverts the name change of the type to keep it consistent with past specs, but leaves in the explicit use of the term "ID" when talking about it. This also formally reserves a block of 256 for private experimental use, using the same language copy/pasted from the other registry definitions above it. The main block is set to "Standards Action", which essentially just covers new TLS version standards (or some sort of other standards track RFC the TLS WG were to publish).

draft-ietf-tls-tls13.md
```diff
@@ -1415,7 +1413,7 @@ one extension (at least "key_share" or "pre_shared_key").
 max_supported_version
 : The latest (highest valued) version of the TLS protocol offered by the
   client. This SHOULD be the same as the latest version supported. For this
-  version of the specification, the version will be { 3, 4 }. (See
+  version of the specification, the version ID will be { 3, 4 }. (See
```
martinthomson: "the version will be 0x0304 (TLS 1.3)"

No point mucking around with the struct.

draft-ietf-tls-tls13.md
```diff
@@ -3951,6 +3949,25 @@ by IANA
   "Recommended": ecdsa_secp256r1_sha256, ecdsa_secp384r1_sha384,
   rsa_pss_sha256, rsa_pss_sha384, rsa_pss_sha512, ed25519.
 
+-  TLS ProtocolVersion Registry: This registry formalizes the definition
+  of ProtocolVersion values used to identify TLS versions in TLS messages.
+  Values with the first byte in the range 0-254 (decimal) are assigned
+  via Standards Action {{RFC2434}}. Values with the first byte 255
+  (decimal) are reserved for Private Use {{RFC2434}}. Values with the
+  first byte in the range 0-2 SHALL be marked as reserved for SSL 2.0,
```
martinthomson: You can't reserve 0x0-0x2ff here when you just assigned it above.  I would just mark 0x0-0x2ff out of bounds; 0x305-0xfeff as Standards Action and the remainder as private use.

draft-ietf-tls-tls13.md
```diff
@@ -4195,7 +4212,7 @@ endpoints potentially supporting different versions of TLS.
 TLS 1.x and SSL 3.0 use compatible ClientHello messages. Servers can also handle
 clients trying to use future versions of TLS as long as the ClientHello format
 remains compatible and the client supports the highest protocol version available
-in the server.
+in the server. See {{iana-considerations}} for a full list of available version IDs.
```
martinthomson: Not needed.

draft-ietf-tls-tls13.md
davegarrett: It's not inane, it's pedantic. This is a mess, so I'm trying to be explicit as possible.

As to struct vs enum, yeah, I originally wanted to just use a uint16, but I kept things with this notation here to avoid changing too much in one PR. I'll add a second commit here that changes the notation.

davegarrett:
Rebased, though, I get the sense that we're leaning towards not doing a formal registry.

ekr:
Closing per chairs

PR #637, opened 2016-09-11
Move 'Unauthenticated Operation' to backwards compatibility section.

davidben:
Right now it's under 'Cipher Suites' which is weird with the cipher
suite split. Backwards compatibility seems a better place for it since
we're talking about features in older versions of TLS that are now gone.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4321,11 +4321,10 @@ been shown to be insecure in some scenarios.
 
 ## Unauthenticated Operation
 
-Previous versions of TLS offered explicitly unauthenticated cipher suites
-based on anonymous Diffie-Hellman. These cipher suites have been deprecated
-in TLS 1.3. However, it is still possible to negotiate cipher suites
-that do not provide verifiable server authentication by several methods,
-including:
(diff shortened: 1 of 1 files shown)


davidben:
Also noticed there were some uses of 'cipher suites' that were off, so I've replaced them with more generic words.

davegarrett:
It doesn't belong in backwards compatibility, either. It's primarily about ways to do unauthenticated in TLS 1.3; just the first sentence mentions that the old suites are gone. Not sure where this should go now, though.

davidben:
The end of "Cryptographic Negotiation", perhaps?
https://tlswg.github.io/tls13-spec/#rfc.section.4.1.1

I guess backwards compatibility here currently only talks about protocol-level backwards compatibility. There's also semantic backwards compatibility like the fact that 1.2 and 1.3 PSKs aren't the same animal.

ekr:
I agree with @davegarrett that this doesn't belong in backward compatibility. @davidben how about implementation notes?

davidben:
SGTM. Implementation notes it is!

davegarrett:
On Wednesday, September 21, 2016 02:21:32 pm David Benjamin wrote:
> SGTM. Implementation notes it is!

Ditto. Sounds like a good enough spot.

PR #638, opened 2016-09-11
Tidy up the text around AEADs and ciphers.

davidben:
We are inconsistent about "AEAD cipher" and "AEAD algorithm". RFC 5116
uses "AEAD algorithm" and "AEAD cipher" is confusing with "cipher
suite", so standardize on algorithm.

IANA considerations were never updated to actually add the cipher
suites. Fix that. Match the cipher suite name column with the IANA name
so we can reuse the table.

Finally, since we are defining new cipher suites, we should take care to
cite where our AEADs come from. Do so.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3880,18 +3880,9 @@ and their allocation policies are below:
 
   IANA [SHALL add/has added] the cipher suites listed in {{cipher-suites}} to
   the registry. The "Value" and "Description" columns are taken from the table.
-  The "DTLS-OK" column is marked as "Yes".
-
-  IANA [SHALL add/has added] a "Recommended" column
-  to the cipher suite registry. All cipher suites listed in
-  {{cipher-suites}} are marked as "Yes". All other cipher suites are
-  marked as "No". IANA [SHALL add/has added] add a note to this column
-  reading:
-
-  > Cipher suites marked as "Yes" are those allocated via
-    Standards Track RFCs. Cipher suites marked as "No" are not;
-    cipher suites marked "No" range from "good" to "bad" from a
-    cryptographic standpoint.
+  The "DTLS-OK" and "Recommended" columns are both marked as "Yes" for each new
+  cipher suite. [[This assumes draft-sandj-tls-iana-registry-updates-00 has been
+  applied.]]
 
 -  TLS ContentType Registry: Future values are allocated via
   Standards Action {{RFC2434}}.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -45,6 +45,7 @@ normative:
   RFC6367:
   RFC6655:
   RFC7251:
+  RFC7539:
   RFC7748:
   RFC7905:
   I-D.irtf-cfrg-eddsa:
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3875,7 +3876,13 @@ and their allocation policies are below:
...
   0-254 (decimal) are assigned via Specification Required {{RFC2434}}.
   Values with the first byte 255 (decimal) are reserved for Private
-  Use {{RFC2434}}. IANA [SHALL add/has added] a "Recommended" column
+  Use {{RFC2434}}.
+
+  IANA [SHALL add/has added] the cipher suites listed in {{cipher-suites}} to
+  the registry. The "Value" and "Description" columns are taken from the table.
+  The "DTLS-OK" column is marked as "Yes".
+
+  IANA [SHALL add/has added] a "Recommended" column
```
davidben: This was another consequence of the new cipher suites. We are currently poised to mark every non-TLS-1.3 cipher as not recommended. Should I leave that as-is or transcribe the list from draft 14?

ekr: No. https://tools.ietf.org/id/draft-sandj-tls-iana-registry-updates-00.txt will be taking over most of the IANA stuff.

davidben: Ah, okay. In that case, should I rephrase this to assume that text has been applied?

draft-ietf-tls-tls13.md
```diff
@@ -4023,32 +4030,35 @@ and renegotiation_info {{RFC5746}}.
...
-   CipherSuite TLS_CIPHER_HASH = VALUE;
+   CipherSuite TLS_AEAD_HASH = VALUE;
 ~~~
 
 | Component | Contents |
 |:----------|:---------|
-| TLS     | The string "TLS" |
-| CIPHER    | The symmetric cipher used for record protection |
+| TLS       | The string "TLS" |
+| AEAD      | The AEAD algorithm used for record protection |
```
davegarrett: The other changes sound good, but I think all the stuff in this section should stay as-is. A cipher suite is a suite of stuff for a cipher. There's also no real need to de-generalize it here, especially when the term "cipher suite" is well established. I think changing this just risks making it a bit more confusing.

davidben: TLS 1.3 is incapable of handling non-AEAD symmetric ciphers (which isn't really a concrete thing). An AEAD is a very particular precomposed thing with a particular interface. TLS 1.3 requires that precomposed thing now, so we should specify in terms of that.

For instance, this sentence does not reflect how any good crypto API works:
> The “CIPHER” component commonly has sub-components used to designate the cipher name, bits, and mode, if applicable. For example, “AES_256_GCM” represents 256-bit AES in the GCM mode of operation.

A good crypto library will not parametrize "give me an AES thingy" with a CBC, CTR, or GCM argument. These three all give back different things with very different properties. This is deciding AEADs and RFC 5116 is a thing.

draft-ietf-tls-tls13.md
davidben: Actually, it looks like draft-sandj-tls-iana-registry-updates-00 currently cites TLS 1.3. Is the ordering to be inverted and the list inlined into draft-sandj-tls-iana-registry-updates-00?

ekr: I think we should assume that all the non 1.3-specific updates happen there.

davidben: Rebased the text.

PR #640, opened 2016-09-17
Add certificate_required alert.

davidben:
There is no alert for a server which requested a client certificate,
required it, but was not sent one. Servers today send a generic
handshake_failure which means clients must use heuristics to show error
messages to the user.

Unlike nearly every other negotiation in TLS, client certificate
selection is actually sometimes a user-level decision, so it is arguably
the most important place for us to have a dedicated alert.

(Also remove a "This alert is always fatal" note on
bad_certificate_status_response. All alerts are now fatal, so this is
redundant.)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3570,6 +3570,8 @@ certificate_required
   the client.
 {:br }
 
+[[TODO: IANA Considerations for new alert values.]]
+
 New Alert values are assigned by IANA as described in {{iana-considerations}}.
 
 #  Cryptographic Computations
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3563,6 +3564,10 @@ unknown_psk_identity
  acceptable PSK identity is provided by the client. Sending this alert
  is OPTIONAL; servers MAY instead choose to send a "decrypt_error"
  alert to merely indicate an invalid PSK identity.
+
+certificate_required
+: Sent by servers when a client certificate is desired but none was provided by
+  the client.
```
davegarrett: Stick a "[[TODO: IANA Considerations.]]" here or amend IANA Considerations for this (and also missing_extension, possibly).

davidben: Added a TODO since there's a bunch of new alerts that all need this.

kaduk:
+1

PR #645, opened 2016-09-21
Add draft version section

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -1749,6 +1749,16 @@ The server MUST NOT send the "supported_versions" extension. The
...
 
+#### Draft Version Indicator
+
+RFC EDITOR: PLEASE REMOVE THIS SECTION
+
+While the eventual version indicator for the RFC version of TLS 1.3 will
+be {3, 4}, implementations of draft versions of this specification SHOULD
+instead advertise {0x7f, <draft-version>} in their "supported_versions" extension.
+This allows pre-RFC implementations to safely negotiate with each other,
+even if they would otherwise be incompatible.
```
davidben: Probably should clarify that this also figures into the ServerHello (and HelloRetryRequest) version.

ekr: I hadn't thought through the multi-version case. You think (I assume) that we should shove it into ServerHello.version?

davidben: Yeah. Even without multi-version, it seems sanest just to say that client and server version spellings always match. Versions are opaque u16s (I guess u8[2]s. Whatever.) which are offered by the client and selected by the server. The draft defines version {0x7f, blah} and the final will define version {3, 4}.

xiaoyinl:
By the way, you also need to update Version Negotiation section of the [wiki page](https://github.com/tlswg/tls13-spec/wiki/Implementations#version-negotiation).

davidben:
lgtm

ekr:
Done

PR #649, opened 2016-09-22
Add guidance for draft versions on the anti-downgrade mechanism.

davidben:
Folks have already gotten this wrong, so add a note here.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1582,6 +1582,13 @@ when static RSA is used.
 Note: This is an update to TLS 1.2 so in practice many TLS 1.2 clients
 and servers will not behave as specified above.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1582,6 +1582,13 @@ when static RSA is used.
 Note: This is an update to TLS 1.2 so in practice many TLS 1.2 clients
 and servers will not behave as specified above.
 
+RFC EDITOR: PLEASE REMOVE THE FOLLOWING PARAGRAPH
```
martinthomson: Note that this won't be a following paragraph, it's markdown, so this will be part of the same paragraph :)

davidben: I'd noticed that, but the one in section 1 did the same thing, so I figured you all had decided that's what "following paragraph" meant. :-)

https://tlswg.github.io/tls13-spec/#rfc.section.1

PR #652, opened 2016-09-22
Clarify g,p are predefined

xiaoyinl:
I think the current spec doesn't explicitly say what the values of g and p (for finite field DHE) are. So I want to add a link to RFC 7919 in Diffie-Hellman Parameters section to make it clear.


ekr:
Good suggestion, but I fixed this a bit differently in https://github.com/tlswg/tls13-spec/commit/76b800c5574ef12cde67bcc8b09059dcc5d1ce86

PR #654, opened 2016-09-23
Move OCSP and SCT into Certificate message

grittygrease:
This change adds a set of extensions to the Certificate message. This is
intended to hold all extension messages that are certificate-specific
as opposed to extensions that apply to the entire connection. This
change facilitates future extensions to the protocol for post-handshake
server authentication and resolves the anomaly of OCSP messages
appearing before certificates in the handshake.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2401,23 +2401,6 @@ In common network topologies for browser clients, small allowances on the
 order of ten seconds are reasonable.  Clock skew distributions are not
 symmetric, so the optimal tradeoff may involve an asymmetric replay window.
 
-### OCSP Status Extensions
-
-{{!RFC6066}} and {{!RFC6961}} provide extensions to negotiate the server
-sending OCSP responses to the client. In TLS 1.2 and below, the
-server sends an empty extension to indicate negotiation of this
-extension and the OCSP information is carried in a CertificateStatus
-message. In TLS 1.3, the server's OCSP information is
-carried in an extension in EncryptedExtensions. Specifically:
-The body of the "status_request" or "status_request_v2" extension
-from the server MUST be a CertificateStatus structure as defined
-in {{RFC6066}} and {{RFC6961}} respectively.
-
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4008,7 +4023,8 @@ is listed below:
    1.3" column with the following four values: "Client", indicating
    that the server shall not send them. "Clear", indicating
    that they shall be in the ServerHello. "Encrypted", indicating that
-   they shall be in the EncryptedExtensions block, and "No" indicating
+   they shall be in the EncryptedExtensions block, "Certificate" indicating that
+   they shall be in the Certificate block, and "No" indicating
```
kaduk: Might be worth a parenthetical that the Certificate message contents are also encrypted.

grittygrease: I'm not sure this is necessary.

draft-ietf-tls-tls13.md
```diff
@@ -2635,8 +2618,13 @@ Structure of this message:
...
 
        struct {
+           ASN1CertData cert_data;
+           Extension extensions<0..2^16-1>;
+       } SingleCertificate;
+
+       struct {
            opaque certificate_request_context<0..2^8-1>;
-           ASN1Cert certificate_list<0..2^24-1>;
+           SingleCertificate certificate_list<0..2^24-1>;
```
ekr: I think "CertificateEntry" is the convention here.

draft-ietf-tls-tls13.md
ekr: +1 to grittygrease here.

kaduk: I'm fine if we land on the "might not be worth" side.

draft-ietf-tls-tls13.md
grittygrease: Fine by me. I'll make the change and then I think it's ready to merge.

grittygrease:
Changed SingleCertificate to CertificateEntry

PR #656, opened 2016-09-23
Use TLS-style extensions in CertificateRequest.

davidben:
Differences:

- Easier to tell which matchers are and aren't defined (look in the
  registry).

- We can define other matchers that affect the same X.509 extension. If,
  say, someone needs a EKU matcher which describes an OR on the
  acceptable OIDs rather than an AND, we can simply define another one.

- If someone needs to match something that's not an extension (RSA
  public key size?), we can define it.

- The mechanism doesn't depend on X.509, so it's more usable for crazier
  things like RFC 7250. (Clllleeeaaaarrrly this means we should move
  certificate_authorities to an extension.)

- Remove MUST-level requirements on the client since they're toothless.
  As the old text already noted, a client is free to "not recognize" any
  extension anyway.

- More compact. OIDs are big.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2472,17 +2472,23 @@ Structure of this message:
 
        opaque DistinguishedName<1..2^16-1>;
 
+       enum {
+           key_usage(0),
+           extended_key_usage(1),
+           (65535)
+       } CertificateRequestExtensionType;
+
        struct {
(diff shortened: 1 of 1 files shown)


ekr:
My sense is that this is not getting consensus. David doesn't seem to feel strongly, so I am closing. @davidben please feel free to re-raise on list if you want the chairs to do a formal call for consenss

Issue #627, opened 2016-09-06
Structures with one member 


kazu-yamamoto:
What is the criteria to choose a structure with *one* member and a *non* structure? The followings are such data types defined in this spec:

    struct {
        opaque random_bytes[32];
    } Random;
    
    uint8 CipherSuite[2];    /* Cryptographic suite selector */
    
    struct {
        opaque cookie<0..2^16-1>;
    } Cookie;
    
    SignatureScheme supported_signature_algorithms<2..2^16-2>;
    
    struct {
        NamedGroup named_group_list<1..2^16-1>;
    } NamedGroupList;
    
    struct {
        Extension extensions<0..2^16-1>;
    } EncryptedExtensions;
   
    opaque DistinguishedName<1..2^16-1>;
    
    opaque ASN1Cert<1..2^24-1>;
    
    struct {
       opaque verify_data[Hash.length];
    } Finished;

If there is no reason, I would recommend to unify their syntax rules. It seems to me that the followings are much easier to read:

    opaque Random[32];
    
    uint8 CipherSuite[2];    /* Cryptographic suite selector */
    
    opaque Cookie<0..2^16-1>;
    
    SignatureScheme SupportedSignatureAlgorithms<2..2^16-2>;
    
    NamedGroup NamedGroupList<1..2^16-1>;

    Extension EncryptedExtensions<0..2^16-1>;
   
    opaque DistinguishedName<1..2^16-1>;
    
    opaque ASN1Cert<1..2^24-1>;
    
    opaque Finished[Hash.length];

ekr:
Good suggestion. Thanks.

davegarrett:
One case where a single field struct is generally the best route is for the payload struct for extensions and messages. (e.g. Cookie, NamedGroupList, EncryptedExtensions, etc.) The section is defining a self-contained data-structure, and using the struct notation consistently to do this does make sense. The ones that aren't a self-contained data-structure, yeah, I think it could be cleaned up a bit. (e.g. Random)

kazu-yamamoto:
In my opinion, if a structure with one member is used, the member should be used somewhere in definitions. For instance, "verify_data" is used as follows:

       verify_data =
           HMAC(finished_key, Hash(
                                   Handshake Context +
                                   Certificate* +
                                   CertificateVerify*
                              ) +
                              Hash(resumption_context)
           )

This makes sense to me. But "cookie" is not used in the definitions. So, I would like to use non structure for "Cookie".

[...]

PR #658, opened 2016-09-26
Fixing #627

kazu-yamamoto:
Criteria for one member structure / non structure is as follows:

Non structure: basic type -- not messages nor extensions
One member structure: others such as messages or extensions


davegarrett:
LGTM

Issue #630, opened 2016-09-07
Syntax issues


kazu-yamamoto:
I have written a syntax validator to check the validity of definitions. In this issue, I would like to point out what I noticed.

kazu-yamamoto:
(1) Some enum definitions use "floor..ceiling" notation which is not defined. Section 3.5 should clearly define this.

kazu-yamamoto:
(2) I think that the `struct` keyword is allowed only in the top level definitions. But `struct {}` is used inside of EarlyDataIndication. I would suggest to defined the following in the top level:

```
struct {} None;
```

kazu-yamamoto:
(3) The following assignment is really strange:

```
struct {
    ContentType opaque_type = application_data(23); /* see TLSInnerPlaintext.type */
```

I would suggest to change this to:

```
struct {
    ContentType opaque_type = { 23 }; /* application_data, see TLSInnerPlaintext.type */
```

davegarrett:
On Wednesday, September 07, 2016 04:24:15 am Kazu Yamamoto wrote:
> (3) The following assignment is really strange:
> 
> ```
> struct {
>     ContentType opaque_type = application_data(23); /* see TLSInnerPlaintext.type */
> ```
> 
> I would suggest to change this to:
> 
> ```
> struct {
>     ContentType opaque_type = { 23 }; /* application_data, see TLSInnerPlaintext.type */
> ```

I don't understand why you consider this current notation here strange. It's the same notation used for enum values elsewhere.

kazu-yamamoto:
What I can tell from Section 3 is as follows:

- The "symbol(num)" notation is allowed in the *declaration* of enum.
- "symbol" can be used as *label* of the case clause.
- "{num, ...}" can be used for assignment.

I don't see that "symbol(num)" can be used for assignment.

I wrote a parse based on Section 3 and what I reported here is the errors reported by the parser.

PR #659, opened 2016-09-27
Fixing #630

kazu-yamamoto:
This patches ensure that the described syntax and examples are consistent. In other words, parsers implementing Section 3 can parse all examples.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2261,13 +2261,15 @@ The "extension_data" field of this extension contains an
 
 %%% Key Exchange Messages
 
+       struct {} None;
+
        struct {
            select (Handshake.msg_type) {
                case client_hello:
                    uint32 obfuscated_ticket_age;
 
                case server_hello:
-                   struct {};
+                   None;
            };
        } EarlyDataIndication;
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1173,6 +1173,7 @@ numerical information may be omitted.
 
        enum { low, medium, high } Amount;
 
+For numerical values, the notation \<floor..ceiling\> is also allowed.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2261,13 +2261,15 @@ The "extension_data" field of this extension contains an
...
+       struct {} None;
+
        struct {
            select (Handshake.msg_type) {
                case client_hello:
                    uint32 obfuscated_ticket_age;
 
                case server_hello:
-                   struct {};
+                   None;
```
davegarrett: If we used this notation in more places, maybe, however just doing things like this feels like excessive indirection. A simpler simplification would be to just drop the word "struct" and state the payload as "{}". I don't feel like this is a particularly useful thing to fiddle with, however. It's fine as-is.

draft-ietf-tls-tls13.md
```diff
@@ -1173,6 +1173,7 @@ numerical information may be omitted.
 
        enum { low, medium, high } Amount;
 
+For numerical values, the notation \<floor..ceiling\> is also allowed.
```
davegarrett: For enums we use parens rather than angle brackets, so this would be "(floor..ceiling)". (we've generally used angle brackets for size ranges and parens for value ranges) With this tweak, I think this is a logical clarification to add.

kazu-yamamoto: Sorry. This is my mistake and you are right.

draft-ietf-tls-tls13.md
kazu-yamamoto: But "{}" is not defined in Section 3. In my opinion, such an ill-defined notation is much confusing than this indirection. I would vote for `None` (or `Empty`) even if it is redundant.

ekr:
I think I prefer it as-is.

PR #665, opened 2016-10-03
Added language to Extensions for session resumption

kmad:
Added to the ClientHello Extensions information from section 2.2 to cover session resumption mechanisms.


ekr:
I'd prefer not to duplicate this guidance, and I tend to think it's better where it is.

PR #666, opened 2016-10-04
Make a cookie have a minimum size

martinthomson:
The spec requires that a cookie (or key_share) be present in HRR, but it's possible to send an empty cookie.  That's just inviting errors.


davegarrett:
I was under the impression that we needed to be capable of having a client include an empty cookie extension so that a server could reply with a cookie, as server's aren't allowed to use extensions without them first being offered by the client.

martinthomson:
The cookie remains the exception to the rule.  Option 2 is to force the client to send a cookie extension.

PR #668, opened 2016-10-05
Allowing Zero-RTT with PSK

hannestschofenig:
I am proposing that a client can also use the Zero-RTT with PSK-based authentication without having to rely on a previously obtained ticket.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -967,7 +967,7 @@ establishment, thus providing forward secrecy.
 
 ## Zero-RTT Data
 
-With the Zero-RTT mode clients can send data on their first flight ("early data") whereby the client uses a PSK either obtained out-of-band or as a ticket (i.e., one with the “early_data_info” extension) from an earlier exchange to authenticate to the server. This early data is encrypted solely under keys derived using the offered PSK as the static secret. 
+With the Zero-RTT mode clients can send data on their first flight ("early data") whereby the client uses a PSK either obtained out-of-band or as a ticket (i.e., one with the "early_data_info" extension) from an earlier exchange to authenticate to the server. This early data is encrypted solely under keys derived using the offered PSK as the static secret. 
 
 As shown in {{tls-0-rtt}}, the Zero-RTT data is just added to the 1-RTT 
 handshake in the first flight. The rest of the handshake uses the same messages
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -967,12 +967,10 @@ establishment, thus providing forward secrecy.
 
 ## Zero-RTT Data
 
-When resuming via a PSK with an appropriate ticket (i.e., one with
-the "early_data_info" extension), clients can also send data on their first
-flight ("early data"). This data is encrypted solely under keys
-derived using the first offered PSK as the static secret.  As shown in
-{{tls-0-rtt}}, the Zero-RTT data is just added to the 1-RTT handshake
-in the first flight. The rest of the handshake uses the same messages
+With the Zero-RTT mode clients can send data on their first flight ("early data") whereby the client uses a PSK either obtained out-of-band or as a ticket (i.e., one with the “early_data_info” extension) from an earlier exchange to authenticate to the server. This early data is encrypted solely under keys derived using the offered PSK as the static secret. 
+
+As shown in {{tls-0-rtt}}, the Zero-RTT data is just added to the 1-RTT 
+handshake in the first flight. The rest of the handshake uses the same messages
 as with a 1-RTT handshake with PSK resumption.
 
 ~~~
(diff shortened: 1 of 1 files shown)


kaduk:
Surely there would need to be additional security considerations for this behavior?!

davegarrett:
Could you please add linebreaks to wrap to about 80 character or so lines? Our line length is often rather inconsistent, but we do generally try to keep lines shorter than in the current changeset.

ekr:
This LGTM. I'll let it sit for a little but in case anyone wants to weigh in (@mt, @davegarrett, @davidben)

draft-ietf-tls-tls13.md
```diff
@@ -967,12 +967,21 @@ establishment, thus providing forward secrecy.
...
+out-of-band or as a ticket (i.e., one with the "early_data_info" 
+extension) from an earlier exchange to authenticate to the server. 
+
+When clients use a PSK obtained out-of-band then the following 
+information MUST be provisioned to both parties: 
+  * PSK identity, 
+  * ciphersuite for use with this PSK, 
+  * key exchange and authentication modes this PSK is allowed to be used with,  
+  * the Application-Layer Protocol Negotiation (ALPN) label(s), and 
+  * optionally the Server Name Indication (SNI).
```
kaduk: MUST and optionally; two great tastes that go together.

draft-ietf-tls-tls13.md
```diff
@@ -967,12 +967,21 @@ establishment, thus providing forward secrecy.
 
 ## Zero-RTT Data
 
-When resuming via a PSK with an appropriate ticket (i.e., one with
-the "early_data_info" extension), clients can also send data on their first
-flight ("early data"). This data is encrypted solely under keys
-derived using the first offered PSK as the static secret.  As shown in
-{{tls-0-rtt}}, the Zero-RTT data is just added to the 1-RTT handshake
-in the first flight. The rest of the handshake uses the same messages
+With the Zero-RTT mode clients can send data on their first flight
```
davegarrett: nit: should be a comma after "mode"

draft-ietf-tls-tls13.md
```diff
@@ -967,12 +967,21 @@ establishment, thus providing forward secrecy.
...
+("early data") whereby the client uses a PSK either obtained 
+out-of-band or as a ticket (i.e., one with the "early_data_info" 
+extension) from an earlier exchange to authenticate to the server. 
+
+When clients use a PSK obtained out-of-band then the following 
+information MUST be provisioned to both parties: 
+  * PSK identity, 
+  * ciphersuite for use with this PSK,
```
davegarrett: cipher suite should be two words

draft-ietf-tls-tls13.md
```diff
@@ -1018,8 +1027,9 @@ https://github.com/tlswg/tls13-spec/issues/443]]
 IMPORTANT NOTE: The security properties for 0-RTT data are weaker than
 those for other kinds of TLS data.  Specifically:
 
-1. This data is not forward secret, because it is encrypted solely
-with the PSK.
+1. The early data is encrypted solely under keys derived using the 
+offered PSK as the static secret. Hence, this data is not forward 
+secret, because it is encrypted solely with the PSK.
```
davegarrett: This repeats the same thought twice. ("solely" twice with different wording) If we want to add the wording of the new sentence here, then a suggested replacement:

"
1. This data is not forward secret, as it is encrypted solely under keys derived using the offered PSK as the static secret.
"

draft-ietf-tls-tls13.md
martinthomson: I suspect that the SNI line needs to be "the Server Name Indication (SNI), if any is to be used"

draft-ietf-tls-tls13.md
```diff
@@ -967,8 +967,19 @@ establishment, thus providing forward secrecy.
...
-With the Zero-RTT mode clients can send data on their first flight ("early data") whereby the client uses a PSK either obtained out-of-band or as a ticket (i.e., one with the "early_data_info" extension) from an earlier exchange to authenticate to the server. This early data is encrypted solely under keys derived using the offered PSK as the static secret. 
-
+With the Zero-RTT mode clients can send data on their first flight
+("early data") whereby the client uses a PSK either obtained 
+out-of-band or as a ticket (i.e., one with the "early_data_info" 
+extension) from an earlier exchange to authenticate to the server. 
+
+When clients use a PSK obtained out-of-band then the following 
+information MUST be provisioned to both parties: 
+  * PSK identity,
```
martinthomson: You also have to provision the key.  So maybe, "In addition to the pre-shared key, ..." or "In order to use 0-RTT with a pre-shared key that is obtained out-of-band, the following information MUST be associated with the key by both parties:"

martinthomson: Note here that there is an important consideration: this only applies to the use of PSK for 0-RTT.  A PSK for 1-RTT only needs the key, the identity and the PRF hash function since the other properties can be negotiated.  I would add that as a note following this list.

PR #670, opened 2016-10-05
Early ems

ekr:
@mt PTAL

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -428,6 +428,12 @@ server: The endpoint which did not initiate the TLS connection.
 (*) indicates changes to the wire protocol which may require implementations
     to update.
 
+draft-17
+
+- Add a 0-RTT exporter and change the transcript for the regular exporter (*).
+
+- Explicitly allow predicting ClientFinished for NST.
+
 draft-16
 
 - Revise version negotiation (*)
@@ -1033,6 +1039,8 @@ protocol. However, 0-RTT data cannot be duplicated within a connection (i.e., th
 will not process the same data twice for the same connection) and
 an attacker will not be able to make 0-RTT data appear to be
 1-RTT data (because it is protected with different keys.)
+
+The same warnings apply to any use of the early exporter secret.
 
 The remainder of this document provides a detailed description of TLS.
 
@@ -2970,6 +2978,12 @@ OPEN ISSUE https://github.com/tlswg/tls13-spec/issues/558]]
 Any ticket MUST only be resumed with a cipher suite that is identical
 to that negotiated connection where the ticket was established.
 
+Note: Although the resumption_psk depends on the client's second
+flight, servers which do not request client authentication MAY compute
+the remainder of the transcript independently and then send a
+NewSessionTicket immediately upon sending its Finished rather than
+waiting for the client Finished.
+
 
 %%% Ticket Establishment
 
@@ -3783,7 +3797,7 @@ operation.
                  |                         = server_traffic_secret_0
                  |
                  +---------> Derive-Secret(., "exporter master secret",
-                 |                         ClientHello...Client Finished)
+                 |                         ClientHello...Server Finished)
                  |                         = exporter_secret
                  |
                  +---------> Derive-Secret(., "resumption master secret",
@@ -3915,18 +3929,19 @@ the TLS PRF. This document replaces the PRF with HKDF, thus requiring
 a new construction. The exporter interface remains the same. If context is
 provided, the value is computed as:
 
-    HKDF-Expand-Label(exporter_secret, label, context_value, key_length)
+    HKDF-Expand-Label(Secret, label, context_value, key_length)
+
+Where Secret is either the early_exporter_secret or the exporter_secret.
 
 If no context is provided, the value is computed as:
 
-    HKDF-Expand-Label(exporter_secret, label, "", key_length)
+    HKDF-Expand-Label(Secret, label, "", key_length)
 
 Note that providing no context computes the same value as providing an empty
 context. As of this document's publication, no allocated exporter label is used
 with both modes. Future specifications MUST NOT provide an empty context and no
 context with the same label and SHOULD provide a context, possibly empty, in
 all exporter computations.
-
 
 
 #  Compliance Requirements
@@ -4564,7 +4579,14 @@ independent, so it is not feasible to compute one from another or
 the session secret from the exported value. Note: exporters can
 produce arbitrary-length values. If exporters are to be
 used as channel bindings, the exported value MUST be large
-enough to provide collision resistance.
+enough to provide collision resistance. The exporters provided
+TLS 1.3 are derived from the same handshake contexts as the
+early traffic keys and the application traffic keys respectively,
+and thus have similar security properties. Note that they do
+not include the client's certificate; future applications
+which wish to bind to the client's certificate may need
+to define a new exporter that includes the full handshake
+transcript.
 
 For all handshake modes, the Finished MAC (and where present, the
 signature), prevents downgrade attacks. In addition, the use of
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -440,7 +440,8 @@ draft-16
   as an empty context.(*)
 
 - New KeyUpdate format that allows for requesting/not-requesting an
-  answer (*)
+  answer. This also means changes to the key schedule to support
+  independent updates (*)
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1346,11 +1357,10 @@ following four sets of options in its ClientHello:
   (EC)DHE shares for some or all of these groups.
 - A "signature_algorithms" ({{signature-algorithms}}) extension which indicates the signature
   algorithms which the client can accept.
-- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains the identities
-  of symmetric keys known to the client and the key exchange
-  modes which each PSK supports.
+- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains a symmetric
+  key known to the client and the key exchange  modes which it supports.
```
martinthomson: nit, extra space after exchange

draft-ietf-tls-tls13.md
```diff
@@ -809,7 +816,7 @@ contains a "key_share" extension with the server's ephemeral
 Diffie-Hellman share which MUST be in the same group as one of the
 client's shares. If PSK key establishment is
 in use, then the ServerHello contains a "pre_shared_key"
-extension indicating which of the client's offered PSKs was selected.
+extension.
```
martinthomson: you need to mention the binder on line 806 (github doesn't let me comment there)

draft-ietf-tls-tls13.md
```diff
@@ -1406,6 +1416,10 @@ ClientHello (without modification) except:
 
 - Including a "cookie" extension if one was provided in the
   HelloRetryRequest.
+
+- Replacing the "pre_shared_key" and "psk_binder" extensions with
+  the one indicated in the HelloRetryRequest if indicated by
+  a "pre_shared_key" entry in the HelloRetryRequest.
```
martinthomson: I assume that this is provisional and you would prefer to go back to multiple PSKs.

draft-ietf-tls-tls13.md
```diff
@@ -1655,11 +1669,13 @@ this specification are:
...
 - key_share (see {{key-share}})
 
+- pre_shared_key (see {{pre-shared-key}})
+
 Note that HelloRetryRequest extensions are defined such that the original
 ClientHello may be computed from the new one, given minimal state about which
 HelloRetryRequest extensions were sent. For example, the key_share extension
 causes the new KeyShareEntry to be appended to the client_shares field, rather
-than replacing it.
+than replacing it.
```
martinthomson: trailing space

draft-ietf-tls-tls13.md
```diff
@@ -1655,11 +1669,13 @@ this specification are:
 
 - key_share (see {{key-share}})
 
+- pre_shared_key (see {{pre-shared-key}})
+
```
martinthomson: also provisional

draft-ietf-tls-tls13.md
```diff
@@ -2169,34 +2186,39 @@ The "extension_data" field of this extension contains a
...
            };
        } PreSharedKeyExtension;
 
-identities
-: A list of the identities (labels for keys) that the client is willing
-  to negotiate with the server. If sent alongside the "early_data"
-  extension (see {{early-data-indication}}), the first identity is the
-  one used for 0-RTT data.
+identity
+: An identity (labels for a key) that the client is willing
```
martinthomson: s/labels/a label

draft-ietf-tls-tls13.md
```diff
@@ -2230,24 +2252,30 @@ mode, the server MUST send Certificate ({{certificate}}) and CertificateVerify
 {:/comment}
 {:br}
 
-In order to accept PSK key establishment, the server sends a
-"pre_shared_key" extension with the selected identity.
+Each PSK MUST be associated with a single Hash algorithm. For PSKs established
```
martinthomson: Nit: I think that the "MUST be" here can be simply "is" and a MUST can be added to "For externally established PSKs, the Hash algorithm MUST be set when the PSK is established."

draft-ietf-tls-tls13.md
```diff
@@ -2314,8 +2341,7 @@ can behave in one of two ways:
   its followup ClientHello.
 
 In order to accept early data, the server server MUST have accepted a
-PSK cipher suite and selected the the first key offered in the
-client's "pre_shared_key" extension. In addition, it MUST verify that
+PSK cipher suite. In addition, it MUST verify that
```
martinthomson: Any point in reiterating that the psk_binder needs to be checked?

draft-ietf-tls-tls13.md
```diff
@@ -2419,7 +2445,44 @@ the existing behavior for SignedCertificateTimestamps {{?RFC6962}},
...
-## Server Parameters Messages
+### PSK Binder
+
+%%% Authentication Messages
+       struct {
+           opaque binding_data[Hash.length];
+       } PskBinder;
+
+The "psk_binder" extension is used to carry a binding between the PSK
+and the current handshake. as well as between the session where the
```
martinthomson: errant period here

draft-ietf-tls-tls13.md
```diff
@@ -2419,7 +2445,44 @@ the existing behavior for SignedCertificateTimestamps {{?RFC6962}},
...
+alert.
+
+PskBinder.binding_data is computed using the same way as the
+Finished.verify data value, using the PSK as the base key and the
+portion of the handshake transcript up to but not including
+PskBinder.binding_data as the Handshake Context. If the handshake
+includes a HelloRetryRequest, the initial ClientHello and
+HelloRetryRequest are included in the transcript.  The actual
+ClientHello, complete with the full "psk_binder" extension is included
+in all other handshake hash computations.
```
martinthomson: This needs to be even more clear, because the risk of getting this wrong is very high.

You need to make it clear that this includes the ClientHello that the "psk_binder" extension is carried.  And that the ClientHello length fields are constructed as though the "psk_binder" extension were present.  And that the entire extension including identifier and length is omitted from the hash.

I think that an example, or illustration would help.  Making the same point several different ways will avoid mistakes and misinterpretations.

For the same reason, you should include a formula for the calculation:

    HMAC(Hash, PSK, [ClientHello + HelloRetryRequest +] Truncated ClientHello)

draft-ietf-tls-tls13.md
```diff
@@ -4557,7 +4612,14 @@ independent, so it is not feasible to compute one from another or
...
 used as channel bindings, the exported value MUST be large
-enough to provide collision resistance.
+enough to provide collision resistance. The exporters provided
+TLS 1.3 are derived from the same handshake contexts as the
+early traffic keys and the application traffic keys respectively,
+and thus have similar security properties. Note that they do
+not include the client's certificate; future applications
+which wish to bind to the client's certificate may need
+to define a new exporter that includes the full handshake
+transcript.
```
martinthomson: You are turning this into a wall of text.  There are several key pieces of information here: exporters produce unique secrets (1); exporters with different labels/contexts produce independent values (1); exporters can produce arbitrary output, but large values are needed to provide collision resistance which is important for channel bindings (3); the early exporter shares all the limitations of 0-RTT keys (4); exporters don't cover the client's certificate, only the client's fresh entropy (2).  I would arrange that as I have noted in parentheses.

Also, this section needs sub-sections.  It's a lot of information to wade through.

draft-ietf-tls-tls13.md
hannestschofenig: Would it be OK to replace the following sentence 

--
The "psk_binder" extension is used to carry a binding between the PSK and the current handshake. as well as between the session where the PSK was established and the session where it was used.
-- 
with 
-- 
The "psk_binder" extension carries the output of an HMAC-based digest computed over the portion of the handshake transcript up to but not including PskBinder.binding_data as the Handshake Context. As such, the PskBinder.binding_data is computed using the same way as the Finished.verify data value, using the PSK as the base key and. If the handshake includes a HelloRetryRequest, the initial ClientHello and HelloRetryRequest are included in the transcript.  The actual ClientHello, complete with the full "psk_binder" extension is included in all other handshake hash computations.
-- 

I would move the text that relates more to the security properties to the end of the section, such as this text: 
"
This requires that both the KDF used to produce the RMS
 +and the MAC used to compute the psk_binder be collision
 +resistant. These are properties of HKDF and HMAC respectively when
 +used with collision resistant hash functions and producing output of
 +at least 256 bits.  Any future replacement of these functions MUST
 +also provide collision resistance.
"

PR #673, opened 2016-10-07
Early ems2

ekr:
Various changes to make exporters/resumption work better. The rationale is to be able to do stuff in 0/0.5RTT.

- Add a 0-RTT exporter and change the transcript for the regular exporter (*).

- Explicitly allow predicting ClientFinished for NST when not doing
  Client auth.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3932,6 +3932,9 @@ provided, the value is computed as:
     HKDF-Expand-Label(Secret, label, context_value, key_length)
 
 Where Secret is either the early_exporter_secret or the exporter_secret.
+Implementations MUST use the exporter_secret unless explicitly specified
+by the application. When adding TLS 1.3 to TLS 1.2 stacks, the exporter_secret
+MUST be for the existing exporter interface.
 
 If no context is provided, the value is computed as:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4557,7 +4579,14 @@ independent, so it is not feasible to compute one from another or
 the session secret from the exported value. Note: exporters can
 produce arbitrary-length values. If exporters are to be
 used as channel bindings, the exported value MUST be large
-enough to provide collision resistance.
+enough to provide collision resistance. The exporters provided
```
colmmacc: are you missing "in", or "with" after "provided" here?

draft-ietf-tls-tls13.md
```diff
@@ -3908,11 +3929,13 @@ the TLS PRF. This document replaces the PRF with HKDF, thus requiring
 a new construction. The exporter interface remains the same. If context is
 provided, the value is computed as:
 
-    HKDF-Expand-Label(exporter_secret, label, context_value, key_length)
+    HKDF-Expand-Label(Secret, label, context_value, key_length)
+
+Where Secret is either the early_exporter_secret or the exporter_secret.
```
nharper: Can this be expanded to say when early_exporter_secret should be used vs exporter_secret? E.g. something like exporter_secret must be used for providing the RFC 5705 interface?

PR #675, opened 2016-10-07
update MTI Extensions section for new negotiation

davegarrett:
The MTI Extensions section is currently out-of-date. It wasn't updated with the new negotiation and still refers to the old cipher suites system. This PR cleans that up.

I also drop the cookie extension universal send requirement, as that was in there as an alternative to allowing a server to send a cookie without first being prompted. We have that in as an explicit exception, currently.


ekr:
I already did some of this in PR#672, so I'm going to hold this PR until that has landed and then we can take stock

davegarrett:
Sure, I hadn't read that PR yet.

A notable difference in the language here is that I'm replacing the "MAY abort the connection" line with specific requirements, as that high level of generalization is counterproductive now that cipher suites and feature extensions are decoupled.

PR #676, opened 2016-10-08
Make all post-handshake messages optional

grittygrease:
This change adds a new extension to advertise support for different post-handshake messages.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1680,6 +1680,7 @@ The extension format is:
            early_data(42),
            supported_versions(43),
            cookie(44),
+           post_handshake(45),
            (65535)
        } ExtensionType;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2942,22 +2943,47 @@ Any records following a 1-RTT Finished message MUST be encrypted under the
...
+~~~~
+    enum {
+        new_session_ticket(0),
+        key_update(1),
+        client_auth_request(2),
+        (255)
+    } MessageType;
+
+    struct {
+        MessageType message_types<0..2^8-1>;
```
kaduk: If it's going to be "which includes at least one MessageType", this should be 1..2^8-1, not 0...

draft-ietf-tls-tls13.md
```diff
@@ -2942,22 +2943,47 @@ Any records following a 1-RTT Finished message MUST be encrypted under the
...
+    } PostHandshake;
+~~~~
+
+The extension data for the post_handshake extension is a PostHandshake structure,
+which includes at least one MessageType. The client includes a
+post_handshake extension containing every type of post-handshake message
+it supports in its ClientHello. If the server supports any post-handshake
+message types in common with the client, it replies with an
+EncryptedExtensions containing a post_handshake extension containing a list of
```
kaduk: The decision on whether to send EncryptedExtensions is not just based on PostHandshake; "it includes a post_handshake extension in the EncryptedExtensions it returns to the client.  The returned post_hanshake extension contains" might be a better wording.

draft-ietf-tls-tls13.md
```diff
@@ -2942,22 +2943,47 @@ Any records following a 1-RTT Finished message MUST be encrypted under the
...
+message types that it supports. The set of MessageTypes in the server’s
```
kaduk: Given the next sentence, maybe no change is needed here, but it could be "supports in common with the client" or similar.

draft-ietf-tls-tls13.md
```diff
@@ -3056,11 +3086,16 @@ ticket_age_add
   the client.
 {:br }
 
+Implementations that receive a NewSessionTicket without having advertised
```
kaduk: Thinking about consistency amongst the various paragraphis similar to this one; should this one specify clients as receiving NST?

draft-ietf-tls-tls13.md
```diff
@@ -3128,6 +3168,10 @@ a KeyUpdate with the old key is received before accepting any messages
 encrypted with the new key. Failure to do so may allow message truncation
 attacks.
 
+Implementations that receive a KeyUpdate without having advertised
+key_update in the post_handshake extension MUST terminate the
```
kaduk: This seems to allow the scenario where a client advertises KeyUpdate support, a server does not include KeyUpdate in its EncryptedExtensions, but then the server goes and sends a KeyUpdate anyway.
That's a pretty silly scenario, but is an asymmetry that could be smoothed over.

ekr:
Closing this per offline discussion. See PR #680 for the current thinking on alternate approaches.

PR #678, opened 2016-10-11
Issue644 remove hrr reconstruction

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -1645,12 +1647,13 @@ HelloRetryRequest MUST NOT contain any extensions that were not first
...
-"unexpected_message" alert in
-response to any second HelloRetryRequest which was sent in the same connection
-(i.e., where the ClientHello was itself in response to a HelloRetryRequest).
+Upon receipt of a HelloRetryRequest, the client MUST verify that the
+extensions block is not empty and otherwise MUST abort the handshake
+with a "decode_error" alert. Clients SHOULD abort the handshake with
+an "unexpected message" if the HelloRetryRequest would not result in
+any change in the ClientHello or in response to any second
+HelloRetryRequest which was sent in the same connection (i.e., where
+the ClientHello was itself in response to a HelloRetryRequest).
```
martinthomson: So the thing to note here is that HRR can't include extensions that you don't agree should be there.  So, if extensions are present, then they have to be ones that change the ClientHello (we shouldn't define any that don't).  That means the only way for HRR to be wrong is:

- no extensions (decode_error)
- unknown extension (unsupported_extension)
- extension doesn't cause a change, for instance key_share uses an unknown group, or one that already had a share

In the last case, I chose to make this an illegal_parameter alert.

And yes, if you receive a second HRR, then unexpected_message is correct (note that you missed the underscore there).

I chose to send illegal_parameter

davegarrett:
On Tuesday, October 11, 2016 08:14:32 pm Martin Thomson wrote:
> - no extensions (decode_error)

or missing_extension

martinthomson:
I think that missing_extension would be OK, but we would hit the lower-level decode_error first, since HRR is defined with:
```
       Extension extensions<2..2^16-1>;
```

PR #680, opened 2016-10-11
Forbid post-handshake client authentication except when permitted by

ekr:
application profile.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3073,10 +3073,15 @@ max_early_data_size
 
 The server is permitted to request client authentication at any time
 after the handshake has completed by sending a CertificateRequest
-message. The client SHOULD respond with the appropriate Authentication
-messages. If the client chooses to authenticate, it MUST send
-Certificate, CertificateVerify, and Finished. If it declines, it
-MUST send a Certificate message containing no certificates followed by Finished.
+message. Servers MUST NOT send this message unless explicitly allowed
+by an application profile. Clients who receive a disallowed CertificateRequest
+SHOULD terminate the connection.
+
+When allowed by application profile, the client SHOULD respond with
+the appropriate Authentication messages. If the client chooses to
+authenticate, it MUST send Certificate, CertificateVerify, and
+Finished. If it declines, it MUST send a Certificate message
+containing no certificates followed by Finished.
 
 Note: Because client authentication may require prompting the user,
 servers MUST be prepared for some delay, including receiving an


draft-ietf-tls-tls13.md
```diff
@@ -3073,10 +3073,15 @@ max_early_data_size
 
 The server is permitted to request client authentication at any time
```
kaduk: I wonder if we want s/is permitted to/can/

draft-ietf-tls-tls13.md
```diff
@@ -3073,10 +3073,15 @@ max_early_data_size
...
 after the handshake has completed by sending a CertificateRequest
-message. The client SHOULD respond with the appropriate Authentication
-messages. If the client chooses to authenticate, it MUST send
-Certificate, CertificateVerify, and Finished. If it declines, it
-MUST send a Certificate message containing no certificates followed by Finished.
+message. Servers MUST NOT send this message unless explicitly allowed
+by an application profile. Clients who receive a disallowed CertificateRequest
+SHOULD terminate the connection.
```
kaduk: unexpected_message?

ekr: Yes, but I think it's clear from the general rules

draft-ietf-tls-tls13.md
davidben: I'd probably also drop "at any time" since that kind of conflicts with the MUST NOT.

draft-ietf-tls-tls13.md
```diff
@@ -3073,10 +3073,15 @@ max_early_data_size
 
 The server is permitted to request client authentication at any time
 after the handshake has completed by sending a CertificateRequest
-message. The client SHOULD respond with the appropriate Authentication
-messages. If the client chooses to authenticate, it MUST send
-Certificate, CertificateVerify, and Finished. If it declines, it
-MUST send a Certificate message containing no certificates followed by Finished.
+message. Servers MUST NOT send this message unless explicitly allowed
+by an application profile. Clients who receive a disallowed CertificateRequest
```
davidben: Do we want some text about the request context too? That too inherently has meaning specific to the application profile. (Alternatively, do we even need it? The old renego-based hack didn't use it.)

ekr:
Yes, you need the request context. It takes the place of the server nonce
for anti-replay purposes

davidben:
Ah! Should there be an explicit random field then? It seems not immediately obvious that if I'm writing how to use this in my protocol that I need entropy in there and not just identifying information.

ekr:
No, you just need it to be fresh/unique which is what it requires

grittygrease:
Can we also be explicit that any other unexpected post-handshake message is fatal to the connection?

martinthomson:
@grittygrease that is probably OK.  Generally speaking, unexpected handshake messages are treated as fatal by implementations (NSS sends a fatal alert I believe).

ekr:
There didn't seem to be consensus to make this change, so I'm parking it.

ekr:
Changed in -20

PR #681, opened 2016-10-12
Clarify interaction of ALPN with 0-RTT data.

knekritz:
Currently the draft specifies that the ALPN must be "the same" as in the connection that established the PSK used with 0-RTT, and that the server must check that the selected ALPN matches what was previously used. I find this unclear if
1) the client should select and offer one (and only one) application protocol
2) the client can offer multiple protocols, but use the most preferred one offered for 0-RTT data
3) the client must send the exact same ALPN extension as in the previous connection, but must use the ALPN previously selected by the server (even if it was not the client's first offer).


draft-ietf-tls-tls13.md
```diff
@@ -2348,17 +2354,17 @@ can behave in one of two ways:
...
-  HelloRetryRequest.  A client MUST NOT include the "early_data" extension in
+  HelloRetryRequest. A client MUST NOT include the "early_data" extension in
   its followup ClientHello.
 
-In order to accept early data, the server server MUST have accepted a
-PSK cipher suite and selected the the first key offered in the
-client's "pre_shared_key" extension. In addition, it MUST verify that
-the following values are consistent with those negotiated in the
-connection during which the ticket was established.
+In order to accept early data, the server MUST have accepted a PSK cipher suite
```
davegarrett: There's no such thing as a "PSK cipher suite" anymore. Please update this text while you're here. I guess just drop this part and make it "...server MUST select the first key offered..."

draft-ietf-tls-tls13.md
```diff
@@ -2348,17 +2354,17 @@ can behave in one of two ways:
...
+and selected the the first key offered in the client's "pre_shared_key"
```
davegarrett: the the

draft-ietf-tls-tls13.md
```diff
@@ -2514,9 +2520,9 @@ certificate_request_context
   which will be echoed in the client's Certificate message. The
   certificate_request_context MUST be unique within the scope
   of this connection (thus preventing replay of client
-  CertificateVerify messages).   This field SHALL be zero length 
-  unless used for the post-handshake authentication exchanges 
-  described in Section 4.5.2. 
+  CertificateVerify messages).   This field SHALL be zero length
+  unless used for the post-handshake authentication exchanges
+  described in Section 4.5.2.
```
davegarrett: Why is this referring to a section by number? Fix this with a section reference.

draft-ietf-tls-tls13.md
```diff
@@ -990,26 +990,32 @@ establishment, thus providing forward secrecy.
...
 
-When clients use a PSK obtained out-of-band then the following 
+When clients use a PSK obtained out-of-band then the following
 information MUST be provisioned to both parties:
 
   * The PSK identity
   * The cipher suite for use with this PSK
   * The key exchange and authentication modes this PSK is allowed to be used with
-  * The Application-Layer Protocol Negotiation (ALPN) label(s)
+  * The Application-Layer Protocol Negotiation (ALPN) labels supported, if any
```
kaduk: Not your change, but I wonder whether pure PSK should be limited to a single ALPN (for 0RTT).

martinthomson: My view is that external PSK (I think that's the term that's been used in the draft) has one set of requirements for 1-RTT with a stricter set of requirements if it is to be used for 0-RTT.

1-RTT only needs identity and hash; whereas 0-RTT needs identity, cipher suite (instead of hash), ALPN (if any), key exchange modes, and SNI (if any).  In 1-RTT, most of those parameters are negotiated; they don't need to be fixed.  (They could be, of course.)

ekr:
My sense is that there's not consensus for this change. I'll defer to the chairs if they feel differently.

knekritz:
I'm fine with the current intended behavior, I still think it could be clarified what the intended behavior is, I'll put up a new PR to clarify that.

PR #683, opened 2016-10-12
Make RFC7919 a normative reference

xiaoyinl:
According to the IESG statement [1]:
> Normative references specify documents that must be read to understand or implement the technology in the new RFC.
> Even references that are relevant only for optional features must be classified as normative if they meet the above conditions for normative references. 

So I think it is better to categorize RFC 7919 as normative rather than informative, because it is a must-read to implement [section 4.2.5.1](https://tlswg.github.io/tls13-spec/#ffdhe-param) Diffie-Hellman Parameters.

[1] https://www.ietf.org/iesg/statement/normative-informative.html

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -48,6 +48,7 @@ normative:
   RFC7539:
   RFC7748:
   RFC7905:
+  RFC7919:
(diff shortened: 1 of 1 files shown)


ekr:
Merged separately

PR #684, opened 2016-10-12
Move ticket age

ekr:
This sits on top of PSK binder.

@mt, @davidben PTAL

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1391,9 +1391,11 @@ following four sets of options in its ClientHello:
   (EC)DHE shares for some or all of these groups.
 - A "signature_algorithms" ({{signature-algorithms}}) extension which indicates the signature
   algorithms which the client can accept.
-- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which contains a symmetric
-  keys known to the client and a "psk_key_exchange_modes" ({{pre-shared-key-exchange-modes}}) extension which indicates
-  the key exchange modes that may be used with PSKs.
+- A "pre_shared_key" ({{pre-shared-key-extension}}) extension which
+  contains a a list of symmetric keys known to the client and a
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2358,28 +2373,13 @@ The "extension_data" field of this extension contains an
...
-  {{NewSessionTicket}}.  This addition prevents passive observers from
-  correlating sessions unless tickets are reused.  Note: because ticket
-  lifetimes are restricted to a week, 32 bits is enough to represent any
-  plausible age, even in milliseconds.
-
-A server MUST validate that the ticket_age is within a small
-tolerance of the time since the ticket was issued (see {{replay-time}}).
-If it is not, the server SHOULD proceed with the handshake but reject
-0-RTT.
+A server MUST validate that the ticket_age for the selected PSK
```
kaduk: There is no "ticket_age" right now; you have to say how to compute it.

draft-ietf-tls-tls13.md
```diff
@@ -3116,14 +3125,7 @@ This document defines one ticket extension, "ticket_early_data_info"
        } TicketEarlyDataInfo;
 
 This extension indicates that the ticket may be used to send 0-RTT data
-({{early-data-indication}})). It contains the following values:
-
-
-ticket_age_add
```
kaduk: You didn't actually remove ticket_age_add from the struct on line 3114 (LHS).

ekr:
> I know we want to have something in the ticket extensions to exercise it, but the case for max_early_data_size vs ticket_age_add is not overwhelming ... the age can be some input into whether the CH is fresh but max_early_data_size only helps if you accept early data, I suppose ... is there anything else?

Yes. max_early_data_size is an extension because you don't need it if you aren't doing 0-RTT. ticket_age_add is (in this PR) not an extension because you need it with every ticket.

ekr:
Rebased and merged by hand

PR #685, opened 2016-10-12
Remove unused references

xiaoyinl:
RFC0793, RFC1948, RFC4302, RFC4303 and RFC5763 are not referenced by any text in the body of the current draft. So remove them from the reference section.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -88,12 +88,8 @@ normative:
 
 
 informative:
-  RFC0793:
-  RFC1948:
   RFC4086:
   RFC4279:
-  RFC4302:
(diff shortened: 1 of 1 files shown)


davegarrett:
Well, whoops... I had the same idea: PR #686 ;)

I've got these and more over there.

xiaoyinl:
@davegarrett Yeah, what a coincidence! I realized I forgot to check the normative section after seeing your PR. Both are merged anyway.

davegarrett:
@xiaoyinl Yep. I had just read your other PR and noticed that we never culled out all the cipher suite refs after the changeover. Just lines removed, so no merge conflict on mid-air collision, and merged before I even noticed.

PR #691, opened 2016-10-14
don't specify that enumerateds can assume only specified values

tomato42:
with few possible exceptions, enumerateds will assume values
different than the ones they were originally defined with.
Backwards compatibility of Client Hello is dependent on
implementations being able to ignore values they do not
know about.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1161,14 +1161,17 @@ equivalent to the decimal value 16909060.
 
 ##  Enumerateds
 
-An additional sparse data type is available called enum. A field of type enum
-can only assume the values declared in the definition. Each definition is a
+An additional sparse data type is available called enum. Each definition is a
 different type. Only enumerateds of the same type may be assigned or compared.
 Every element of an enumerated must be assigned a value, as demonstrated in the
 following example. Since the elements of the enumerated are not ordered, they
 can be assigned any unique value, in any order.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1161,15 +1161,17 @@ equivalent to the decimal value 16909060.
...
-can only assume the values declared in the definition. Each definition is a
+An additional sparse data type is available called enum. Each definition is a
 different type. Only enumerateds of the same type may be assigned or compared.
 Every element of an enumerated must be assigned a value, as demonstrated in the
 following example. Since the elements of the enumerated are not ordered, they
 can be assigned any unique value, in any order.
 
        enum { e1(v1), e2(v2), ... , en(vn) [[, (n)]] } Te;
 
+Future extension or additions to the protocol may define new types.
```
kaduk: The new values are the important part, not the types that they indicate.

tomato42: done a simple s/types/values/

ekr:
This doesn't seem quite right. Some enums are complete in the sense that you should not ignore unknown ones (HandshakeType) and some are not complete (ExtensionType, CipherSuite).

tomato42:
How about then "...to parse and ignore them unless the their deifnition states otherwise".?

ekr:
sure that would be fine

PR #693, opened 2016-10-14
Deprecate alert level

knekritz:
After PR #625 all alerts are required to be sent with fatal AlertLevel except for close_notify, end_of_early_data, and user_canceled. Since those three alerts all have separate specified behavior, the AlertLevel field is not serving much purpose, other than providing potential for misuse. We (Facebook) currently receive a number of alerts at incorrect levels from clients (internal_error warning alerts, etc.). I propose deprecating this field to simplify implementations and require that any misuse be ignored.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3468,29 +3468,25 @@ One of the content types supported by the TLS record layer is the
 alert type.  Like other messages, alert messages are encrypted as
 specified by the current connection state.
 
-Alert messages convey the severity of the message (warning or fatal)
-and a description of the alert. Warning-level messages are used to
-indicate orderly closure of the connection (see {{closure-alerts}}).
-Upon receiving a warning-level alert, the TLS implementation SHOULD
-indicate end-of-data to the application and, if appropriate for
-the alert type, send a closure alert in response.
+Alert messages convey a description of the alert. Alerts listed in
+{{closure-alerts}} are used to indicate an orderly close of the current
+connection state and must be handled as specified.
 
-Fatal-level messages are used to indicate abortive closure of the
-connection (See {{error-alerts}}). Upon receiving a fatal-level alert,
+Alerts listed in {{error-alerts}} are used to indicate abortive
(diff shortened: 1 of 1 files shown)


ekr:
@knekritz can you raise this on-list?

ekr:
Thanks!

draft-ietf-tls-tls13.md
```diff
@@ -3468,30 +3468,26 @@ One of the content types supported by the TLS record layer is the
...
 session tickets (as in many clients) SHOULD discard tickets associated
-with failed connections.
+with failed connections. All unknown alert types MUST be treated as
+error alerts.
 
-All the alerts listed in {{error-alerts}} MUST be sent as fatal and
-MUST be treated as fatal regardless of the AlertLevel in the
-message. Unknown alert types MUST be treated as fatal.
+Prior versions of TLS used the "legacy_alert_level" field to indicate the
+severity of the alert. This field should be set to {2} for all alerts,
```
davegarrett: This "should" needs to be uppercase per RFC2119 language.

also, nit: "fatal(2)" or just "2", rather than "{2}", please

xiaoyinl: I prefer "This field **MUST** be set to 2 for all alerts", because we use "MUST" for most other deprecated fields, such as ClientHello.legacy_version, ClientHello.legacy_compression_methods,  TLSPlaintext.legacy_record_version.

draft-ietf-tls-tls13.md
```diff
@@ -3531,7 +3527,7 @@ message. Unknown alert types MUST be treated as fatal.
        } AlertDescription;
 
        struct {
-           AlertLevel level;
+           uint8 legacy_alert_level = { 2 };
```
xiaoyinl: { 2 } should be changed to 2. It is a number, not a struct.

ekr:
@seanturner: instructions?

ekr:
Closing per:
https://www.ietf.org/mail-archive/web/tls/current/msg21641.html

PR #694, opened 2016-10-15
Include text warning of 0RTT properties

wbl:
Based on list conversations, I'm including some API guidance.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1056,6 +1056,12 @@ will not process the same data twice for the same connection) and
 an attacker will not be able to make 0-RTT data appear to be
 1-RTT data (because it is protected with different keys.)
 
+These properties mean that 0-RTT data must be specially treated by applications. They SHOULD be required
+to opt-in to 0-RTT data transmission or recieving it: implementations SHOULD NOT preappend recieved 0-RTT data
+to 1-RTT data automatically,but provide special functions to examine 0-RTT data.
+
+Protocols on top of TLS 1.3 should carefully consider when 0-RTT can be used.
+
 The same warnings apply to any use of the early exporter secret.
 
 The remainder of this document provides a detailed description of TLS.


draft-ietf-tls-tls13.md
```diff
@@ -1056,6 +1056,12 @@ will not process the same data twice for the same connection) and
 an attacker will not be able to make 0-RTT data appear to be
 1-RTT data (because it is protected with different keys.)
 
+These properties mean that 0-RTT data must be specially treated by applications. They SHOULD be required
+to opt-in to 0-RTT data transmission or recieving it: implementations SHOULD NOT preappend recieved 0-RTT data
+to 1-RTT data automatically,but provide special functions to examine 0-RTT data.
```
davegarrett: missing space after comma

draft-ietf-tls-tls13.md
```diff
@@ -1056,6 +1056,12 @@ will not process the same data twice for the same connection) and
 an attacker will not be able to make 0-RTT data appear to be
 1-RTT data (because it is protected with different keys.)
 
+These properties mean that 0-RTT data must be specially treated by applications. They SHOULD be required
+to opt-in to 0-RTT data transmission or recieving it: implementations SHOULD NOT preappend recieved 0-RTT data
```
davegarrett: 'I' before 'E' except after 'C', so the correct spellings are "receiving" and "received".
"preappend" should be "prepend"

(also, wrap lines to around 80 characters, please)

draft-ietf-tls-tls13.md
```diff
@@ -1056,6 +1056,12 @@ will not process the same data twice for the same connection) and
...
+to 1-RTT data automatically,but provide special functions to examine 0-RTT data.
+
+Protocols on top of TLS 1.3 should carefully consider when 0-RTT can be used.
```
davegarrett: This sentence is a bit superfluous down here. Might be better to put a comma at the end of the first new sentence up top and stick this sentence fragment there. e.g. "... treated by applications, and protocols on top of ..."

martinthomson:
I like that you have included a requirement to have a profile for each protocol that intends to use 0-RTT, that's a fine thing to have (and it will make some people happier), but the first paragraph is a bit woolly.  You have a lowercase must, followed by a few SHOULD/SHOULD NOT statements.  You really need to be clearer and more definitive here:

> Protocols MUST NOT use 0-RTT data without a profile that defines its use.  That profile needs to identify which messages or interactions are safe to use with 0-RTT.
>
> In addition, to avoid accidental misuse, implementations SHOULD NOT enable 0-RTT unless specifically requested.  Special functions for 0-RTT data are RECOMMENDED to ensure that an application is always aware that it is sending or receiving data that might be replayed.

ekr:
I adopted Martin Thomson's text.

PR #696, opened 2016-10-15
Clarify what to do when TLSPlaintext.length is too large

xiaoyinl:
Before record protection is established, if an endpoint receives a TLSPlaintext structure whose "length" field is too large (>2^14), should it send a "record_overflow" alert?

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3228,7 +3228,9 @@ legacy_record_version
 
 length
 : The length (in bytes) of the following TLSPlaintext.fragment. The
-  length MUST NOT exceed 2^14.
+  length MUST NOT exceed 2^14. An endpoint that receives a record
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3228,7 +3228,9 @@ legacy_record_version
 
 length
 : The length (in bytes) of the following TLSPlaintext.fragment. The
-  length MUST NOT exceed 2^14.
+  length MUST NOT exceed 2^14. An endpoint that receives a record
+  that exceeds this length MUST terminate the connection with a
+  “record_overflow” alert.
```
davegarrett: please replace these with standard quote characters

xiaoyinl: @davegarrett Fixed. Thanks!

PR #699, opened 2016-10-17
Clean up key expansion

ekr:


ekr:
@martinthomson @davidben This came from Valdez's suggestion. This seems super-simple but should be safe now. Can you make sure I haven't missed anything?

martinthomson:
LGTM.  Extra layers of key separation is no longer needed now that we don't need to split the early secret into multiple keys.

draft-ietf-tls-tls13.md
```diff
@@ -3942,27 +3942,20 @@ keys.
...
 * The length of the key
 
 The keying material is computed using:
 
-       key = HKDF-Expand-Label(Secret,
-                               phase + ", " + purpose,
-                               "",
-                               key_length)
-
+       key = HKDF-Expand-Label(Secret, purpose, "", key_length_
```
davegarrett: last '_' should be a ')'

davegarrett:
The key/iv table (next table following the edited table) is verbosity overkill. Could you just scrap that table and state that it's one of two short strings inline, instead?

davidben:
LGTM, existing editorial comments aside.

Perhaps just write something like:

---
The traffic keying material is generated from an input traffic secret value using:

```
  [sender]_write_key = HKDF-Expand-Label(Secret, "key", "", key_length)
  [sender]_write_iv = HKDF-Expand-Label(Secret, "iv", "", iv_length)
```

[sender] denotes the sending side. The following table specifies the input traffic secret for each class of traffic keys:

[the table in this PR]

All the traffic keying material is recomputed whenever the underlying traffic secret changes (e.g., when changing from the handshake to application data keys or upon a key update).

---

This matches the `iv_length` and `[sender]_write_key` naming used elsewhere in the spec.

ekr:
Merged by hand

PR #702, opened 2016-10-17
Make inappropriate_fallback reserved

xiaoyinl:
The "inappropriate_fallback" alert is no longer needed for TLS 1.3, since it already has the downgrade protection mechanism of checking the server's random value.


ekr:
This doesn't seem right to me. If you have the SCSV you should still be sending this alter.

xiaoyinl:
You are right. I missed the case when a TLS 1.3 client fallback to TLS 1.1 with SCSV. In this case TLS 1.3 server does need to send this alert, because it doesn't know whether the client knows to check the server random.

PR #703, opened 2016-10-17
Alert clarifications

tomato42:
Fix few typos and clarify when and if alerts should be sent.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3512,7 +3512,8 @@ specified by the current connection state.
 
 Alert messages convey the severity of the message (warning or fatal)
 and a description of the alert. Warning-level messages are used to
-indicate orderly closure of the connection (see {{closure-alerts}}).
+indicate orderly closure of the connection or transition to
+messages ecrypted with traffic_secret (see {{closure-alerts}}).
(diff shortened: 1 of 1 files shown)


ekr:
This PR does not match the WG consensus which is that sending alerts was optional

tomato42:
But it does say that sending alerts is optional, it just explains why its not good thing to do: https://github.com/tlswg/tls13-spec/pull/703/files#diff-9d84740dcc569a0a5a359d0fba461a05R3656

ekr:
See above.

tomato42:
new line 3656 clearly states that not sending alerts is acceptable, the patches do not modify the existing text in line 3632/3645

so which line nullifies this?

ekr:
The line where you removed the word "optional"

tomato42:
That is general description of the protocol, there are no RFC 2119 keywords there. The whole section is very low on details

I don't see how that overrides the text in line 3622/364:

>  Whenever an implementation encounters a fatal error condition, it
> SHOULD send an appropriate fatal alert and MUST close the connection
> without sending or receiving any additional data.

text which  the patches don't modify

ekr:
I took the parts of this that I think are useful improvements. I don't think that the text exhorting people harder to send alerts helps and that the alerts to send in the cases of fragment errors are clear enough already.

PR #712, opened 2016-10-19
using struct to defined PskBinderEntry.

kazu-yamamoto:
According our criteria of one-field-struct vs non-struct, this should be a one-field-struct. (My parser warns this.)


ekr:
This seems like an incorrect application. It's not a standalone message, just a field

kazu-yamamoto:
OK. Let's close this.

PR #716, opened 2016-10-19
s/TLS v1.x/TLS v1.0/

kazu-yamamoto:
The comment of legacy_version in ClientHello was updated and fixed.
Record layer should follow this.


tomato42:
No, I think that's correct as this value is shared with all the TLS versions since TLSv1.0 up to and including TLSv1.3. So that value indicates that it's negotiating a TLS connection, not SSL 3 one.

ekr:
I concur with @tomato42

PR #717, opened 2016-10-19
Revise syntax description

ekr:
- The caveats at the beginning are silly
- We don't seem to every use  Foo = { A, B} but we do use Foo = A;

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1088,12 +1088,7 @@ The remainder of this document provides a detailed description of TLS.
 
 This document deals with the formatting of data in an external representation.
 The following very basic and somewhat casually defined presentation syntax will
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1088,12 +1088,7 @@ The remainder of this document provides a detailed description of TLS.
 
 This document deals with the formatting of data in an external representation.
 The following very basic and somewhat casually defined presentation syntax will
-be used. The syntax draws from several sources in its structure. Although it
-resembles the programming language "C" in its syntax and XDR {{RFC4506}} in
```
davegarrett: This is the only reference to RFC4506 in this document, so you can also remove it from the informative references list.

ekr:
Fixed, thanks

PR #723, opened 2016-10-19
A bunch of minor editorial fixes

martinthomson:
I believe that these are just cleanup, though you might want to check the PSK one a little more carefully (happy to separate it if it is a problem).


draft-ietf-tls-tls13.md
```diff
@@ -2242,10 +2247,12 @@ selected_identity
...
-Each PSK is associated with a single Hash algorithm. For PSKs established
-via the ticket mechanism ({{NewSessionTicket}}), this is the Hash used for
-the KDF. For externally established PSKs, the Hash algorithm MUST be set when the
-PSK is established.
+Each PSK is associated with a single cipher suite. For PSKs established
+via the ticket mechanism ({{NewSessionTicket}}), this is the cipher
+suite selected for the connection where the ticket was created. For
+externally established PSKs, the cipher suite MUST be set when the PSK
+is established.  The PRF for the cipher suite determines the Hash
+function used for calculating the PSK binder.
```
ekr: Yeah, so I explicitly went the other way in my latest PR. It was inconsistent between external PSKs and resumption. And if you bind a PSK to the cipher suite than it means that you can't have a PSK that you establish externally and then have any negotiation, so you're stuck with (e.g., GCM) forever.

There are three choices:
1. Bind it to cipher suite always
2. Bind it to KDF always (and then require AEAD match for 0-RTT)
3. Bind to KDF for external PSK and cipher suite for resumption (ugh).

I went for #2, but I can see people feeling differently.

PR #726, opened 2016-10-20
minor cleanup Early Data Indication

davegarrett:
This extension is now empty. Minor cleanup.


ekr:
I think "it's own" is useful, but my instinct is that having the empty struct makes life easier.

davegarrett:
If you would rather keep the empty struct, I don't have a strong objection. Repushed with just the minor wording tweak.

PR #728, opened 2016-10-20
specify what to do when receive wrong fragments

xiaoyinl:
This clarifies what to do when an endpoint receives a zero-length fragment of Handshake or Alert type, or a fragmented Alert message [1]. This addresses the issue that @tomato42 posted on list [2].

[1] https://www.ietf.org/mail-archive/web/tls/current/msg21631.html.
[2] https://www.ietf.org/mail-archive/web/tls/current/msg21629.html


ekr:
I think this is clear enough in the decode_error text.

tomato42:
decode_error only talks about incorrect message lengths, not incorrect record lengths

PR #729, opened 2016-10-21
extend insufficient_security to not apply only to ciphers

tomato42:
not only ciphers are negotiatied during handshake, also group
parameters and signature algorithms, all of which can cause
abort because the intersection between ones supported by server
and ones advertised by client is null.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3611,7 +3611,8 @@ record_overflow
 handshake_failure
 : Reception of a "handshake_failure" alert message indicates that the
   sender was unable to negotiate an acceptable set of security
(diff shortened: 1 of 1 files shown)


davegarrett:
I think this is a good idea since the new negotiation split up the cipher suites in favor of individual algorithm negotiation. I would suggest adding blurbs in the actual sections that would be applicable, rather than just changing the definition. (also, please add a changelog entry)

ekr:
I don't have a problem with expanding the set of things you send insufficient parameters for, but this seems like a larger change than needed. All that's needed here is to replace "ciphers" with "security parameters" in the second block.

ekr:
I made the change indicated in my comment above

PR #731, opened 2016-10-22
add/tweak some references

davegarrett:
Just noticed that the MTI suites section didn't link to the suites, so here's a minor PR with some reference fiddling for this section:
* Add a link in MTI suites to the suites section, now that they're defined in this document.
* Nitpick ordering of extensions in MTI extensions to have them in same order as sections (so numbers in links are consistient).
* Change plaintext reference to appendices in Security Considerations to actual references.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3979,7 +3979,7 @@ all exporter computations.
 In the absence of an application profile standard specifying otherwise, a
 TLS-compliant application MUST implement the TLS_AES_128_GCM_SHA256
 cipher suite and SHOULD implement the TLS_AES_256_GCM_SHA384 and
-TLS_CHACHA20_POLY1305_SHA256 cipher suites.
+TLS_CHACHA20_POLY1305_SHA256 cipher suites. (see {{cipher-suites}})
 
 A TLS-compliant application MUST support digital signatures with
 rsa_pkcs1_sha256 (for certificates), rsa_pss_sha256 (for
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3993,11 +3993,11 @@ In the absence of an application profile standard specifying otherwise, a
 TLS-compliant application MUST implement the following TLS extensions:
 
   * Supported Versions ("supported_versions"; {{supported-versions}})
+  * Cookie ("cookie"; {{cookie}})
```
ekr: Why did you move this?

draft-ietf-tls-tls13.md
```diff
@@ -3979,7 +3979,7 @@ all exporter computations.
 In the absence of an application profile standard specifying otherwise, a
 TLS-compliant application MUST implement the TLS_AES_128_GCM_SHA256
 cipher suite and SHOULD implement the TLS_AES_256_GCM_SHA384 and
-TLS_CHACHA20_POLY1305_SHA256 cipher suites.
+TLS_CHACHA20_POLY1305_SHA256 cipher suites. (see {{cipher-suites}})
```
ekr: This seems unnecesary.

draft-ietf-tls-tls13.md
davegarrett: https://tools.ietf.org/html/draft-ietf-tls-tls13-17#section-8.2

When linkified, it currently goes Section 4.2.1, 4.2.3, 4.2.4, 4.2.5, 4.2.6, 4.2.2. Just putting them in order. (as I said, just a nitpick)

draft-ietf-tls-tls13.md
davegarrett: They're bran new suites, so to someone first reading this document, nothing about that has really been stated as of yet. It occurred to me that could be confusing, hence the PR just to add a ref here. It's a big spec; cross-referencing can help, where applicable.

PR #736, opened 2016-10-25
Clean up Finished handshake table. No semantic change. Fixes #733

### LINKED ISSUE #733: Clean up Finished key table
https://tlswg.github.io/tls13-spec/#rfc.section.4.4

The 1-RTT annotations are redundant and you also don't need [sender] in brackets because we can just fill in the values, as client/server have separate tables lines

ekr:


ekr:
@martinthomson please double-check

PR #741, opened 2016-10-26
make "insufficient_security" replacement for "handshake_failure"

tomato42:
In all cases where handshake_failure is valid, insufficient_security
is also logically acceptable. Make this explicit.

(as per @davegarrett suggestion from https://github.com/tlswg/tls13-spec/pull/729#issuecomment-255448836)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -431,6 +431,8 @@ draft-18
 
 - Restate rule from RFC 6066 that you can't resume unless SNI is the same.
 
+- Explicitly state that "insufficient_security" may be a replacement for
+  "handshake_failure".
 
 draft-17
 
@@ -1566,7 +1568,8 @@ or HelloRetryRequest message.
 The server will send this message in response to a ClientHello message when
 it was able to find an acceptable set of algorithms and the client's
 "key_share" extension was acceptable. If it is not able to find an acceptable
-set of parameters, the server will respond with a "handshake_failure" fatal alert.
+set of parameters, the server will respond with a "handshake_failure" or
+"insufficient_security" fatal alert.
 
 Structure of this message:
 
@@ -1648,7 +1651,7 @@ groups that are mutually supported, but
 the client's ClientHello did not contain sufficient information to
 proceed with the handshake.
 If a server cannot successfully select algorithms, it MUST abort
-the handshake with a "handshake_failure" alert.
+the handshake with a "handshake_failure" or "insufficient_security" alert.
 
 Structure of this message:
 


ekr:
This isn't correct.

Say that the client only offers P521 but the server only supports P384. That would be handshake_failure but not insufficient_security. Another example would be if the client
only offered pre_shared_key but not key_shares, and the server didn't accept PSK.

tomato42:
the alert description says when insufficient_security should be used, but the standard doesn't say that in all cases the handshake_failure is applicable the insufficient_security MAY be used

but yes, it is highly dependent on what server has enabled and what client advertised

we can add examples saying when you should send insufficient_security and when you should send handshake_failure

richsalz:
I don't think we should have inssuficient_security.  Just like we no longer say "right user, wrong password" kinds of messages.  Sadly, the world kinda stinks these days.

tomato42:
we also didn't have a tls test suite that could suggest implementing it. I don't think we tried hard enough to abandon it just yet

richsalz:
Sorry I was not clear.  My point is that this gives out too much information these days.

tomato42:
how is saying "I see you've advertised export grade ciphersuites and RC4 only, I won't use this garbage" giving "too much information"?

ekr:
"the alert description says when insufficient_security should be used, but the standard doesn't say that in all cases the handshake_failure is applicable the insufficient_security MAY be used".

It doesn't say that because it's not true, as I indicated in my initial comment. If you feel strongly about this, you need to raise it on the list and get consensus.

tomato42:
but it is true the other way round, with client supporting only P384 and server supporting only P521, or more likely, client advertising only aes-128 and server supporting aes-256 and chacha20 only, then it is applicable

ekr:
You use "in all instances" in your comment 10 hours ago.

In any case, I believe that this is an issue of editor discretion.

tomato42:
by that I meant all instances of "handshake_failure" occurrence in the text, not all instances when "handshake_failure" can be sent

read it as "I inspected all places where there are references to 'handshake_failure', given my analysis, in all situations where sending that alert is required there can exist situations in which 'insufficient_security' is also valid alert to send".

I intended the note to be for You/editor, not implementer. The implementer should refer only to RFC text, shouldn't it?

richsalz:
Those ciphers aren't defined in TLS 1.3 :)  But maybe it's not export grade, but maybe it's a particular curve or DHparam size.

ANY information other than "no" can, apparently, be turned into some kind of oracle.

tomato42:
> Those ciphers aren't defined in TLS 1.3 :)

It was just an example, in future it may be AES-128 and P-256 instead of RC-4

The point is to differentiate the failure of "I (the server) know the ciphers you've advertised but I don't implement them or don't enable them because they are too weak" (insufficient_security) from "None of the ciphers you've listed match the ones I have enabled, but I see also ones I don't know of, so we may be unable to negotiate because *I* implement too old ciphers or a different set of ciphers" (handshake_failure)

> ANY information other than "no" can, apparently, be turned into some kind of oracle.

yes, including the information how quickly you have rejected the connection...

PR #742, opened 2016-10-26
extend "decode_error" to records with invalid length

tomato42:
see also https://github.com/tlswg/tls13-spec/pull/728#issuecomment-256307450

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3647,7 +3647,9 @@ access_denied
 
 decode_error
 : A message could not be decoded because some field was out of the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3647,7 +3647,9 @@ access_denied
 
 decode_error
 : A message could not be decoded because some field was out of the
-  specified range or the length of the message was incorrect.
+  specified range, the length of the message was incorrect or the record
```
tomato42: done

ekr:
I think this is clear enough already

PR #756, opened 2016-11-09
Editorial

kaduk:
It's less overwhelming to do this in chunks as I make a review pass over the -18 for WGLC.
I may add more commits as I progress, but feel free to merge before I am done.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1486,7 +1486,7 @@ Structure of this message:
            opaque legacy_session_id<0..32>;
            CipherSuite cipher_suites<2..2^16-2>;
            opaque legacy_compression_methods<1..2^8-1>;
-           [[ Extension extensions<8..2^16-1>; ]]
+           Extension extensions<8..2^16-1>;
        } ClientHello;
 
 All versions of TLS allow extensions to optionally follow the
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -387,7 +387,7 @@ implementors of protocols that run on top of TLS.
 
 This document defines TLS version 1.3. While TLS 1.3 is not directly
 compatible with previous versions, all versions of TLS incorporate a
-versioning mechanism which allows clients and servers to interoperably
+versioning mechanism which allows compliant clients and servers to interoperably
 negotiate a common version if one is supported.
 
 ##  Conventions and Terminology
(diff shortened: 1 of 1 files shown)


ekr:
You'll probably want to make this build

martinthomson:
It probably does build.  kramdown-rfc2629 has been getting consistent certificate errors from xml2rfc.tools.ietf.org.  I've told Henrik that he needs to add an intermediate.

draft-ietf-tls-tls13.md
```diff
@@ -1483,11 +1486,12 @@ Structure of this message:
            opaque legacy_session_id<0..32>;
            CipherSuite cipher_suites<2..2^16-2>;
            opaque legacy_compression_methods<1..2^8-1>;
-           Extension extensions<0..2^16-1>;
+           [[ Extension extensions<8..2^16-1>; ]]
```
davegarrett: I don't think inventing new syntax is a good idea here. Adding a comment on the line stating "OPTIONAL in old versions" would be better.

martinthomson:
Build should now succeed if another commit is pushed (maybe one to remove the new notation Dave identified).

draft-ietf-tls-tls13.md
martinthomson: Or we could take the 8 (is it only 8?) on the basis that we are describing what a TLS 1.3 client has to send, which always includes key_share or pre_shared_key and signature_algorithms.

davegarrett:
On Thursday, November 10, 2016 10:34:02 am kaduk wrote:
> It's not new syntax; Section 3.2 notes that optional components are denoted by enclosing them in double brackets.

I stand corrected. It appears to only be used in the Presentation Language section. It's valid, but underused. If we're ok with not using it, I would still prefer to avoid it. Adding some note in a comment would be my preference vs. the current state.

kaduk:
> It appears to only be used in the Presentation Language section

Right.  It occurred to me after I wrote the previous note, that perhaps the intent was to only
use the double-brackets-for-optional syntax when describing the presentation language itself (i.e., for optional struct tags), and not necessarily for structures defined using the presentation language.  Maybe @ekr has a better sense?

ekr:
I merged most of this, and then did some more editorial work. Note in particular I didn't take the "compliant" piece because I don't think that makes sense in a spec context.

Yes, I read [[]] as being for the presentation language.

PR #757, opened 2016-11-10
Clarify encoding of sequence number

siyengar:
Clarify encoding of sequence number in nonce. I believe the intention of it is to be little endian. 
This could help offer a hint to implementations to "do the right thing"


davidben:
Unless I'm completely confused, it's big endian. RFC 7905 even explicitly says big endian.

ekr:
It's big-endian. That's certainly what we do

siyengar:
ah i misunderstood then, ill change the text to be big-endian :).

siyengar:
Thanks for pointing out rfc7905 @davidben. Would be useful to have that big endian text here too.

PR #759, opened 2016-11-13
Sequence numbers do not wrap

xiaoyinl:
If the sequence number is incremented by 1 each time an endpoint sends a record, how could it wrap, since it is so large (64-bit)?

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3354,13 +3354,11 @@ A 64-bit sequence number is maintained separately for reading and writing
 records.  Each sequence number is set to zero at the beginning of a connection
 and whenever the key is changed.
 
-The sequence number is incremented after reading or writing each record.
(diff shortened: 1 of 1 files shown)


davegarrett:
Sending enough records to need to wrap is a real possibility, and has come up in discussion at some point, I think. Long running servers could eventually hit the limit, theoretically. In any case, how to handle needs to be specified. The current text is fine, as-is.

martinthomson:
Sending at 1Tbps with 1 octet records would take 93 years to reach the limit.  This assumes that we remove the three redundant bytes in the record header.  Maybe we shouldn't worry about this.

xiaoyinl:
Let's say there is a 4 GHz (=2^32 clock cycles/s) processor continuing executing this code only: `while (true) sequence_number++;` It still takes 136.2 years (=2^32 s) to overflow a uint64.

PR #762, opened 2016-11-14
Preliminary PR to remove the fixed three-byte header pending

ekr:
measurements.

This is a placeholder for discussion.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1686,6 +1686,15 @@ this specification are:
 
 - key_share (see {{key-share}})
 
+If the client offers 0-RTT and the server responds with HelloRetryRequest,
+the server will need to be able to discard any 0-RTT encrypted data
+while waiting for the repeated ClientHello. This is somewhat complicated
+by the fact that the plaintext ClientHello will have a five byte header
+and the encrypted records have a two-byte header starting with the
+length. The server can distinguish these by examining the first bit of
+the first byte of the record, which will be high if and only if the
+record is encrypted.
+
 
 ##  Extensions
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3264,17 +3273,18 @@ by an encrypted body, which itself contains a type and optional padding.
...
 
-type
-: The content type of the record.
+:masked_length
+
+The length value with the high bit set. Records are always less than
+2^15 bytes, so this bit is unused. Having the high bit set allows
+receivers to easily distinguish TLS 1.3 encrypted packets from
+TLS 1.2 compatible framed plaintext packets (which start with the
+content type).
```
martinthomson: Might want to point out that plaintext packets are always either alerts or handshake content types, which don't set the high bit.

draft-ietf-tls-tls13.md
```diff
@@ -3264,17 +3273,18 @@ by an encrypted body, which itself contains a type and optional padding.
        } TLSInnerPlaintext;
 
        struct {
-           ContentType opaque_type = 23; /* application_data */
-           ProtocolVersion legacy_record_version = 0x0301; /* TLS v1.x */
-           uint16 length;
+           uint16 masked_length;
```
tmshort: A 32-bit header would be better for crypto off-loading, but 16-bits is better than 5 bytes.

mimoo:
Why not using a `ContentType encrypted_type = 24` instead of `masked_length`.

This would make it easier to differentiate it from a `TLSPlaintext` message as well.

ekr:
Well, this PR is intended to remove that byte entirely.

You still need to have legacy_record_version in unencrypted records.

mimoo:
> Well, this PR is intended to remove that byte entirely.

I just find it cleaner (and it would add some sense to the specification as well) compared to a "hack" bit inside of an already used `uint16`. But I have no real argument for it :)

> You still need to have legacy_record_version in unencrypted records.

Right, although there are still some mentions to `TLSCiphertext.legacy_record_version`.

ekr:
Closed per WG consensus in Chicago

PR #763, opened 2016-11-14
EC point validation (not yet compiled)

matthewdgreen:
This pull request includes additional text and references specifying public value validation when using EC Diffie-Hellman. This PR has not been compiled (because my environment is set up wrong and I'm lazy) so I'm throwing it over the fence as is.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -138,6 +138,23 @@ informative:
          ins: M. Dworkin
        seriesinfo:
          NIST: Special Publication 800-38D
+
+  KEYAGREEMENT:
+       title: "Recommendation for Pair-Wise Key Establishment Schemes Using Discrete Logarithm Cryptography"
+       date: 2013-05
+       author:
+         org: American National Standards Institute
+          -
+           ins: E. Barker
+          -
+           ins: Lily Chen
+          - 
+           ins: A. Roginsky
+          -
+           ins: M. Smid
+        seriesinfo:
+          NIST: Special Publication 800-38D
+
   PKCS6:
        title: "PKCS #6: RSA Extended Certificate Syntax Standard, version 1.5"
        author:
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2178,10 +2178,10 @@ specified in this document MUST only be used with the uncompressed
 point format (the format for all ECDH functions is considered
 uncompressed). Peers MUST validate each other's public value Y by ensuring
 that the point is a valid point on the elliptic
-curve, and that the point does not generate a small subgroup.
+curve.
 
 For the curves secp256r1, secp384r1 and secp521r1, the appropriate
-validation procedures are defined in Section 4.3.7 of {{X962}
+validation procedures are defined in Section 4.3.7 of {{X962}}
 and alternatively in Section 5.6.2.6  of {{KEYAGREEMENT}}.
 This process consists of three steps: (1) verify that Y is not the point at 
 infinity (O), (2) verify that for Y = (x, y) both integers are in the correct
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2178,17 +2178,24 @@ specified in this document MUST only be used with the uncompressed
 point format (the format for all ECDH functions is considered
 uncompressed). Peers MUST validate each other's public value Y by ensuring
 that the point is a valid point on the elliptic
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2159,7 +2176,15 @@ Although X9.62 supports multiple point formats, any given curve
 MUST specify only a single point format. All curves currently
 specified in this document MUST only be used with the uncompressed
 point format (the format for all ECDH functions is considered
-uncompressed).
+uncompressed). Peers MUST validate each other's public value Y by ensuring
+that the point is a valid point on the elliptic
+curve. For all elliptic curves except for x25519 and x448, the appropriate
```
dconnolly: This reads as requiring X25519 and X448 to validate Y too, albeit with a different procedure that for the NIST curves, but that procedure is not linked. While X25519/X448 are supposed to be twist-secure, does this addition require point validation for them? If so, the technique should be referenced here.

matthewdgreen: My view is that there is a validation procedure for X25519/X448, and it involves a range check and a few comparison checks for a limited number of generators of small subgroups. I can add this if you think it would be helpful, though I expect people will immediately start fighting about it.

ekr: If this is going to be a specification requirement it does need to be clear. That means either:

1. Specify a procedure
2. Point to a procedure
3. Rewrite the text so it doesn't require this for X25519/X448 and say why.

briansmith:
Rather than add new text saying that the points must be validated, the specification should just say that the key agreement must be implemented according to the specification for that key agreement protocol. Then give the NIST document and the X25519/X448 document as the specifications for the specific groups.

In particular, the NIST document has a lot more to say about correctly implementing ECDH than just the public key validation. IMO it doesn't make sense to call out only one requirement of the NIST document and ignore the other requirements in the spec.

matthewdgreen:
The text already says that Finite Field DH points have to be validated. So that bridge has already been crossed. There's no good reason to omit it for ECDH. The question is whether it's enough to chuck the reader at a complex NIST document, or if we can summarize the important points. 

(My feeling on this issue: go look at every small TLS library that is not OpenSSL or by Thomas Pornin and see if they screwed it up, despite the availability of FIPS documents. Hint: they did.)

draft-ietf-tls-tls13.md
matthewdgreen: I am fine with specifying the details for X25519 and X448. But I really need to research it before we include anything. In particular, there are some bad points in Curve25519 that may have implications for things like Triple-Handshake style attacks. I don't know if these cases are already handled by the X25519 specification, so I'll try to figure it out.

ekr:
@matthewdgreen: assuming WG consensus, I am happy to accept this PR with a placeholder for X25519/X448 validation

draft-ietf-tls-tls13.md
dconnolly: Some relevant but not completely fleshed-out bits that might be good to reference:
https://tools.ietf.org/html/rfc7748#section-7

> Designers should also be aware that implementations of these curves
   might not use the Montgomery ladder as specified in this document,
   but could use generic, elliptic-curve libraries instead.  These
   implementations could reject points on the twist and could reject
   non-minimal field elements.  While not recommended, such
   implementations will interoperate with the Montgomery ladder
   specified here but may be trivially distinguishable from it.  For
   example, sending a non-canonical value or a point on the twist may
   cause such implementations to produce an observable error while an
   implementation that follows the design in this text would
   successfully produce a shared key.

dconnolly: @matthewdgreen the X25519/X448 RFC linked above does not mention Triple Handshake attacks.

FiloSottile:
As an implementer, please call this out explicitly, not just by referencing an ANSI document.

dconnolly:
Since RFC7748 is lacking a bit in this area and linking to NIST-800-38D hasn't worked well in the past, summarizing validation inline here sounds better for implementers.

draft-ietf-tls-tls13.md
matthewdgreen: I checked the X25519 specification, and it does include an *optional* check in Section 6.1 that detects when the DH share is in a small subgroup. My proposal is to cite the X25519/X448, and specify that the small subgroup check should be mandatory. 

The argument is that even if TLS does not specifically require this behavior, many different applications *around* TLS (e.g., channel bindings, RFC 5929) may assume contributory behavior. This is an inexpensive check that may kill a whole class of attacks.

PR #764, opened 2016-11-16
TLSCiphertext has no member called "fragment"

mattcaswell:


kaduk:
Yes.  There is a fair amount of other cleanup needed in this area; I have lots of red pen markings on a printout but haven't translated them into a patch yet.

kaduk:
Yes.  There is a fair amount of other cleanup needed in this area; I have lots of red pen markings on a printout but haven't translated them into a patch yet.

kaduk:
Bah, github was angry-unicorning on me and did not indicate that the comment went through.
I guess github doesn't like me this week...

Issue #761, opened 2016-11-14
Cite Abdalla/Bellare paper on rekeying


martinthomson:
Note that we use a serialized key derivation for KeyUpdate, an approach that is supported by the paper.

Title: Increasing the Lifetime of a Key: A Comparative Analysis of the
Security of Re-Keying Techniques

PR #765, opened 2016-11-17
(for Quynh) Limits on Key Usage

seanturner:
This is a PR for Quynh to address this thread on the TLS WG list:
https://mailarchive.ietf.org/arch/msg/tls/OB47E9Qltiry2e5ec0sjskkn-1I

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3428,18 +3428,17 @@ mechanism through TLS extensions or some other means.
 
 ## Limits on Key Usage
 
-There are cryptographic limits on the amount of plaintext which can be
-safely encrypted under a given set of keys.  {{AEAD-LIMITS}} provides
-an analysis of these limits under the assumption that the underlying
-primitive (AES or ChaCha20) has no weaknesses. Implementations SHOULD
-do a key update {{key-update}} prior to reaching these limits.
+Using AES-GCM to provide authenticity of authenticated data, content of the plaintext and information leakage (see Note below) protection for the plaintext, the maximum amount of data (padded records) which can be safely encrypted with a single key is 2^48 128-bit blocks (2^64 bytes). When the maximum amount of data is reached, the chance of having a collision among 128-bit blocks of the ciphertext is below 1^(-32) which is negligible. The recommended maximum amount of data (2^64 bytes) is made with the assumption that the (padded) record size is either 128 bits, a multiple of 128 bits or a lot larger than 128 bits. If the (padded) record size is 2^x bytes, then the data limit with GCM for each key is 2^(64-x) records.
 
-For AES-GCM, up to 2^24.5 full-size records (about 24 million)
-may be encrypted on a
-given connection while keeping a safety margin of approximately
-2^-57 for Authenticated Encryption (AE) security. For
-ChaCha20/Poly1305, the record sequence number would wrap before the
-safety limit is reached.
+When the padded record size is less than 128 bits (16 bytes), if l is the length of the padded record in bits, then the data limit with GCM for each key is reduced to 2^64 x l /128 bytes or 2^(64-x) x l/128 records.
+
+[AEAD-LIMITS] provides the analysis for the data limit recommendation for GCM above.
+
+Since the block size of AES is 128 bits, there will be collisions among different sets of ciphertext from multiple sessions using GCM (or any other modes of AES) when the total amount of the ciphertext of all considered sessions is more than 2^64 128-bit blocks. This fact does not seem to create a practical security weakness of using AES GCM. 
+
+For ChaCha20/Poly1305, the record sequence number would wrap before the safety limit is reached.
+
+Note: Information leakage in the context of TLS is a chosen-plaintext distinguishing attack where the attacker provides 2 128-bit plaintext blocks to a GCM encryption engine, after seeing one encrypted block for one of the 2 plaintext blocks, the attacker knows which plaintext block was encrypted. Or, it means that there is a collision among 128-bit blocks of the ciphertext. 
 
 #  Alert Protocol
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3428,17 +3428,18 @@ mechanism through TLS extensions or some other means.
 
 ## Limits on Key Usage
 
-Using AES-GCM to provide authenticity of authenticated data, content of the plaintext and information leakage (see Note below) protection for the plaintext, the maximum amount of data (padded records) which can be safely encrypted with a single key is 2^48 128-bit blocks (2^64 bytes). When the maximum amount of data is reached, the chance of having a collision among 128-bit blocks of the ciphertext is below 1^(-32) which is negligible. The recommended maximum amount of data (2^64 bytes) is made with the assumption that the (padded) record size is either 128 bits, a multiple of 128 bits or a lot larger than 128 bits. If the (padded) record size is 2^x bytes, then the data limit with GCM for each key is 2^(64-x) records.
+To use AES-GCM to provide authenticity of authenticated data, content of the plaintext and information leakage [0] protection for the plaintext safely, the limit of total ciphertext under a single key is ( (TLSCipherText.length / 16) / ceiling (TLSCipherText.length / 16) ) times 2^48 128-bit blocks.
 
-When the padded record size is less than 128 bits (16 bytes), if l is the length of the padded record in bits, then the data limit with GCM for each key is reduced to 2^64 x l /128 bytes or 2^(64-x) x l/128 records.
+When the data limit is reached, the chance of having a collision among 128-bit blocks of the ciphertext is below 2^(-32) which is negligible.
 
-[AEAD-LIMITS] provides the analysis for the data limit recommendation for GCM above.
+Since the block size of AES is 128 bits, there will be collisions among different sets of ciphertext from multiple sessions using GCM (or any other modes of AES) when the total amount of the ciphertext of all considered sessions is more than 2^64 128-bit blocks. This fact does not seem to create a practical security weakness of using AES GCM.
 
-Since the block size of AES is 128 bits, there will be collisions among different sets of ciphertext from multiple sessions using GCM (or any other modes of AES) when the total amount of the ciphertext of all considered sessions is more than 2^64 128-bit blocks. This fact does not seem to create a practical security weakness of using AES GCM. 
+For ChaCha20/Poly1305, the record sequence number would wrap before the safety limit is reached. See [AEAD-LIMITS] for further analysis.
 
-For ChaCha20/Poly1305, the record sequence number would wrap before the safety limit is reached.
+NOTE: Information leakage in the context of TLS is a chosen-plaintext distinguishing attack where the attacker provides 2 128-bit plaintext blocks to a GCM encryption engine, after seeing one encrypted block for one of the 2 plaintext blocks, the attacker knows which plaintext block was encrypted. Or, it means that there is a collision among 128-bit blocks of the ciphertext.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3428,18 +3428,17 @@ mechanism through TLS extensions or some other means.
...
-may be encrypted on a
-given connection while keeping a safety margin of approximately
-2^-57 for Authenticated Encryption (AE) security. For
-ChaCha20/Poly1305, the record sequence number would wrap before the
-safety limit is reached.
+Using AES-GCM to provide authenticity of authenticated data, content of the plaintext and information leakage (see Note below) protection for the plaintext, the maximum amount of data (padded records) which can be safely encrypted with a single key is 2^48 128-bit blocks (2^64 bytes). When the maximum amount of data is reached, the chance of having a collision among 128-bit blocks of the ciphertext is below 1^(-32) which is negligible. The recommended maximum amount of data (2^64 bytes) is made with the assumption that the (padded) record size is either 128 bits, a multiple of 128 bits or a lot larger than 128 bits. If the (padded) record size is 2^x bytes, then the data limit with GCM for each key is 2^(64-x) records.
+
+When the padded record size is less than 128 bits (16 bytes), if l is the length of the padded record in bits, then the data limit with GCM for each key is reduced to 2^64 x l /128 bytes or 2^(64-x) x l/128 records.
```
martinthomson: This is incorrect.  The reduction happens whenever a record contains a partial block.  The formula is therefore not simple.

draft-ietf-tls-tls13.md
```diff
@@ -3428,18 +3428,17 @@ mechanism through TLS extensions or some other means.
...
-primitive (AES or ChaCha20) has no weaknesses. Implementations SHOULD
-do a key update {{key-update}} prior to reaching these limits.
-
-For AES-GCM, up to 2^24.5 full-size records (about 24 million)
-may be encrypted on a
-given connection while keeping a safety margin of approximately
-2^-57 for Authenticated Encryption (AE) security. For
-ChaCha20/Poly1305, the record sequence number would wrap before the
-safety limit is reached.
+Using AES-GCM to provide authenticity of authenticated data, content of the plaintext and information leakage (see Note below) protection for the plaintext, the maximum amount of data (padded records) which can be safely encrypted with a single key is 2^48 128-bit blocks (2^64 bytes). When the maximum amount of data is reached, the chance of having a collision among 128-bit blocks of the ciphertext is below 1^(-32) which is negligible. The recommended maximum amount of data (2^64 bytes) is made with the assumption that the (padded) record size is either 128 bits, a multiple of 128 bits or a lot larger than 128 bits. If the (padded) record size is 2^x bytes, then the data limit with GCM for each key is 2^(64-x) records.
```
martinthomson: Nit: 1^(-32) is 1, which is hardly negligible.

This analysis contradicts the citation we have, which will create confusion.

The AEAD-LIMITS paper suggests that for a 2^-32 margin, you need to limit data to 2^38.5 records, which is 2^48.5 blocks or 2^52.5 octets.

The last sentence is incorrect in the same way.  It's even more incorrect for x<4.

As for assuming that records are multiples of 128 bits, that hides the fact that partial blocks still count toward the limit.

martinthomson:
cc @wbl for further checking

kaduk:
It's also worth noting that if people are only rekeying after LARGENUM records, then the rekeying code may ossify.  Sure, we can mention it in the "Implementation Pitfalls" section, but if we the big implementations are rekeying after every (say) 4 GB, interop testing is more likely to exercise the rekeying codepath.

draft-ietf-tls-tls13.md
QuynhDangNIST: My calculation is correct as I explained in the email. 1^(-32) was a typo, it should have been 2^(-32).

Why did you say that the last sentence is incorrect?

When padded record size is much larger than 128 bits (one block), then the few bits padding done by AES (due to the partial last block) is negligible comparing to the record size. 

When a (padded or not) record is less than 1 AES block, then the data limit is reduced to the ration of the x (bits) over 128 because each AES encryption produces 128-bit output.

draft-ietf-tls-tls13.md
QuynhDangNIST: See my explanation above. That was why I wrote "The recommended maximum amount of data (2^64 bytes) is made with the assumption that the (padded) record size is either 128 bits, a multiple of 128 bits or a lot larger than 128 bits."

draft-ietf-tls-tls13.md
martinthomson: We should not be building these recommendations based on assumptions that we know to be false (that blocks are multiples of 128 bits), and we should definitely not be taking shortcuts (assuming that blocks are always much larger than 128 bits).  This guidance is important.

QuynhDangNIST: I gave a recommended number when the (padded or not) record size is less than 128 bits. If you want a precise number for each record size, then you could do this. Let's say x be the length of (padded or not) record size (in bits), y = 128- (x mod 128), then the data limit is ((x - y) / x ) . 2^64 bytes.

QuynhDangNIST: Interesting? When someone else gave a recommendation with the assumption that the record size was the max record size, you did not raise any issue.

martinthomson:
Unfortunately, according to the paper on this, rekeying is optimal at 2^(128/3) blocks.  Less than that has other problems (that I don't understand).

draft-ietf-tls-tls13.md
QuynhDangNIST: There are nothings misleading here. You count 128-bit blocks generally, but at the meeting yesterday KDG asked about what if the record was 1 byte.

QuynhDangNIST: Typo: " If the (padded) record size is 2^x bytes, then the data limit with GCM for each key is 2^(64-x) records." should have been "If the (padded) record size is 2^x 128-bit blocks, then the data limit with GCM for each key is 2^(48-x) records."

draft-ietf-tls-tls13.md
QuynhDangNIST: Again here: " 2^64 x l /128 bytes or 2^(64-x) x l/128 records." should have been " 2^64 x l /128 bytes ".

wbl:
@martinthomson What rekeying problems? I'd like to see that paper.

draft-ietf-tls-tls13.md
QuynhDangNIST: Ok. I revised the text above as below. The text stated the record count with the max record size of 2^14 bytes. 

"   
Using AES-GCM to provide authenticity of authenticated data, content of the plaintext and information leakage [0] protection for the plaintext, the maximum amount of data (padded records) which can be safely encrypted with a single key is 2^48 128-bit blocks (2^64 bytes). When the maximum amount of data is reached, the chance of having a collision among 128-bit blocks of the ciphertext is below 1^(-32) which is negligible. The recommended maximum amount of data (2^64 bytes) is made with the assumption that the (padded) record size is either 128 bits, a multiple of 128 bits or a lot larger than 128 bits. If the (padded) record size is 2^x 128-bit blocks, then the data limit with GCM for each key is 2^(48-x) records.  When the record size is 2^14 bytes (2^10 128-bit blocks), the data limit with GCM for each key is 2^38 records. 

Generally, if l is the length of the (padded or not) record size in bits, y = 128 – (l mod 128), then the data limit for GCM is ((l – y) / l) x 2^64 bytes. 

 "

wbl: So is the data limit that the implementation is maintaining either bytes or records? I think we need more clarity that they should count bytes and rekey before then.

briansmith:
> So is the data limit that the implementation is maintaining either bytes or records? I think we need more clarity that they should count bytes and rekey before then.

I recommend that the limit be specified precisely, which seems to mean specifying it in blocks, not bytes or records. The attempts to put this into byte or record terms have only created a lot of unnecessary confusion.

draft-ietf-tls-tls13.md
QuynhDangNIST: Counting bytes would be ok. The problem is that when the record is not multiple of 128-bit block, the last partial block could be 1 byte and encrypting 1 byte or 128-bit block is considered the same as 1 query.  So, if the record size is 1 byte, the data limit would need to be reduced to 16 times.

richsalz:
+1 to Brian's recommendation.

QuynhDangNIST:
Rich and Brian, 

Agree. But, somebody wants bytes, somebody else wants records. When saying blocks, we must assume that the record size is multiple of blocks and there is at least one person not happy about that.  The revised text should provide the answer.

wbl:
I've got no objection to counting blocks: It might be easier to implement that way.

briansmith:
> Agree. But, somebody wants bytes, somebody else wants records. When saying blocks, we must assume that the record size is multiple of blocks and there is at least one person not happy about that. The revised text should provide the answer.

Every record is potentially a different size, so specifying the limit in terms of records simply doesn't make sense. I think it should be relatively easy to get people to agree on that.

Stating something like "The number of blocks in an AES-GCM TLSCiphertext is (len(TLSCiphertext.encrypted_record) + 15) / 16 where "/" denotes integer division." is sufficient to allow people to convert to bytes.

brainhub:
I have a few concerns with the original text. 

1. "2^48 128-bit blocks" is not 2^64 bytes. It is 2^48 * 16 = 2^52 bytes (needed for 2^-32 probability of collision).

2. The original text makes claims about beyond birthday bound security (BBB), e.g.:

> "This fact does not seem to create a practical security weakness of using AES GCM." 

AES GCM is not a BBB block cipher. I think discussion of BBB security of AES-GCM is beyond the scope of this document. For example, the padding feature in TLS 1.3, which provides privacy of plaintext size, is broken at BBB limits of AES-GCM, as noted [here](https://www.ietf.org/mail-archive/web/cfrg/current/msg07516.html) . If one cares about hiding the plaintext size, the suggestion above is misleading.

3. It is not clear to me that the formula for partial blocks is correct. This use is basically a small-domain encryption. I can't make sense of 2^(64-x) x l/128 . What's x here? Is l=8 in my scenario? Note that with probability 2^-8 we know each plaintext, well before the 2^-32 margin assumed at the beginning.

QuynhDangNIST:
The probability for a collision of 2^(-32) was already taken into account.  Right, can't get much of a margin for a plaintext of a few bits.  That was why I wrote "Generally" and the revised text was "  Generally, if l is the length of the (padded or not) record size in bits, y = 128 – (l mod 128), then the data limit for GCM is ((l – y) / l) x 2^64 bytes. ". Counting the amount of plaintext, the last block could be a partial block (less than 128 bits), encrypting this partial block is still considered as one query. And, that was what Martin asked about. 

I don't see anything wrong about saying this " Since the block size of AES is 128 bits, there will be collisions among different sets of ciphertext from multiple sessions using GCM (or any other modes of AES) when the total amount of the ciphertext of all considered sessions is more than 2^64 128-bit blocks. This fact does not seem to create a practical security weakness of using AES GCM.  "

QuynhDangNIST:
Consider how much AES has been used, specifically how much AES GCM has been used, of course there are collisions in the collection of all ciphertext.

martinthomson:
@wbl check out the citation in #761 for more info on when to rekey.  The paper says that the optimal time to rekey is after 2^(Q/3) blocks, where Q is the block size.  I confess that I haven't read all the paper, just the conclusions.

QuynhDangNIST:
Brian, 
What you suggested makes sense to me. To address the ciphertext limit in term of blocks, we could add text like this "The limit of total ciphertext under a single key is ( (TLSCipherText.length / 16) / ceiling (TLSCipherText.length / 16) ) . 2^48 128-bit blocks."

The calculation took into account the partial block situation.

PR #768, opened 2016-11-20
D-H key re-use considerations

yoavnir:
Added a subsection to the Security Considerations section. It discusses key reuse (do it carefully or do it not).
It has the "don't do this or this grape juice might ferment" weasel words suggested by someone at the meeting.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -324,6 +324,15 @@ informative:
          ins: Florian Weimer
          org: Red Hat Product Security
        date: 2015-09
+  Menezes:
+       title: "On Reusing Ephemeral Keys In Diffie-Hellman Key Agreement Protocols"
+       author:
+       -
+         ins: A. Menezes
+       -
+         ins: B. Ustaoglu
+       seriesinfo: International Journal of Applied Cryptography, Volume 2 Issue 2, Pages 154-158 
+       date: January, 2010
 --- abstract
 
 This document specifies version 1.3 of the Transport Layer Security
@@ -4037,6 +4046,26 @@ lacking a "server_name" extension by terminating the connection with a
 Security issues are discussed throughout this memo, especially in
 {{implementation-notes}}, {{backward-compatibility}}, and {{security-analysis}}.
 
+## Re-using Ephemeral shares
+
+This document refers to client and server key shares as ephemeral. This literally
+means that they are used for a short time. Keeping the private key secret forever
+is necessary for the property of Perfect Forward Secrecy. The simplest and safest 
+way to achieve this is to generate a new Diffie-Hellman share for each TLS 
+handshake. This has some cost in processing power. 
+
+Alternatively, it is possible to save some CPU cycles by generating a 
+Diffie-Hellman share and using it for several TLS handshakes for a short amount
+of time. If a key is used more than once, then the public key validations in 
+{{ffdhe-param}} and {{ecdhe-param}} become MUST requirements, or else parts of
+the private key may leak. See {{Menezes}} for more details.
+
+In all cases it is mandatory to keep the private key of the Diffie-Hellman key
+share from leaking. Measures to ensure that are implementation-dependent, but
+often include wiping the memory that held the key after the last use, preventing
+swap-out of that memory, and not providing an interface for the export of such 
+private keys. Possession of such keys will allow an attacker to decrypt the 
+contents of any TLS connection that was set up using this key.
 
 #  IANA Considerations
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4046,26 +4046,28 @@ lacking a "server_name" extension by terminating the connection with a
 Security issues are discussed throughout this memo, especially in
 {{implementation-notes}}, {{backward-compatibility}}, and {{security-analysis}}.
 
-## Re-using Ephemeral shares
+## Re-Using Ephemeral Shares
 
 This document refers to client and server key shares as ephemeral. This literally
 means that they are used for a short time. Keeping the private key secret forever
 is necessary for the property of Perfect Forward Secrecy. The simplest and safest 
 way to achieve this is to generate a new Diffie-Hellman share for each TLS 
-handshake. This has some cost in processing power. 
+handshake, and to destroy keys immediately after they are used. This has some 
+cost in processing power. 
 
-Alternatively, it is possible to save some CPU cycles by generating a 
-Diffie-Hellman share and using it for several TLS handshakes for a short amount
-of time. If a key is used more than once, then the public key validations in 
-{{ffdhe-param}} and {{ecdhe-param}} become MUST requirements, or else parts of
-the private key may leak. See {{Menezes}} for more details.
+It is possible to save some CPU cycles by generating a Diffie-Hellman share and 
+using it for several TLS handshakes for a short amount of time. If a key is 
+used more than once, then the public key validations in {{ffdhe-param}} and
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4037,6 +4046,26 @@ lacking a "server_name" extension by terminating the connection with a
 Security issues are discussed throughout this memo, especially in
 {{implementation-notes}}, {{backward-compatibility}}, and {{security-analysis}}.
 
+## Re-using Ephemeral shares
```
martinthomson: Consistent capital Letters?

draft-ietf-tls-tls13.md
```diff
@@ -4037,6 +4046,26 @@ lacking a "server_name" extension by terminating the connection with a
...
+
+This document refers to client and server key shares as ephemeral. This literally
+means that they are used for a short time. Keeping the private key secret forever
+is necessary for the property of Perfect Forward Secrecy. The simplest and safest 
+way to achieve this is to generate a new Diffie-Hellman share for each TLS 
+handshake. This has some cost in processing power.
```
martinthomson: The simplest and safest way to achieve this is to generate a new Diffie-Hellman share for each TLS handshake, **and to destroy keys after they are used**.

draft-ietf-tls-tls13.md
```diff
@@ -4037,6 +4046,26 @@ lacking a "server_name" extension by terminating the connection with a
...
 
+## Re-using Ephemeral shares
+
+This document refers to client and server key shares as ephemeral. This literally
+means that they are used for a short time. Keeping the private key secret forever
+is necessary for the property of Perfect Forward Secrecy. The simplest and safest 
+way to achieve this is to generate a new Diffie-Hellman share for each TLS 
+handshake. This has some cost in processing power. 
+
+Alternatively, it is possible to save some CPU cycles by generating a
```
martinthomson: I would drop "Alternatively, "

draft-ietf-tls-tls13.md
```diff
@@ -4037,6 +4046,26 @@ lacking a "server_name" extension by terminating the connection with a
...
+of time. If a key is used more than once, then the public key validations in 
+{{ffdhe-param}} and {{ecdhe-param}} become MUST requirements, or else parts of
+the private key may leak. See {{Menezes}} for more details.
+
+In all cases it is mandatory to keep the private key of the Diffie-Hellman key
+share from leaking. Measures to ensure that are implementation-dependent, but
+often include wiping the memory that held the key after the last use, preventing
+swap-out of that memory, and not providing an interface for the export of such 
+private keys. Possession of such keys will allow an attacker to decrypt the 
+contents of any TLS connection that was set up using this key.
```
martinthomson: Possession of **either a client or server Diffie-Hellman private key** will allow an attacker to decrypt the contents of any TLS connection [...]

draft-ietf-tls-tls13.md
yoavnir: I'm changing the text, but you want to destroy the keys after use whether they are single-use or not.

draft-ietf-tls-tls13.md
yoavnir: OK

draft-ietf-tls-tls13.md
yoavnir: OK

tomato42:
From the paper:
> Safe prime groups. In these groups, G is the subgroup of prime order q of the multiplicative group of a prime field Zp, where p = 2q+ 1. If A checks that the public key B is an integer in the interval [2, p−2] then, since the multiplicative group of Zp has order 2q, attacks like the ones described above can reveal at most a single bit of a. Thus small-subgroup attacks are not effective in the case of safe prime groups.

And for ECDHE:
> Let E:V2=U3+αU+β be an elliptic curve of prime order q defined over a prime field Fp, (...)A can easily thwart the attack by checking that B is a point in E(Fp) (and is not the point at infinity).

Given the simplicity of checks, I don't see why they shouldn't be mandatory in the first place. Secondly, the ecdhe section (#ecdhe-param) doesn't specify what kind of checks should be performed on the share...

yoavnir:
@tomato42 

> Given the simplicity of checks, I don't see why they shouldn't be mandatory in the first place. 

I agree.

> Secondly, the ecdhe section (#ecdhe-param) doesn't specify what kind of checks should be performed on the share...

Well, there's #763 that proposes this, and the feeling in the room was that people liked this. My PR is a little forward-looking then.

draft-ietf-tls-tls13.md
```diff
@@ -4037,6 +4046,28 @@ lacking a "server_name" extension by terminating the connection with a
...
 
+## Re-Using Ephemeral Shares
+
+This document refers to client and server key shares as ephemeral. This literally
+means that they are used for a short time. Keeping the private key secret forever
+is necessary for the property of Perfect Forward Secrecy. The simplest and safest 
+way to achieve this is to generate a new Diffie-Hellman share for each TLS 
+handshake, and to destroy keys immediately after they are used. This has some 
+cost in processing power. 
+
```
ekr: @tvdmerwe @prosecco @adl, I'd like to get some guidance here. Do your proofs assume that the DH keys are fresh?

tvdmerwe: Yes, our symbolic proofs assume that the DH keys are fresh. This needs to be true to achieve PFS as defined in [DOW92]. The definition in [DOW92] refers to the 'safety' (secrecy) of exchanged keys from earlier runs; perhaps the definition in Appendix D should be altered to reflect this, the current phrasing is strange. 

Also, this phrasing seems somewhat odd to me "Keeping the private key secret forever is necessary for the property of Perfect Forward Secrecy." As phrased here, this suggests that compromise should never occur but PFS is defined in terms of a compromise.

tvdmerwe: But you don't get PFS if you're using the RMS only, right? That's why you need new DH shares.

ekr: @prosecco @tvdmerwe thanks.

I agree it's clear that you don't get PFS if you don't use fresh DH keys. I think the question is whether the AKE properties of your proofs depend on the DH keys being fresh. @tvdmerwe is this something you could relax in the Tamarin model easily?

Basically, my concern here is that this text seems to say that it's OK to use non-fresh DH shares for 1.3, with the cost being PFS. If we're to incorporate text like that, it would be nice to have some analytic results which indicate that it's true.

ekr: @tvdmerwe: I believe the properties wrt the RMS are as follows:

- Knowledge of the RMS established in connection N is not sufficient to decrypt N
- If you establish a new connection N+1 with the RMS as the PSK and do not do DH, then knowledge of the RMS is sufficient to decrypt N+1. However, N+1 generates its own RMS and that is not sufficient to decryption N+1 (though of course you can use it for N+2).
- If you operate a session cache type system, then the server can delete the RMS once it has been used to establish the session, giving you PFS even w/o DH
- If you operate a ticket-based system, you don't get PFS until the ticket key has been destroyed.

tvdmerwe: Quick answer: yes (barring any complications that don't immediately come to mind). 

Remind me, if DH shares aren't fresh, what secrets will be new? The RMS?

ekr: I believe that in a non-resumed connection, the DH shared secret would be non-unique but the Handshake Secret and anything derived from it will be unique as long as the nonces are fresh. This also implies that the RMS will be different for two handshakes where the client and server both reuse DH shares.

ekr:
Closed per WG consensus in Chicago

PR #775, opened 2016-11-23
Rewrite the anti-downgrade protection.

ol-anssi:
Following the mail "[TLS] Draft 18 review: Downgrade protection" (https://www.ietf.org/mail-archive/web/tls/current/msg21987.html), this is a proposal to amend the anti-downgrade protection when a TLS 1.3 implementation accepts to use an earlier version.


ekr:
Merged by hand

PR #777, opened 2016-11-23
Add support for resumption across multiple domains

vasilvv:
This feature provides servers with ability to indicate that the
session ticket it issued may be used to resume a session with any other
domain for which the certificate is valid.  This allows performance
improvement for some important use cases, specifically particular types
of CDNs and multi-homed websites.


ekr:
Parking based on list discussion.

ekr:
WGLC is over, so closing

PR #789, opened 2016-11-29
Using phrase instead of C function call.

kazu-yamamoto:
It took time to understand `max(8 bytes, N_MIN)` is a C function call because this is folded in the internet-draft version and because of "bytes". I think that using a phrase is safer for folding.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3386,8 +3386,8 @@ wrap. If a TLS implementation would need to
 wrap a sequence number, it MUST either rekey ({{key-update}}) or
 terminate the connection.
 
-The length of the per-record nonce (iv_length) is set to max(8 bytes,
-N_MIN) for the AEAD algorithm (see {{RFC5116}} Section 4). An AEAD
+The length of the per-record nonce (iv_length) is set to the larger of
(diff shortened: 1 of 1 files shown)


kaduk:
Well, it would also have been less confusing if the closing parenthesis was moved to the right place, after "for the AEAD algorithm"

draft-ietf-tls-tls13.md
```diff
@@ -3386,8 +3386,8 @@ wrap. If a TLS implementation would need to
 wrap a sequence number, it MUST either rekey ({{key-update}}) or
 terminate the connection.
 
-The length of the per-record nonce (iv_length) is set to max(8 bytes,
-N_MIN) for the AEAD algorithm (see {{RFC5116}} Section 4). An AEAD
+The length of the per-record nonce (iv_length) is set to the larger of
+8 bytes and N_MIN for the AEAD algorithm (see {{RFC5116}} Section 4). An AEAD
```
martinthomson: nit: I think "and" is grammatically correct, but I would change it to "or".

kazu-yamamoto: I would like to accept opinions from English native speakers since my first language is Japanese. I cannot tell the difference.

ekr:
I find and easier.

PR #790, opened 2016-11-29
The table of where extensions could appear was getting unweildy and

ekr:
was going to get even more so when I added CertificateRequest. I've
changed it to be a list of messages.


ekr:
This shouldn't be a semantic change. @martinthomson can I get a quick once-over

kaduk:
+1
I verified that the semantics are the same.

PR #793, opened 2016-11-29
Update .travis.yml

martinthomson:
I think that this will work, maybe wait for travis to finish before merging it.  That won't be definitive, because the error you are having is in the push phase, and I can't push.


ekr:
I took a more minimal PR

PR #804, opened 2016-12-01
random values are not used anymore in the cryptographic computations

mimoo:


ekr:
They absolutely are used. You just hash in the whole transcript.

mimoo:
This paragraph is I believe only talking about the derivation of keys. In that sense the random values are not used anymore to form the "pre master secret" like they used to.

If you want to talk about signing the whole transcript, then you need the whole transcript not just the knowledge of the random values.

Unless I'm missing something?

kaduk:
Yeah, the whole transcript is needed as well as the master secret, as inputs to Derive-Secret().  So this text should probably change, but not as the current form of this diff.

PR #805, opened 2016-12-02
Using Value instead of HashValue.

kazu-yamamoto:
HashValue is confusing because this argument is not
always a hash value. For instance, "" is used sometime.


kazu-yamamoto:
Instead of this patch, adding a comment that "" is an empty hash value is also fine to me.

PR #806, opened 2016-12-02
Align end_of_early_data text and diagram with processing order.

davidben:
The alert is described as being part of the first flight, but it is
really sent when the server Finished is received, per the processing
order section. Notably, servers MUST NOT wait for end_of_early_data
before sending ServerHello through Finished. Group it with the client
Finished flight, and fix the text about it terminating the first flight.

(The division between the first and second client flight isn't
especially well-defined because early data may flow in parallel with the
ServerHello flight. But if there is to be a division, end_of_early_data
certainly falls into the second one.)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1048,8 +1048,7 @@ as with a 1-RTT handshake with PSK resumption.
          + key_share*
          + psk_key_exchange_modes
          + pre_shared_key
-         (Application Data*)
-         (end_of_early_data)     -------->
+         (Application Data*)     -------->
                                                          ServerHello
                                                         + early_data
                                                     + pre_shared_key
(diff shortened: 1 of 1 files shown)


martinthomson:
Yes, this should help avoid one of the most common mistakes in 0-RTT implementations.

draft-ietf-tls-tls13.md
```diff
@@ -1048,15 +1048,15 @@ as with a 1-RTT handshake with PSK resumption.
          + key_share*
          + psk_key_exchange_modes
          + pre_shared_key
-         (Application Data*)
-         (end_of_early_data)     -------->
+         (Application Data*)     -------->
                                                          ServerHello
```
kaduk: I thought I had seen ekr move the arrows onto their own line in order to make subsequent diffs less noisy, but it probably doesn't matter at this point.

PR #809, opened 2016-12-07
better introduction to the Cryptographic Computations section

mimoo:
Another try to this PR: https://github.com/tlswg/tls13-spec/pull/804

This time it should be more representative of the whole section.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3754,8 +3754,8 @@ New Alert values are assigned by IANA as described in {{iana-considerations}}.
 
 In order to authenticate and to start encrypting the exchange, the TLS Record
 Protocol needs to derive a suite of secrets via the key schedule. To perform
-these operations, a peer makes use of observed handshake messages, an optional
-pre-shared secret and an optional output of the (EC)DHE key exchange.
+these operations, a peer makes use of observed handshake messages and at least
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3752,9 +3752,10 @@ New Alert values are assigned by IANA as described in {{iana-considerations}}.
 
 #  Cryptographic Computations
 
-In order to begin connection protection, the TLS Record Protocol
-requires specification of a suite of algorithms, a master secret, and
-the client and server random values.
+In order to authenticate and to start encrypting the exchange, the TLS Record
+Protocol needs to derive a suite of secrets via the key schedule. To perform
+these operations, a peer makes use of observed handshake messages, an optional
+pre-shared secret and an optional output of the (EC)DHE key exchange.
```
kaduk: I would prefer "at least one of a pre-shared secret and the output of an (EC)DHE key exchange", since they are only optional if the other one is present.

mimoo: agree

mimoo:
Added the change

ekr:
I rewrote this in a somewhat different fashion

mimoo:
this is what I see, I think there is a typo:

> as is the case when +the same PSK is used

PR #810, opened 2016-12-12
Editorial (mostly)

kaduk:
Same deal as #756 , more commits may get added as they are ready, but I wanted to get this up so I can avoid having to deal with merge conflicts to the extent possible.

Some possibly controversial bits are broken out as separate commits.


draft-ietf-tls-tls13.md
```diff
@@ -1615,7 +1620,7 @@ Structure of this message:
            ProtocolVersion version;
            Random random;
            CipherSuite cipher_suite;
-           Extension extensions<0..2^16-1>;
+           Extension extensions<1..2^16-1>;
```
kaduk: I did not go through and calculate the actual minimum, but I'm pretty sure you need to send at least one extension in the ServerHello.

draft-ietf-tls-tls13.md
```diff
@@ -1517,7 +1517,7 @@ Structure of this message:
            Random random;
            opaque legacy_session_id<0..32>;
            CipherSuite cipher_suites<2..2^16-2>;
-           opaque legacy_compression_methods<1..2^8-1>;
+           opaque legacy_compression_methods[2] = 0x0100;
```
martinthomson: I considered doing this to legacy_session_id as well, but the understanding is that you might generate a TLS 1.3 ClientHello when resuming a TLS 1.2 sesssion.

draft-ietf-tls-tls13.md
```diff
@@ -1569,11 +1569,13 @@ cipher_suites
...
 legacy_compression_methods
-: Versions of TLS before 1.3 supported compression with the list of
-  supported compression methods being sent in this field. For every TLS 1.3
-  ClientHello, this vector MUST contain exactly one byte set to
-  zero, which corresponds to the "null" compression method in
-  prior versions of TLS. If a TLS 1.3 ClientHello is
+: Versions of TLS before 1.3 supported compression of the plaintext prior to
+  encryption, with the list of supported compression methods being sent
+  in this field.  TLS 1.3 does not support this compression functionality
+  but must send a field that would correspond to the "null" compression
```
martinthomson: who must?  TLS 1.3 clients MUST send a field that corresponds to the "null" compression method in older versions of TLS (that is, a vector containing a single zero octet: 0x0100).

draft-ietf-tls-tls13.md
```diff
@@ -1583,16 +1585,19 @@ legacy_compression_methods
...
 {:br }
 
 In the event that a client requests additional functionality using
 extensions, and this functionality is not supplied by the server, the
 client MAY abort the handshake. Note that TLS 1.3 ClientHello messages
 always contain extensions (minimally they must contain
 "supported_versions" or they will be interpreted as TLS 1.2 ClientHello
-messages). TLS 1.3 servers may receive TLS 1.2 ClientHello messages
-without extensions. If negotiating TLS 1.2, a server MUST check that
+messages). TLS 1.3 servers may receive ClientHello messages from versions
```
martinthomson: may or MAY or could

draft-ietf-tls-tls13.md
```diff
@@ -1691,7 +1696,7 @@ message if they were able to find an acceptable set of algorithms and
 groups that are mutually supported, but
 the client's ClientHello did not contain sufficient information to
 proceed with the handshake.
-If a server cannot successfully select algorithms, it MUST abort
+If a server cannot successfully select algorithms and groups, it MUST abort
```
martinthomson: or groups I think, and what about PSK?

draft-ietf-tls-tls13.md
```diff
@@ -1890,14 +1895,16 @@ Cookies serve two primary purposes:
 When sending a HelloRetryRequest, the server MAY provide a "cookie" extension to the
 client (this is an exception to the usual rule that the only extensions that
 may be sent are those that appear in the ClientHello). When sending the
-new ClientHello, the client MUST echo the value of the extension.
-Clients MUST NOT use cookies in subsequent connections.
+new ClientHello, the client MUST echo the value of the extension and should
```
martinthomson: SHOULD?

draft-ietf-tls-tls13.md
```diff
@@ -1890,14 +1895,16 @@ Cookies serve two primary purposes:
...
+subsequently delete the local copy of the cookie value.
+Clients MUST NOT reuse cookie values in subsequent connections.
 
 
 ###  Signature Algorithms
 
 The client uses the "signature_algorithms" extension to indicate to the server
-which signature algorithms may be used in digital signatures. Clients which
+which signature algorithms may be used in digital signatures over TLS protocol
```
martinthomson: s/may/can/

draft-ietf-tls-tls13.md
```diff
@@ -1976,12 +1983,13 @@ ECDSA algorithms
...
-: Indicates a signature algorithm using RSASSA-PSS {{RFC3447}} with MGF1. The
+: Indicates a signature algorithm using RSASSA-PSS {{RFC3447}} with mask
+  generation function 1. The
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is defined for use with TLS 1.2
-  as well as TLS 1.3.
+  length of the digest output.  The corresponding codepoint is also
+  defined for use with TLS 1.2 as well as with TLS 1.3.
```
martinthomson: I'd drop the "as well as clause"

draft-ietf-tls-tls13.md
```diff
@@ -4930,6 +4937,10 @@ Archives of the list can be found at:
   Facebook \\
   subodh@fb.com
 
+* Benjamin Kaduk \\
+  Akamai \\
+  kaduk@mit.edu
+
```
martinthomson: You should probably put this in a separate PR, it's fairly uncontroversial, whereas the other changes are significant.

draft-ietf-tls-tls13.md
kaduk: Just to clarify: we cannot do this for legacy_session_id because of the stated case, but it is okay for legacy_compression_methods?  (If you send a 1.3 CH, we say these bytes MUST be exactly as listed.)

draft-ietf-tls-tls13.md
kaduk: sure.

draft-ietf-tls-tls13.md
kaduk: Hmm, I might go with "might" or "can" over any of those.

draft-ietf-tls-tls13.md
kaduk: The server's pre_shared_key is a uint16 selected_identity, which would make it 6.
(KeyShareEntry has a NamedGroup and a non-empty key_exchange).

draft-ietf-tls-tls13.md
kaduk: This was just to match with L1695-1696 (new) which uses parallel wording.

draft-ietf-tls-tls13.md
kaduk: It's not related to interoperability, just best practice.  But I would not oppose SHOULD if there is desire to do that.

draft-ietf-tls-tls13.md
kaduk: okay.
(I originally noted this spot on my printout because I had it in my head that it only applied to the TLS protocol structures, but the following text seemed to make it apply to certificates as well.)

draft-ietf-tls-tls13.md
kaduk: Okay.  (The main point here is that the codepoints have different types in 1.2 and 1.3.)

draft-ietf-tls-tls13.md
kaduk: No harm in splitting it off, I suppose.

draft-ietf-tls-tls13.md
martinthomson: I wonder why we don't make this uint16 at the same time (same goes for CipherSuite).

draft-ietf-tls-tls13.md
martinthomson: I think that we can use uppercase here: it's a security requirement and we use normative language for that all the time.  In any case, I was just reminding you that using should/must/may is confusing.

draft-ietf-tls-tls13.md
kaduk: legacy_compression_methods as opaque[2] feels more natural than uint16 to me, as it's just hardcoding the encoding of a specific vector; it's not really a two-byte integer or constant.  (Well, it is a two-byte constant, but that's "coincidence".)  For CipherSuite it seems to make sense, though.

draft-ietf-tls-tls13.md
kaduk: switched to the majuscule in my local copy.

kaduk:
Okay, that should be all the bits for this pull request.  I skipped a few things from my paper notes that require more research, but those can go into a separate pull request (as can the item or two that I promised in my WGLC feedback).

draft-ietf-tls-tls13.md
```diff
@@ -1583,16 +1584,19 @@ legacy_compression_methods
 extensions
 : Clients request extended functionality from servers by sending
   data in the extensions field.  The actual "Extension" format is
-  defined in {{extensions}}.
+  defined in {{extensions}}.  In TLS 1.3, support for and the presence
```
davegarrett: The grammar here is unnecessarily complex, and could be confusing. I suggest simply "In TLS 1.3, use of certain extensions is mandatory, ...".

draft-ietf-tls-tls13.md
```diff
@@ -1583,16 +1584,19 @@ legacy_compression_methods
...
+  of certain extensions is mandatory, as functionality is moved into
+  extensions to preserve ClientHello compatibility with previous versions.
```
davegarrett: nit: suggest "previous versions of TLS"

draft-ietf-tls-tls13.md
```diff
@@ -1670,7 +1674,7 @@ values, it is not possible for an active attacker to modify the
 randoms without detection as long as ephemeral ciphers are used.
 It does not provide downgrade protection when static RSA is used.
 
-Note: This is an update to TLS 1.2 so in practice many TLS 1.2 clients
+Note: This is a change from RFC 5246, so in practice many TLS 1.2 clients
```
davegarrett: please linkify the reference: "{{RFC5246}}" instead of just "RFC 5246"

draft-ietf-tls-tls13.md
```diff
@@ -2077,14 +2084,14 @@ The "extension_data" field of this extension contains a
...
        enum {
            /* Elliptic Curve Groups (ECDHE) */
-           obsolete_RESERVED(1..22),
-           secp256r1(23), secp384r1(24), secp521r1(25),
-           obsolete_RESERVED(26..28),
-           x25519(29), x448(30),
+           obsolete_RESERVED(0x0001..0x0016),
+           secp256r1(0x0017), secp384r1(0x0018), secp521r1(0x0019),
+           obsolete_RESERVED(0x001A..0x001C),
+           x25519(0x001D), x448(0x001E),
```
davegarrett: I think changing this to hex here is counter productive. Likewise for the next block.

draft-ietf-tls-tls13.md
```diff
@@ -3331,7 +3355,7 @@ useful as a traffic analysis countermeasure.
 
        struct {
            ContentType type;
-           ProtocolVersion legacy_record_version = 0x0301;    /* TLS v1.x */
+           ProtocolVersion legacy_record_version;
```
davegarrett: I don't see a need to remove this.

draft-ietf-tls-tls13.md
```diff
@@ -4147,7 +4186,7 @@ MUST meet the following requirements:
  * If not containing a "pre_shared_key" extension, it MUST contain both
    a "signature_algorithms" extension and a "supported_groups" extension.
  * If containing a "supported_groups" extension, it MUST also contain a
-   "key_share" extension, and vice versa. (an empty KeyShare.client_shares
+   "key_share" extension, and vice versa. (An empty KeyShare.client_shares
    vector is permitted)
```
davegarrett: nit: if capitalizing "an", then this parenthetical sentence should also end with a period

draft-ietf-tls-tls13.md
```diff
@@ -4609,6 +4653,9 @@ Old versions of TLS permitted the use of very low strength ciphers.
 Ciphers with a strength less than 112 bits MUST NOT be offered or
 negotiated for any version of TLS for any reason.
 
+The security of SSL 3.0 {{SSL3}} is considered insufficient for the reasons enumerated
+in {{RFC7568}}, and MUST NOT be negotiated for any reason.
+
```
davegarrett: nit: SSL2 should be before SSL3; if you're moving this around, just stick it after the SSL2 bit below.

Alternatively, we could just merge them to simplify things a little bit:
"The security of both SSL 2.0 {{SSL2}} and SSL 3.0 {{SSL3}} are considered insufficient for the reasons enumerated in {{RFC6176}} and {{RFC7568}}, and MUST NOT be negotiated for any reason."

draft-ietf-tls-tls13.md
kaduk: sure

draft-ietf-tls-tls13.md
kaduk: I waffled a bit about linkifying, but I guess it does make more sense to do so than not.

draft-ietf-tls-tls13.md
kaduk: There are all uint16s, and should be presented in a consistent style to be consistent with them being part of the same enumerated type, IMO.

draft-ietf-tls-tls13.md
kaduk: The argument is roughly: if we leave it in, a TLS 1.3 server is justified in enforcing the literal bytes and throwing a decode_error if they don't match.  But that means that it will choke hard on (e.g.) certain TLS 1.1 ClientHellos, when it should really be sending an abort noting that the versions don't match (as appropriate).  At least EKR and davidben were in support of the position that you should not be checking these bytes prior to encryption.

draft-ietf-tls-tls13.md
kaduk: Indeed; not  sure how I missed that.

draft-ietf-tls-tls13.md
kaduk: As I noted in the commit message, the point of the move was to bring sslv2 and sslv3 together while keeping sslv2 and v2-compatible ClientHellos together.  I'm somewhat supportive of combining them, but I think I'll leave it as-is for now and let the editor exercise editorial discretion.

draft-ietf-tls-tls13.md
davidben: If we believe including a value means the receiver is justified in enforcing it (section 3.7 doesn't say much, but I think I agree that's the more natural read), `ClientHello.legacy_version` probably should get the same treatment. It's a little awkward because this is one of the few places where senders always send one value, but receivers must tolerate multiple.

(And, yeah, enforcing the record-layer version prior to encryption is fraught with peril..)

draft-ietf-tls-tls13.md
tomato42: I don't think this is a good idea, it is likely to cause issues on servers that receive client hello messages from TLSv1.2 implementations that do enabled compression methods

kaduk: Hmm, that's probably true.  I had wanted to replace some long-ish running text describing what to send with just a literal value, but this may be too far.

ekr: I agree with tomato42. This seems like it is going to make things more confusing

ekr: I see this is a separate commit.

draft-ietf-tls-tls13.md
ekr: I think we should just use parameters here. I will adapt.

draft-ietf-tls-tls13.md
```diff
@@ -1890,14 +1894,16 @@ Cookies serve two primary purposes:
 When sending a HelloRetryRequest, the server MAY provide a "cookie" extension to the
 client (this is an exception to the usual rule that the only extensions that
 may be sent are those that appear in the ClientHello). When sending the
-new ClientHello, the client MUST echo the value of the extension.
-Clients MUST NOT use cookies in subsequent connections.
+new ClientHello, the client MUST echo the value of the extension and SHOULD
+subsequently delete the local copy of the cookie value.
+Clients MUST NOT reuse cookie values in subsequent connections.
```
ekr: What's the point of deleting the cookie? It's not secret because it just went over the wire.

draft-ietf-tls-tls13.md
```diff
@@ -1890,14 +1894,16 @@ Cookies serve two primary purposes:
...
 
 
 ###  Signature Algorithms
 
 The client uses the "signature_algorithms" extension to indicate to the server
-which signature algorithms may be used in digital signatures. Clients which
+which signature algorithms can be used in digital signatures over TLS protocol
+structures and digital signatures of X.509 certificates. Clients which
```
ekr: This seems kinda rdundant.

draft-ietf-tls-tls13.md
```diff
@@ -1976,12 +1982,13 @@ ECDSA algorithms
...
 RSASSA-PSS algorithms
-: Indicates a signature algorithm using RSASSA-PSS {{RFC3447}} with MGF1. The
+: Indicates a signature algorithm using RSASSA-PSS {{RFC3447}} with mask
+  generation function 1. The
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is defined for use with TLS 1.2
-  as well as TLS 1.3.
+  length of the digest output.  The corresponding codepoint is also
```
ekr: It's not corresponding, it's the same.

draft-ietf-tls-tls13.md
```diff
@@ -2205,16 +2211,17 @@ these checks fails, then the client MUST abort the handshake with an
...
 
 If using (EC)DHE key establishment, servers offer exactly one
 KeyShareEntry in the ServerHello. This value MUST correspond to the KeyShareEntry value offered
 by the client that the server has selected for the negotiated key exchange.
 Servers MUST NOT send a KeyShareEntry for any group not
 indicated in the "supported_groups" extension.
-If a HelloRetryRequest was received, the client MUST verify that the
-selected NamedGroup matches that supplied in the selected_group field and MUST
-abort the handshake with an "illegal_parameter" alert if it does not.
+Servers MUST NOT send a KeyShareEntry when using the "psk_ke" PskKeyExchangeMode.
```
ekr: Note to self. Merge these two sentences.

draft-ietf-tls-tls13.md
```diff
@@ -2351,9 +2359,9 @@ identity.
 Clients MUST verify that the server's selected_identity is within the
 range supplied by the client, that the server selected the cipher
 suite associated with the PSK,
-and that the "key_share", and
-"signature_algorithms" extensions are consistent with the indicated
-ke_modes and auth_modes values. If these values are not consistent,
+and that the "key_share" extension in the ServerHello is consistent
+with the ke_modes offered in the ClientHello psk_key_exchange_modes
+extension.  If these values are not consistent,
```
ekr: I actually don't think we should be requiring the client to check for KE not being present. It's just an unnecessary check.

draft-ietf-tls-tls13.md
```diff
@@ -2640,7 +2649,9 @@ certificate_request_context
   of this connection (thus preventing replay of client
   CertificateVerify messages). This field SHALL be zero length
   unless used for the post-handshake authentication exchanges
-  described in {{post-handshake-authentication}}.
+  described in {{post-handshake-authentication}}.  This does not
+  affect the uniqueness within the connection, since there will
+  only be one CertificateRequest during the initial handshake.
```
ekr: This seems unnecessary

draft-ietf-tls-tls13.md
```diff
@@ -2817,32 +2831,32 @@ extensions:
 : A set of extension values for the CertificateEntry. The "Extension"
   format is defined in {{extensions}}. Valid extensions include
   OCSP Status extensions ({{!RFC6066}} and {{!RFC6961}}) and
-  SignedCertificateTimestamps ({{!RFC6962}}).  Any extension presented
-  in a Certificate message must only be presented if the corresponding
+  SignedCertificateTimestamps ({{!RFC6962}}).  An extension MUST only be presented
+  in a Certificate message if the corresponding
```
ekr: Good. I'm also going to change presented -> present

draft-ietf-tls-tls13.md
```diff
@@ -3124,14 +3141,17 @@ first.
...
 
 Note: Although the resumption master secret depends on the client's second
 flight, servers which do not request client authentication MAY compute
 the remainder of the transcript independently and then send a
 NewSessionTicket immediately upon sending its Finished rather than
-waiting for the client Finished.
+waiting for the client Finished.  This might be appropriate in cases
+where the client is expected to open multiple TLS connections in
+parallel and would benefit from the reduced overhead of a resumption
+handshake, for example.
```
ekr: Thanks for this text. We had a sketchier reason for this, so it's nice to have a really solid one. Also, this is a really nice performance benefit of TLS 1.3 over TLS 1.2 with H1, b/c in TLS 1.2 the NST comes in the server's second flight, even with False Start.

draft-ietf-tls-tls13.md
```diff
@@ -3164,8 +3184,11 @@ ticket_age_add
 ticket
 : The value of the ticket to be used as the PSK identity.
 The ticket itself is an opaque label. It MAY either be a database
-lookup key or a self-encrypted and self-authenticated value. Section
-4 of {{RFC5077}} describes a recommended ticket construction mechanism.
+lookup key or a self-encrypted and self-authenticated value. It
+MUST NOT be unprotected plaintext that can be modified by the
+recipient or an attacker without being rejected when presented back
+to the server.  Section 4 of {{RFC5077}} describes a recommended
+ticket construction mechanism.
```
ekr: I'm not sure this is correct. Consider the case where it's a database key and we just issue them in sequence. Is that unsafe?

draft-ietf-tls-tls13.md
```diff
@@ -3226,16 +3249,16 @@ the responses).
...
 {:br }
 
 The KeyUpdate handshake message is used to indicate that the sender is
 updating its sending cryptographic keys. This message can be sent by
-the server after sending its first flight and the client after sending
-its second flight. Implementations that receive a KeyUpdate message
-prior to receiving a Finished message as part of the 1-RTT handshake
+either peer after it has sent a Finished message.
+Implementations that receive a KeyUpdate message
+prior to receiving a Finished message
```
ekr: Nice simplification

draft-ietf-tls-tls13.md
ekr: I think @kaduk is right here, unfortunately.

draft-ietf-tls-tls13.md
```diff
@@ -4147,8 +4186,8 @@ MUST meet the following requirements:
  * If not containing a "pre_shared_key" extension, it MUST contain both
    a "signature_algorithms" extension and a "supported_groups" extension.
  * If containing a "supported_groups" extension, it MUST also contain a
-   "key_share" extension, and vice versa. (an empty KeyShare.client_shares
-   vector is permitted)
+   "key_share" extension, and vice versa. (An empty KeyShare.client_shares
```
ekr: Ugh, a whole sentence in parens. Note to self, fix on merge.

ekr:
Merged by hand.

draft-ietf-tls-tls13.md
kaduk: Well, my notes just said "provide a justification/example", so I had to come up with something.  Now I want to know what the other reason for it was... ;)

ekr: Stateless reject in cases where you wanted to force a RT for anti-DoS (e.g., QUIC or DTLS).

You complete the handshake, send an NST, and then don't wait for Finished, using the Ticket as a state token.

PR #812, opened 2016-12-13
Change end_of_early_data into a handshake message. This makes all

ekr:
key changes happen at Handshake boundaries.

Suggested by David Benjamin

PR #814, opened 2016-12-13
Expand the handshake transcript at least once

martinthomson:
This should help clarify things.  Assumes #812 is merged.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -146,7 +146,7 @@ informative:
            ins: E. Barker
          -
            ins: Lily Chen
-         - 
+         -
            ins: A. Roginsky
          -
            ins: M. Smid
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3974,6 +3974,13 @@ even with the same secret. In a 0-RTT exchange, Derive-Secret is
 called with four distinct transcripts; in a 1-RTT only exchange
 with three distinct transcripts.
 
+The complete transcript passed to Derive-Secret is always taken from the
+following sequence of handshake messages, starting at the first ClientHello and
+including only those messages that were sent: ClientHello, HelloRetryRequest,
+ClientHello, ServerHello, EncryptedExtensions, Server CertificateRequest, Server
+Certificate, Server CertificateVerify, Server Finished, EndOfEarlyData, Client
```
kaduk: And there is no problem with the client sending an EndOfEarlyData that the server can't decrypt [to include in its transcript] because the client MUST NOT send one if the server rejects early data.  But do we need to mention that explicitly here?

ekr:
Rebased and merged.

PR #815, opened 2016-12-13
Using "identity" instead of "label".

kazu-yamamoto:
"label" is a term for derive-key.
We should avoid using "label" for PSK.


ekr:
I don't think this is necessary. Label is a common term and the scope is clear enough

kazu-yamamoto:
OK. Let's close.

PR #816, opened 2016-12-13
clarifying what PSK is in key schedule.

kazu-yamamoto:
I needed to guess what PSK is in this diagram when I implemented PSK handshake.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3893,7 +3893,7 @@ etc.  The initial secret is simply a string of zeroes as long as the size
 of the Hash that is the basis for the HKDF. Concretely, for the
 present version of TLS 1.3, secrets are added in the following order:
 
-- PSK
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3893,7 +3893,7 @@ etc.  The initial secret is simply a string of zeroes as long as the size
 of the Hash that is the basis for the HKDF. Concretely, for the
 present version of TLS 1.3, secrets are added in the following order:
 
-- PSK
+- PSK (a pre-shared key established externally or resumption_secret).
```
mimoo: I don't think this is good if we remove `resumption_secret` as @ekr said [here](https://www.ietf.org/mail-archive/web/tls/current/msg22204.html)

> This is actually called the resumption master
secret. I'll clean this up.

kazu-yamamoto: Yes. We should take care of consistency.

PR #818, opened 2016-12-13
Consolidate ticket_early_data_info and early_data into a single

ekr:
extension, as suggested by Steven Valdez/David Benjamin.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2456,7 +2456,7 @@ The "extension_data" field of this extension contains an
                   uint32 max_early_data_size;
 
                case client_hello:
-               case hello_retry_request:
+               case encrypted_extensions:
                   // empty
            };
        } EarlyDataIndication;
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2444,13 +2446,20 @@ supply an "early_data" extension as well as the "pre_shared_key"
...
 
 %%% Key Exchange Messages
 
        struct {
+           select (Handshake.msg_type) {
+               case client_hello:
+               case hello_retry_request:
+               case new_session_ticket:
+                  uint32 max_early_data_size;
        } EarlyDataIndication;
```
kazu-yamamoto: Is "}" missing?

kazu-yamamoto: This is fall through. So, `uint32` is contained in `client_hello`. But the original extension is empty. How should we use `uint32` for `client_hello`?

draft-ietf-tls-tls13.md
```diff
@@ -2454,6 +2454,7 @@ The "extension_data" field of this extension contains an
            select (Handshake.msg_type) {
                case client_hello:
                case hello_retry_request:
+                  // empty.
```
kazu-yamamoto: According to our old discussion, I think that `struct {}` is correct.

ekr: Ugh. I re-ordered instead.

kaduk:
LGTM

PR #820, opened 2016-12-15
Edits from https://www.ietf.org/mail-archive/web/tls/current/msg22203.html

mimoo:


draft-ietf-tls-tls13.md
```diff
@@ -1566,7 +1566,9 @@ cipher_suites
   order of client preference. If the list contains cipher suites
   the server does not recognize, support, or wish to use, the server
   MUST ignore those cipher suites, and process the remaining ones as
-  usual. Values are defined in {{cipher-suites}}.
+  usual. Values are defined in {{cipher-suites}}. If the client is
+  attempting a PSK key establishment, it MUST only advertise the single
+  cipher suite associated with the PSK.
```
ekr: This actually isn't correct. You MUST advertise it, but it's not right to require that you MUST only advertise it because this makes it impossible for servers to disable any cipher while a ticket is still valid because it might be in use by some client.

ekr: The server is responsible for only picking PSK if it also selects a compatible cipher suite. So the situation you describe is a bug in the server. The critical concern is that the handshake has to complete successfully if at all possible. If you compute the wrong key, get a MAC error and retry, then you have problems.

Consider the case where the client offers A and B, and the server selects A. If the client's attempt to resume only includes A, then if the server has disabled A, then you will have a negotiation failure even if you could have connected. By contrast, if the client offers both A and B but the PSK can only be used with A, then if the server has disabled B then you just get no resumption but B.

ekr:
On Wed, Dec 14, 2016 at 10:31 PM, David Benjamin <notifications@github.com>
wrote:

> *@davidben* commented on this pull request.
> ------------------------------
>
> In draft-ietf-tls-tls13.md <https://github.com/tlswg/tls13-spec/pull/820>:
>
> > @@ -1566,7 +1566,9 @@ cipher_suites
>    order of client preference. If the list contains cipher suites
>    the server does not recognize, support, or wish to use, the server
>    MUST ignore those cipher suites, and process the remaining ones as
> -  usual. Values are defined in {{cipher-suites}}.
> +  usual. Values are defined in {{cipher-suites}}. If the client is
> +  attempting a PSK key establishment, it MUST only advertise the single
> +  cipher suite associated with the PSK.
>
> Now that the PSK / cipher matching is weakened to just the PRF hash
> needing to match, can we even say that the client MUST offer "the" cipher
> for the PSK? "The" cipher isn't well-defined. Agreed with EKR that the
> server is responsible for making the handshake succeed, even if that means
> declining the PSK. Anything else will make TLS 1.3 a deployment hazard. (It
> must never be the case that statefulness like session resumption prevents
> servers from rolling back a configuration change.)
>
> Our implementation, for instance, first negotiates the cipher
> independently of the resumption offer, then it only accepts the resumption
> if the cipher suite it chose matches the session. That means we don't
> actually care if the client offered ciphers that match the session.
>
This is how NSS behaves as well.


> Say we fixed it to say the client MUST offer at least one cipher suite
> with a matching PRF. A server cannot take advantage of this. If it sees no
> PRF-matching ciphers, maybe this is just because the only matching ones
> were ciphers it didn't know about. If this rule is unenforceable and can't
> be relied on, it's better to leave it alone.
>

This seems like a reasonable argument.

-Ekr

draft-ietf-tls-tls13.md
mimoo: @ekr: are you saying that the server should understand that the very first cipher suite, in the client hello's cipher suites list, is the one to be used with the PSK? And that if the server does not support that first choice, it should then dismiss the PSK and continue with whatever other choice of cipher suite exist? I do not see this in the spec.

mimoo: @davidben: true, only the hash part needs to be the one associated to the PSK.

@ekr: from the spec: "Prior to accepting PSK key establishment, the server MUST validate the corresponding binder value (see Section 4.2.6.1 below). If this value is not present or does not validate, the server MUST abort the handshake."

ekr: I don't think I'm quite saying that. What the spec says is that you can't resume without the cipher suite having the same PRF:

"Any ticket MUST only be resumed with a cipher suite that has the same KDF hash as that used to establish the original connection, and if the client provides the same SNI value as described in Section 3 of [RFC6066]."

It's up to the server to implement this requirement however it chooses. As @davidben says, the way that NSS and BoringSSL implement this is to negotiate the cipher suite first (using whatever their conventional mechanism is, which may be server's preference, client's preference, or something else) and then condition resumption on whether the PSK has the appropriate KDF. Note that this cipher suite might be the client's last choice if, for instance, it's the server's first choice.

ekr: "@ekr: from the spec: "Prior to accepting PSK key establishment, the server MUST validate the corresponding binder value (see Section 4.2.6.1 below). If this value is not present or does not validate, the server MUST abort the handshake."

this contradicts your scenario."

Hm.... I don't think so... Here's one order of operations.

1. Negotiate the cipher suite.
2. If a PSK is present, check to see that it has the same hash. If not, do a full handshake.
3. If a PSK is present and has the same hash, check the binder. If that fails, abort the handshake. If it succeeds, accept the PSK.

The key point here is that you don't get a mismatch in the binder unless something is really wrong (just like with the Finished MAC) and then you should just generate an error, not try to fall back to a non-PSK handshake.

mimoo: in your 2., you presume that the server knows what hash is associated with a PSK. Is that always the case?

davidben: Yes. This is a requirement for TLS 1.3 PSK, just like it is typically a requirement that servers know whether this blob of bytes is a ECDSA P-256 private key an RSA private key.

mimoo: OK I understand this better. I did not know that. Also reading the spec it's hard to see it.

From the [New Session Ticket](https://tlswg.github.io/tls13-spec/#NewSessionTicket) section:

> The value of the ticket to be used as the PSK identity. The ticket itself is an opaque label. It MAY either be a database lookup key or a self-encrypted and self-authenticated value. Section 4 of [RFC5077] describes a recommended ticket construction mechanism. 

Note here the usage of "MAY" as well as "recommended".

There is also nothing in the current draft that talks about how the server should do that: "If a PSK is present, check to see that it has the same hash. If not, do a full handshake."

From the [Pre Shared Key extension](https://tlswg.github.io/tls13-spec/#pre-shared-key-extension) section, I think this is one more mention to the "cipher suite associated with the PSK" that could be changed to something along the lines of "a cipher suite with a hash function associated with the PSK".

> Clients MUST verify that the server’s selected_identity is within the range supplied by the client, that the server selected the cipher suite associated with the PSK

mimoo: Alright, found this from the spec that clarifies this part for me:

> Each PSK is associated with a single Hash algorithm. For PSKs established via the ticket mechanism ({{NewSessionTicket}}), this is the Hash used for the KDF. For externally established PSKs, the Hash algorithm MUST be set when the PSK is established.

mimoo:
added the changes previously discussed.

draft-ietf-tls-tls13.md
```diff
@@ -1567,8 +1567,8 @@ cipher_suites
   the server does not recognize, support, or wish to use, the server
   MUST ignore those cipher suites, and process the remaining ones as
   usual. Values are defined in {{cipher-suites}}. If the client is
-  attempting a PSK key establishment, it MUST only advertise the single
-  cipher suite associated with the PSK.
+  attempting a PSK key establishment, it MUST advertise at least one
+  cipher suite containing a Hash associated with the PSK.
```
mimoo: It's an asymmetric MUST. Only the Client needs to enforce it.

davidben: Weren't you objecting in your thread that asymmetric MUSTs are confusing? It seems this particular asymmetric MUST doesn't serve much purpose and only risks encouraging servers towards an unsound decision algorithm.

Hash vs ciphers was an intentional change to align the rules for external PSKs (where expecting an externally-configured AEAD is a bit much) and ticket-based PSKs. And, yes, it means 0-RTT has stricter requirements than resumption. That is already the case, and necessarily so, because 0-RTT cannot change ALPN or basically any property at all. One reasonable selection algorithm is:

1. Decide if you were going to resume or not.
2. Select all other parameters. ALPN, etc.
3. If all of the following are true, acknowledge 0-RTT. Otherwise, decline it.
   - We are resuming.
   - The client offered 0-RTT and we are willing to accept it.
   - The PSK is okay for use with 0-RTT.
   - All connection parameters associated with the PSK match those selected organically.

mimoo: > Weren't you objecting in your thread that asymmetric MUSTs are confusing?

I did but ekr's justification seems fine that I changed my opinion on it :) If you think this is too much I could change it for a SHOULD. What do you think?

> 0-RTT has stricter requirements than resumption

I see. Thanks for the explanations!

davidben: I think I would rather it not be mentioned. On a server, the text should be ignored. On a client, it actually should be ignored too. Either you keep all your sessions in-memory, in which case you necessarily satisfy this and send the cipher list you would have sent anyway. Or you have some persistence and might offer a session from a prior configuration, in which case you probably ought to avoid offering that session but if you did, you certainly would still send the cipher list you would have sent anyway and not pick an arbitrary hash-matching cipher.

Moreover, having clients vary their preferences based on a stateful thing (session resumption) is very dangerous and risks turning TLS 1.3 into a deployment hazard. Deploying a server change and then rolling it back should be as close to a no-op as possible.

mimoo: > On a client, it actually should be ignored too. Either you keep all your sessions in-memory, in which case you necessarily satisfy this and send the cipher list you would have sent anyway.

Hmm... I think this situation doesn't encompass out-of-bands PSKs, or some other non-web scenarios.

davidben: Hrm. Perhaps there should be some text in 4.2.6 then, as guidance when picking parameters? If you're doing out-of-band PSKs where there isn't a cert-based default, you actually to not just say "please include some ciphers that match the PSK" but that "the only ciphers that will work are ones that match the PSK, so avoid saying silly things like AES_256_GCM_SHA384 when you don't have a cert-based default and all your PSKs are SHA-256 ones".

mimoo: That section is already pretty loaded though. I still think the mention would be better in that cipher suite definition of the client hello. I tried a change to a SHOULD, if this is really not satisfying I can revert this commit.

kaduk:
I did not verify that the moved text is only move and no edits, but the other chunks seem okay.

ekr:
@davidben are you OK with this SHOULD?

davidben:
Sounds reasonable. Mostly I just don't want to encourage servers to enforce anything about this.

PR #821, opened 2016-12-15
editorial cleanup

cem-:
I believe these are all non-controversial, editorial changes.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -840,7 +840,7 @@ Auth | {CertificateVerify*}
                  messages/extensions that are not always sent.
 
               {} Indicates messages protected using keys
-                 derived from handshake_traffic_secret.
+                 derived from a [sender]_handshake_traffic_secret.
 
               [] Indicates messages protected using keys
                  derived from traffic_secret_N
@@ -985,7 +985,7 @@ can then use that PSK identity in future handshakes to negotiate use
 of the PSK. If the server accepts it, then the security context of the
 new connection is tied to the original connection and the key derived
 from the initial handshake is used to bootstrap the cryptographic state
-instead of a full hanshake. In TLS 1.2 and
+instead of a full handshake. In TLS 1.2 and
 below, this functionality was provided by "session IDs" and
 "session tickets" {{RFC5077}}. Both mechanisms are obsoleted in TLS
 1.3.
@@ -1036,9 +1036,10 @@ Subsequent Handshake:
 
 As the server is authenticating via a PSK, it does not send a
 Certificate or a CertificateVerify message. When a client offers resumption
-via PSK, it SHOULD also supply a "key_share" extension to the server as well
-to allow the server to decline resumption and fall back to a full handshake,
-if needed. The server responds with a "pre_shared_key" extension
+via PSK, it SHOULD also supply a "key_share" extension to the server (to provide
+forward secrecy) as well as allow the server to decline resumption and fall back
+to a full handshake, if needed. The server responds with a "pre_shared_key"
+extension to negotiate use of PSK key establishment and can (as shown here)
 to negotiate use of PSK key establishment and can (as shown here)
 respond with a "key_share" extension to do (EC)DHE key
 establishment, thus providing forward secrecy.
@@ -1091,7 +1092,7 @@ as with a 1-RTT handshake with PSK resumption.
                   derived from client_early_traffic_secret.
 
                {} Indicates messages protected using keys
-                  derived from handshake_traffic_secret.
+                  derived from a [sender]_handshake_traffic_secret.
 
                [] Indicates messages protected using keys
                   derived from traffic_secret_N
@@ -1459,7 +1460,7 @@ non-PSK case discussed in the previous paragraph.
 The server indicates its selected parameters in the ServerHello as
 follows:
 
-- If PSK is being used then the server will send a
+- If PSK is being used, then the server will send a
 "pre_shared_key" extension indicating the selected key.
 - If PSK is not being used, then (EC)DHE and certificate-based
 authentication are always used.
@@ -1667,7 +1668,7 @@ protection against downgrade attacks over and above that provided
 by the Finished exchange: because the ServerKeyExchange, a message
 present in TLS 1.2 and below, includes a signature over both random
 values, it is not possible for an active attacker to modify the
-randoms without detection as long as ephemeral ciphers are used.
+random values without detection as long as ephemeral ciphers are used.
 It does not provide downgrade protection when static RSA is used.
 
 Note: This is an update to TLS 1.2 so in practice many TLS 1.2 clients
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4119,7 +4122,7 @@ CertificateVerify and certificates), and ecdsa_secp256r1_sha256. A
 TLS-compliant application MUST support key exchange with secp256r1
 (NIST P-256) and SHOULD support key exchange with X25519 {{RFC7748}}.
 
-##  MTI Extensions
+##  Mandatory-to-Implement Extensions
```
cem-: Changing this header breaks the ref on line 1907 (and is why the travis-ci broke). [So that will need fixing.]

martinthomson: One way to fix that is to add `{#mti-extensions}` to the end of this line.

draft-ietf-tls-tls13.md
```diff
@@ -1459,7 +1460,7 @@ non-PSK case discussed in the previous paragraph.
 The server indicates its selected parameters in the ServerHello as
 follows:
 
-- If PSK is being used then the server will send a
+- If PSK is being used, then the server will send a
```
kaduk: I find that the RFC Editor staff have a different idea of when commas are appropriate than I do, so I try to restrain myself from making too many changes of that nature.  (Just FYI; I would probably add this if it was my document.)

draft-ietf-tls-tls13.md
```diff
@@ -1884,7 +1885,7 @@ Cookies serve two primary purposes:
 
 - Allowing the server to offload state to the client, thus allowing it to send
   a HelloRetryRequest without storing any state. The server does this by
-  pickling that post-ClientHello hash state into the cookie (protected
+  setting the post-ClientHello hash state as the cookie (protected
```
kaduk: I don't think this is quite right, as the server can (and should) put other state into the cooke as well as just the hash state.  "pickling" is perhaps a needless term of art, though, and "including [state] in the cookie" would do just fine.

draft-ietf-tls-tls13.md
```diff
@@ -1901,7 +1902,7 @@ which signature algorithms may be used in digital signatures. Clients which
 desire the server to authenticate itself via a certificate MUST send this extension.
 If a server
 is authenticating via a certificate and the client has not sent a
-"signature_algorithms" extension then the server MUST
+"signature_algorithms" extension then, the server MUST
 abort the handshake with a "missing_extension" alert
```
kaduk: This comma, though, I disagree with.  (Maybe it was intended to be before "then" and not after?)

draft-ietf-tls-tls13.md
```diff
@@ -3520,7 +3521,9 @@ Application Data records may contain a zero-length TLSInnerPlaintext.content if
 the sender desires.  This permits generation of plausibly-sized cover
 traffic in contexts where the presence or absence of activity may be
 sensitive.  Implementations MUST NOT send Handshake or Alert records
-that have a zero-length TLSInnerPlaintext.content.
+that have a zero-length TLSInnerPlaintext.content, if such a message
+is recieved, the receiving implementation MUST terminate the connection
+with an "unexpected_message" alert.
```
kaduk: This adds a comma splice.  A new sentence would be fine.

draft-ietf-tls-tls13.md
```diff
@@ -4423,7 +4426,7 @@ TLS protocol issues:
   (see {{backward-compatibility}})
 
 -  Do you handle TLS extensions in ClientHello correctly, including
-  unknown extensions.
+  unknown extensions?
```
kaduk: I have this in #810 too, but leaving it here shouldn't cause any issues.

draft-ietf-tls-tls13.md
```diff
@@ -1036,9 +1036,10 @@ Subsequent Handshake:
 
 As the server is authenticating via a PSK, it does not send a
 Certificate or a CertificateVerify message. When a client offers resumption
-via PSK, it SHOULD also supply a "key_share" extension to the server as well
-to allow the server to decline resumption and fall back to a full handshake,
-if needed. The server responds with a "pre_shared_key" extension
+via PSK, it SHOULD also supply a "key_share" extension to the server (to provide
+forward secrecy) as well as allow the server to decline resumption and fall back
```
ekr: This seems unduly prescriptive. It's up to people whether they want to do a DH exchange here. The reason to provide a key is to allow the full handshake.

draft-ietf-tls-tls13.md
```diff
@@ -2045,8 +2046,8 @@ CertificateAuthoritiesExtension structure.
 authorities
 : A list of the distinguished names {{X501}} of acceptable
   certificate authorities, represented in DER-encoded {{X690}} format.  These
-  distinguished names specify a desired distinguished name for a
-  root CA or for a subordinate CA; thus, this message can be used to
+  distinguished names specify a desired distinguished name for a root CA,
+  subordinate CA, or trust anchor; thus, this message can be used to
```
ekr: root CA and trust anchor are the same thing in this context.

draft-ietf-tls-tls13.md
```diff
@@ -3520,7 +3521,9 @@ Application Data records may contain a zero-length TLSInnerPlaintext.content if
 the sender desires.  This permits generation of plausibly-sized cover
 traffic in contexts where the presence or absence of activity may be
 sensitive.  Implementations MUST NOT send Handshake or Alert records
-that have a zero-length TLSInnerPlaintext.content.
+that have a zero-length TLSInnerPlaintext.content, if such a message
```
ekr: content; if

draft-ietf-tls-tls13.md
```diff
@@ -3520,7 +3521,9 @@ Application Data records may contain a zero-length TLSInnerPlaintext.content if
...
+is recieved, the receiving implementation MUST terminate the connection
```
ekr: i before e, except after c

draft-ietf-tls-tls13.md
```diff
@@ -4389,8 +4392,8 @@ unsatisfactory, {{RFC4086}} provides guidance on the generation of random values
 Implementations are responsible for verifying the integrity of certificates and
 should generally support certificate revocation messages. Certificates should
 always be verified to ensure proper signing by a trusted Certificate Authority
-(CA). The selection and addition of trusted CAs should be done very carefully.
-Users should be able to view information about the certificate and root CA.
+(CA). The selection and addition of trusted anchors should be done very carefully.
```
ekr: trusted -> trust

draft-ietf-tls-tls13.md
ekr: I ended up using serialize

ekr:
Rebased and merged. @cem- please provide a PR adding yourself to contributors in the appropriate location

PR #822, opened 2016-12-15
better explanations of the PSK key establishment process

mimoo:
After the discussions of https://github.com/tlswg/tls13-spec/pull/820 I took a shot at clarifying this section a bit.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2469,24 +2469,40 @@ selected_identity
 
 Each PSK is associated with a single Hash algorithm. For PSKs established
 via the ticket mechanism ({{NewSessionTicket}}), this is the Hash used for
-the KDF. For externally established PSKs, the Hash algorithm MUST be set when the
-PSK is established.
+the KDF. For externally established PSKs, the Hash algorithm MUST be
+decided during the establishment of the PSK.
 
-Prior to accepting PSK key establishment, the server MUST validate the
-corresponding binder value (see {{psk-binder}} below). If this value is
-not present or does not validate, the server MUST abort the handshake.
-Servers SHOULD NOT attempt to validate multiple binders; rather they
-SHOULD select a single PSK and validate solely the binder that
-corresponds to that PSK. In order to accept PSK key establishment, the
-server sends a "pre_shared_key" extension indicating the selected
-identity.
+In order to attempt a PSK key establishment, Clients SHOULD advertise at least one
+cipher suite containing a Hash associated with the PSK.
 
-Clients MUST verify that the server's selected_identity is within the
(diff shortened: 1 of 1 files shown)


mimoo:
After the discussions of https://github.com/tlswg/tls13-spec/pull/820 I took a shot at clarifying this section a bit.

draft-ietf-tls-tls13.md
```diff
@@ -2469,24 +2469,40 @@ selected_identity
...
+If the associated binder value (see {{psk-binder}} below) is not present or does
+not validate, the Server MUST abort the handshake. Servers SHOULD NOT attempt
+to validate multiple binders.
+
+If the binder validates, to finally accept the PSK key establishment, Servers MUST
+send a "pre_shared_key" extension indicating the selected identity.
+
+Note that this process is more involved if the Client wishes to use the early_data
+extension. Not only the Hash function should match, but the entire cipher suite as well.
+Other extensions can also define an early_data requirement to match.
```
ekr: I'm not sure we really need to be prescriptive here. There's at least one other algorithm:

1. Select PSK if possible.
2. Check to see if the PSK can be used with any client-offered cipher suite and if so do PSK. If not, go do regular negotiation.

This isn't what NSS or BoringSSL does, but I'm not aware of any strong argument for why it's a sufficiently bad algorithm that we should be telling people not to use it.

davidben: One argument not to do this is, unlike TLS 1.2, TLS 1.3 ticket renewal admits longer lifetimes. With dhe_psk, tickets are less scary and renewal actually refreshes the key. There's still the authentication freshness lifetime cliff, but if we later add a renewal mode that refreshes authentication, it'll be possible to maintain continuous resumption.

Now suppose today I prefer AES_128_GCM_SHA256 > AES_256_GCM_SHA384 and clients support both. All my tickets will mint AES_128_GCM_SHA256 tickets and everything's lined up for continuous resumption. Now I flip my preferences to AES_256_GCM_SHA384 > AES_128_GCM_SHA256 (or deploy a new cipher or whatever). New clients will get the new preferences, but clients whose renewal train started earlier never rotate to the new settings. You can expect these kinds of pref changes to be less common, so most servers should find it preferable to miss resumption once and rotate everyone to the new settings.

We didn't have this problem in TLS 1.2 because no one does 1.2 ticket renewal and, since it didn't change the master secret, ticket lifetime should not have extended.

It's best not to be prescriptive since implementations may have different priorities, but some guidance about this issue may be worthwhile. I think this issue isn't immediately obvious.

ekr:
@mimoo: I added some implementation advice in https://github.com/tlswg/tls13-spec/commit/c65f98738af29aaacfff140d6a7fa37db744ca3d

PR #830, opened 2016-12-19
Reword the exporters section for clarity

### LINKED ISSUE #829: Is it true that the exporter interface is the same?
In previous versions of TLS, passing no context was different to passing an empty context.  TLS 1.3 doesn't distinguish between these.  I don't know how important it is to highlight that minor difference, but it's probably worth mentioning.

martinthomson:
I found the wording a little hard to follow.  This also fixes #829 by being very explicit about the presence of the change.


davegarrett:
This changeset includes some ".targets.mk" cruft that isn't part of a reword.

PR #831, opened 2016-12-19
Quiesce build warnings

martinthomson:
Most of these are simple enough, kramdown treats all bracketed text as a reference.  So these can be replaced with `\[` and the warnings go away.  The biggest change here is slightly stylistic: a few places had `[{{server-hello}}]`, which kramdown complained about, but rendered correctly.  However, this produced a non-sentence.

Also, I fixed some inconsistent use of references and s/data/date in one of the references.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4089,8 +4089,8 @@ The traffic keying material is generated from the following input values:
 The traffic keying material is generated from an input traffic secret value using:
 
 ~~~~
-    \[sender]_write_key = HKDF-Expand-Label(Secret, "key", "", key_length)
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -912,19 +912,19 @@ Certificate:
...
-  certificate. [{{certificate-verify}}]
+  certificate. See {{certificate-verify}}.
 
 Finished:
 : a MAC (Message Authentication Code) over the entire handshake.
   This message provides key confirmation, binds the endpoint's identity
   to the exchanged keys, and in PSK mode
-  also authenticates the handshake. [{{finished}}]
+  also authenticates the handshake. See {{finished}}.
 {:br }
```
ekr: These are actually intentional. I don't think you need See here.

PR #832, opened 2016-12-19
adding mimoo to contributors

mimoo:


ekr:
Thanks for remembering to do this. I had dropped the state on the theory you would retransmit :)

mimoo:
awesome thanks :)

PR #837, opened 2016-12-28
Corrections for forward&backward secrecy

jensguballa:
- E.2 Record Layer:
  KeyUpdate provides backward secrecy, (EC)DHE provides forward secrecy

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4942,7 +4942,7 @@ Length concealment.
 : Given a record with a given external length, the attacker should not be able
 to determine the amount of the record that is content versus padding.
 
-Forward security after key change.
+Backward security after key change.
 : If the traffic key update mechanism described in {{key-update}} has been
 used and the previous generation key is deleted, an attacker who compromises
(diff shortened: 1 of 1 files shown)


mimoo:
You got it backward, the specification is correct here. Updating your keys provide forward secrecy but does not provide backward secrecy.

katrielalex:
Shameless author-plug: there is a lot of confusion about forward/backward/future secrecy, in part because the naming scheme is not super intuitive. We wrote [_On Post-Compromise Security_](http://ia.cr/2016/221) (CSF '16) to try to clear up some of the confusion.

ekr:
@katrielalex: perhaps you would like to suggest some new text here? Thats what you get for shamelessly-plugging :)

ekr:
@katrielalex: can we expect some text?

katrielalex:
Sorry, holidays and RWC got in the way. Bumped up the stack, will get something to you asap.

katrielalex:
ETA a day or two, am just conferring with the group

jensguballa:
I think this PR is obsoleted now by #867.

PR #838, opened 2016-12-28
Certificate verify

jensguballa:
The client should be required to verify the CertificateVerify message. In addition I prefer to specify the alert type in case the verification fails (also for the Finished).

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3105,6 +3105,10 @@ CertificateVerify. All SHA-1 signature algorithms in this specification are
 defined solely for use in legacy certificates, and are not valid for
 CertificateVerify signatures.
 
+The receiver of a CertificateVerify MUST verify the signature. If the 
+verification fails, the connection MUST be terminated with an 
+"illegal_parameter" alert.
+
 Note: When used with non-certificate-based handshakes (e.g., PSK), the
 client's signature does not cover the server's certificate directly.
 When the PSK was established through a NewSessionTicket, the client's
@@ -3123,7 +3127,10 @@ block. It is essential for providing authentication of the handshake
 and of the computed keys.
 
 Recipients of Finished messages MUST verify that the contents are
-correct. Once a side has sent its Finished message and received and
+correct. If the verification fails, the connection MUST be terminated
+with an "illegal_parameter" alert. 
+
+Once a side has sent its Finished message and received and
 validated the Finished message from its peer, it may begin to send and
 receive application data over the connection.
 Early data may be sent prior to the receipt of the peer's Finished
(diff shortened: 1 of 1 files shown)


ekr:
@jensguballa can you please rebase this?

jensguballa:
I tried my best to rebase the PR. Not sure though I did it the right way. @ekr, can you provide feedback?

kaduk:
I don't think it looks like the desired rebase.
IIRC you have to use command-line git to do this rebase, and the workflow is something like:

```
git checkout CertificateVerify
git fetch origin
git rebase origin/master
git push github CertificateVerify
```
which assumes that origin is github.com/tlswg/tls13-spec and the 'github' remote is your personal github fork; modify accordingly.
I am not sure if there will be any rebase conflicts that need to be resolved (there may not be, since
you just have merge commits for changes you want to rebase past), but if not, there should be some documentation online for generic git conflict resolution scenarios.

jensguballa:
Ok, so some homework for me to do. :)

jensguballa:
Hopefully the PR now looks better. Sorry for my missing git experience. :)

kaduk:
Yes, the commit structure is as desired, now.
(No comment on the actual contents, yet, though ;) )

kaduk:
Probably worth saying, but I agree with MT that there's room for more detail about what exactly is verified, how it relates to the immediately prior Certificate message, how the x509 certificates therein are bound to any SNI in the handshake, etc.

mimoo:
btw, @vtlynch pointed to me that [the TLS 1.2 spec](https://tools.ietf.org/html/rfc5246#section-7.4.5) was more ambiguous about this:

> the client **SHOULD** verify that the server provided a valid certificate, if required

I find that weird, but I'm happy that TLS 1.3 would change that to a MUST.

jensguballa:
Some things need to be sorted out first. IMO there are several validations regarding the certficate:
  - validation of the identity provided in the certificate. On the client side his can be a consistency check against the SNI, if provided by the application, but it can be something else as well (e.g. externally received fingerprint of the certificate). I believe it depends too much on the use case to turn this validation into a MUST.
  - validation of the certificate's structure. This is out of scope of TLS, see section 4.4.1.4.
  - verification that the sender possesses the private key. Here I agree this should be a MUST, and that is addressed by this PR.

kaduk:
> verification that the sender possesses the private key. Here I agree this should be a MUST, and that is addressed by this PR.

The text in the PR could be a little more clear that that is what is being validated.

jensguballa:
I just updated the PR and provided more details for the verification process (which required some restructuring of the CertificateVerify section).

ekr:
@jensguballa: please provide a PR to add yourself to the contributors list.

jensguballa:
A separate PR has been created (#854, https://github.com/tlswg/tls13-spec/pull/854) to extend the distribution list.

PR #840, opened 2016-12-29
Require that (EC)DHE public values be fresh.

agl:
This change requires that (EC)DHE public values be fresh, for both parties, in every handshake.

For clients, this is standard practice (as far as I'm aware) so should make no difference. For servers, this is not always the case:

Springall, Durumeric & Halderman [note](https://dl.acm.org/citation.cfm?id=2987480) that with TLS 1.2:

- 4.4% of the Alexa Top 1M reuse DHE values and 1.3% do so for more than a day.
- 14.4% of the Top 1M reuse ECDHE values, 3.4% for more than a day.

Since this defeats forward security, and is clearly something that implementations of previous versions have done, this change specifically calls it out as a MUST NOT. Implementations would then be free to detect and reject violations of this.

This does have a cost because it also excludes the reasonable practice of amortising public value generation over all connections for a few seconds. The draft could attempt to specify a precise, maximum duration for reuse but that is more complex and no value is clearly optimal.

Also, this cost doesn't seem too high: 85.6% of servers _don't_ reuse values and manage fine today. The generation of (EC)DH public values is also a fixed-based operation and thus can be much faster than DH key-agreement.

Lastly, some have [proposed](https://datatracker.ietf.org/doc/draft-green-tls-static-dh-in-tls13/) (EC)DH reuse as a mechanism for enabling TLS connections to be decrypted and monitored by a middlebox. TLS is not designed to be decrypted by third-parties—that's kind of the point. Thus anyone doing this should not be surprised to hit a few MUST NOTs and, potentially, to have to configure implementations to allow such a deployment.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2244,9 +2244,10 @@ FFDHE groups.  The key_exchange values for each KeyShareEntry MUST be
 generated independently.  Clients MUST NOT offer multiple
 KeyShareEntry values for the same group.  Clients MUST NOT offer any
 KeyShareEntry values for groups not listed in the client's
-"supported_groups" extension.  Servers MAY check for violations of
-these rules and abort the handshake with an
-"illegal_parameter" alert if one is violated.
+"supported_groups" extension. Clients MUST NOT send the same KeyShareEntry
+in two ClientHello messages, even to different servers.  Servers MAY check for
+violations of these rules and abort the handshake with an "illegal_parameter"
+alert if one is violated.
 
 Upon receipt of this extension in a HelloRetryRequest, the client MUST
 verify that (1) the selected_group field corresponds to a group which was provided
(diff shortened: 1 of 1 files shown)


ekr:
@agl: can you please propose this on the list. It's clearly a substantive change and we have a conflicting PR that is based on the assumption that reuse is permissible

ekr:
Ah, I see you have. Thanks

agl:
@ekr: mail has already been sent. I think the mailing list machine is grinding away.

ekr:
@agl: so I see. Maybe it had to generate a fresh DH share :)

agl:
@brainhub yes, there are limits to what level of dumb we can detect. This is more about helping those who aren't thinking too much about it. We know that sites have gotten it wrong with TLS 1.2 because, presumably, they didn't mean to configure a forward-secure cipher suite and then invalidate it by keeping the same public-value for weeks.

But an actively hostile server can always just encrypt their entropy seed in the server nonce or so and we would never know.

briansniffen:
Of course they did. Configuring it as you describe gets the marketing benefits of ECDH"E", and avoids some problems with RSA Kx, at low cost. I do not endorse that decision, but I understand why some might make it.

It seems especially apt for sites that only serve one object — say, a JavaScript framework CDN. They can't have perfect forward secrecy, because they have no secrecy.  It is also helpful as an option when under certain sort of attack — to continue handshakes while capping the number of cryptographic operations performed.

In any case, it seems unwise to require that entropy be used in this way without an open discussion of the purposes of the extended random extension.

-- 
Brian Sniffen

ekr:
AGL seems to have disowned this proposal in:
https://www.ietf.org/mail-archive/web/tls/current/msg22285.html

Accordingly, in the interest of keeping the tracker clean I am closing this PR. We can re-open it if consensus changes.

PR #846, opened 2016-12-29
Use fewer newlines in verify_data expression.

davidben:
The other line-breaked expressions in the document do not use quite so
many newlines. I believe this matches the existing style, based on how
Derive-Secret is defined.


ekr:
This wasn't accidental, but I'm willing to change it.

davidben:
Oh if there's some particular reason for this styling, I don't have particular cares. It just seemed inconsistent with everything else and I was bored.

PR #850, opened 2017-01-16
Fix things, add things, use 0-RTT in paragraphs for consistency

ttaubert:
Reading through the spec again I found a few tiny things to fix. I'm suggesting to replace "Zero-RTT" with "0-RTT" in the paragraph because I found "Zero-RTT" used only in headlines, but "0-RTT" consistently in paragraphs.


ekr:
@ttaubert: Please provide a PR to add yourself to acknowledgements.

PR #857, opened 2017-01-22
removing the "subgroup attack" mention for FFDH

mimoo:
From the [Diffie-Hellman parameters](https://tlswg.github.io/tls13-spec/#ffdhe-param) section:

> Peers SHOULD validate each other’s public key Y by ensuring that 1 < Y < p-1. This check ensures that the remote peer is properly behaved and isn’t **forcing the local system into a small subgroup**.

Here at most you're avoiding the -1 public key that generates a subgroup of order 2, but that would never be enough to recover the full private key. 

So I think the "and isn’t forcing the local system into a small subgroup" is not necessary here. It just adds more words.

(A better check here would be to raise Y to the order of the used subgroup. [Looking at the list of groups](https://tools.ietf.org/html/rfc7919#appendix-A.1) I see that they are all safe primes so this check is useless (there is just no subgroup attack possible))

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2300,8 +2300,7 @@ Note: For a given Diffie-Hellman group, the padding results in all public keys
 having the same length.
 
 Peers SHOULD validate each other's public key Y by ensuring that 1 < Y
-< p-1. This check ensures that the remote peer is properly behaved and
-isn't forcing the local system into a small subgroup.
+< p-1. This check ensures that the remote peer is properly behaved.
 
 
 #### ECDHE Parameters {#ecdhe-param}


ekr:
I would prefer to leave this as-is.

1. This text appears in RFC 7919, which is defining the groups used here.
2. We do not want to be forced into the subgroup of order 2, which, as you say, this check prevents.

mimoo:
(There is nothing bad about being forced in the subgroup of order 2, I think here you mostly want to prevent computing things for nothing. It is not a primordial check imo.)

tomato42:
> Here at most you're avoiding the -1 public key that generates a subgroup of order 2, but that would never be enough to recover the full private key.

but it would be enough for eavesdropper to decrypt the connection

the check is trivial to perform and very cheap computationally, there's no reason to skip it

mimoo:
> but it would be enough for eavesdropper to decrypt the connection

Yeah, but same thing if a client sends the values g^1, g^2, g^3, g^4, etc... as public keys, do we test for these also?

davidben:
The check ensures that either side cannot force the secret to some particular value, which was relevant for [3-SHAKE](https://mitls.org/pages/attacks/3SHAKE). As I understand, the protocol is intended not to rely on this property, but we check it anyway to be safe. The spec similarly mandates the all-zeros check in RFC 7748 and rejecting the point at infinity for NIST curves.

(I agree that allowing an eavesdropper to decrypt the connection in the face a colluding endpoint is not a meaningful attack. 3-SHAKE was a bit more complicated than that.)

Issue #861, opened 2017-01-24
A broken syntax of  EarlyDataIndication


kazu-yamamoto:
The current syntax of `EarlyDataIndication` is broken because case arms must be terminated by `;`.

```
    struct {
       select (Handshake.msg_type) {
           case new_session_ticket:
              uint32 max_early_data_size;

           case client_hello:
           case encrypted_extensions:
              // empty
       };
   } EarlyDataIndication;
```

If we respect to TLS 1.2's syntax, I would suggest to change this into:

```
    struct {
       select (Handshake.msg_type) {
           case new_session_ticket:
              uint32 max_early_data_size;

           case client_hello:
           case encrypted_extensions:
              struct {};
       };
   } EarlyDataIndication;
```

mimoo:
1. I think this is more confusing than the `// empty`. The presentation language is not a formally specified syntax so I don't think it matters.

2. I remember tripping while reading this part of the specs because of the two cases following each other, I would rather replace them by this:

```
case client_hello:
    // empty
case encrypted_extensions:
    // empty
```

richsalz:
I don't like the comment, I'd rather have a syntactic token there and I don't care what it is.

mimoo:
What about just removing the cases for `client_hello` and `encrypted_extensions`?

richsalz:
Don't care. :)  Just don't want comments to be semantically meaningful.

kaduk:
Removing the cases is wrong, as then it leaves unspecified a structure that needs to be sent in a wire protocol message.

kazu-yamamoto:
`struct {}` is actually used to express empty in RFC 5246.
Let's just define its semantics in Section 3.

But I recommend most is to define `Empty`:

```
    struct {} Empty;

    struct {
       select (Handshake.msg_type) {
           case new_session_ticket:
              uint32 max_early_data_size;

           case client_hello:
           case encrypted_extensions:
              Empty;
       };
   } EarlyDataIndication;

```

mimoo:
Any other opinions on this? Possibilities I see:

* an `empty` field could be defined in the presentation language → but it's more text to the presentation language for few usages
* the text in the presentation language about cases that must cover all the cases could be relaxed → but now that I think more about it, it's nicer for the implementer to see all the cases in this snippet
* this `struct {} Empty;` could be defined  → I don't like it so much, I think a more casual `// empty` is actually clearer.

This makes me thing we should probably drop this issue.

[...]

PR #863, opened 2017-01-25
removed empty cases for EarlyDataIndication presentation language

mimoo:
removing confusing cases syntax after https://github.com/tlswg/tls13-spec/issues/861 discussion

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2395,10 +2395,6 @@ The "extension_data" field of this extension contains an
            select (Handshake.msg_type) {
                case new_session_ticket:
                   uint32 max_early_data_size;
-
-               case client_hello:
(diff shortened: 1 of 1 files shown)


kaduk:
I don't think this is correct -- we need to specify the structure that is sent in the body of this extension when carried in those messages; it just happens to be an empty structure.

mimoo:
I'm not sure how you would read that differently. An empty structure would be the same as no case

kaduk:
Well, we do say "The selector must be an enumerated type that defines the possible variants the structure defines. There must be a case arm for every element of the enumeration declared in the select."  So my claim would be more that it is invalid to have a select(Handshake.msg_type) that does not include a case arm for a message type that is allowed to contain this extension.

mimoo:
Ah I see! I missed that.

PR #866, opened 2017-01-27
Add reference to draft-ietf-tls-tls13-vectors

xiaoyinl:
Since [draft-ietf-tls-tls13-vectors](https://datatracker.ietf.org/doc/draft-ietf-tls-tls13-vectors/) has been adopted, I think we need to mention it in the TLS 1.3 spec, don't we? It seems to me Appendix C is a right section to mention it.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4536,7 +4536,8 @@ New cipher suite values are assigned by IANA as described in
 # Implementation Notes
 
 The TLS protocol cannot prevent many common security mistakes. This section
-provides several recommendations to assist implementors.
+provides several recommendations to assist implementors. Also
+{{?I-D.ietf-tls-tls13-vectors}} provides test vectors for TLS 1.3 handshakes.
 
 ## API considerations for 0-RTT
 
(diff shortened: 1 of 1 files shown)


seanturner:
If you're going to do this, then you're also going to need to add the reference in s11.2.

xiaoyinl:
Done. Thanks!

draft-ietf-tls-tls13.md
```diff
@@ -4536,7 +4537,8 @@ New cipher suite values are assigned by IANA as described in
 # Implementation Notes
 
 The TLS protocol cannot prevent many common security mistakes. This section
-provides several recommendations to assist implementors.
+provides several recommendations to assist implementors. Also
```
davegarrett: minor nit: should be a comma after "Also", here

Alternatively, just drop the "Also" entirely; it's unneeded.

seanturner:
It should be informative so I think it's { { ? followed by the RFC#.

draft-ietf-tls-tls13.md
xiaoyinl: I removed "Also". Thanks!

xiaoyinl:
@seanturner I don't think it's necessary to add `?` after `{{`, because [their syntax](https://github.com/cabo/kramdown-rfc2629#references) says if a reference is added to `Informative` section in the header, we don't need `?` or `!`.

seanturner:
Ooops I think you might be right.   It's in the informative section in the front part of the draft.

Issue #847, opened 2016-12-31
Session resume terminology


kroeckx:
There seem to be 3 cases left where we the draft talks about resumption:
- In 4.5.1: Any ticket MUST only be resumed with a cipher suite that has the same KDF hash [...]
- In 6.1: Failure to properly close a connection does not prohibit a session from being resumed.
- In 1.2: Restate rule from RFC 6066 that you can't resume unless SNI is the  same. (Talking about 4.5.1)

I think the term resume is misleading, you aren't resuming anything. You're really reusing parameters from that session, and can have multiple connections to the same server that use the same parameters.

jensguballa:
The term "resumption" occurs 38 times in the draft. :-)

Anyway, I am more interested in a good definition of the term "session". 

Some thoughts:
  - the term "session" is only required for resumption. That's my observation: if resumption would not be a feature of TLS, we would not need the term "session" at all.
  - a handshake produces a context, which comprises:
    - resumption_master_secret
    - the used KDF hash function
    - the fact that the server is authenticated
    - an indication if the client is authenticated or not
    - SNI (if provided by the client)
    - others (e.g. protocol version, cipher method, etc)
  - this context is used when generating a NewSessionTicket, or we can say: a ticket refers to this context
  - We can give this context a name: "session"
  - Thus each handshake produces a new unique session (because the resumption_master_secret will be unique)
  - A handshake may bootstrap from a previous session (via a ticket, "session resumption"), but even in this case a new session will be created
  - There is a one-to-one relationship between a connection and a session (the one created)
  - multiple connections may bootstrap from the same session

Do these thoughts make sense? If so, then probably some more considerations are needed, e.g. regarding externally provided PSKs ("Are these external sessions? Or external tickets?").

kaduk:
Many of those thoughts make sense, yes.

In some sense we may think about TLS connections more than TLS sessions, with a connection being defined as starting at initial ClientHello and ending at TCP close, timeout, or termination by alert.  A TLS session is then a way to extend some context from one connection to another (which causes trouble for the generic "one-to-one releationship between connection and session" you list above).

In some sense this is all editorial, though, so we may be subject to @ekr 's discretion.

katrielalex:
(moving discussion from #867) @jensguballa points out that the term "session key" appears throughout the appendix, starting with 
>  A "session key" (the master secret) from which can be derived a set of working keys.

[...]

PR #867, opened 2017-01-30
Clarify post-compromise (in)security and KCI resistance.

katrielalex:


katrielalex:
(requested in https://github.com/tlswg/tls13-spec/pull/837)

draft-ietf-tls-tls13.md
```diff
@@ -4949,6 +4980,12 @@ server is not able to guarantee full uniqueness of the handshake
 state. See {{early-data-indication}} for one mechanism to limit
 the exposure to replay.
 
+KCI resistance describes the security guarantees that a party may still have
+after its long-term secret has been compromised. TLS does not provide security
+for data communicated after the peer's long-term secret (signature key or
+external PSK) is compromised. It therefore does not provide post-compromise
+security {{CCG16}}, sometimes also referred to as backwards or future security.
```
ekr: 1. You start out by describing KCI but then don't say what TLS's behavior with respect to KCI is, but rather talk about security with respect to the *peer*'s long term secret rather than the local one.

2. This text doesn't seem quite correct. Consider the following sequence of events:

- TLS handshake between A and B using ECDHE
- A's long-term secret is compromised
- A sends message "X" to B

The compromise of A's long-term secret does not (as far as I know) threaten either the confidentiality or integrity of "X".

I think you need to split this up into compromises which happen before the handshake and those which happen after the handshake but before data is sent on the channel

draft-ietf-tls-tls13.md
```diff
@@ -5010,11 +5047,12 @@ pseudorandom function (PRF).  In addition, as long as this function is truly
...
 
-TLS does not provide security for data which is sent after a traffic secret is
-compromised (backward secrecy). Compromise of a traffic secret also compromises
-future confidentiality and integrity.  Systems which want backward secrecy need
-to do a fresh handshake and establish a new session key with an (EC)DHE
-exchange.
+TLS does not provide security for data which is communicated after a traffic
+secret is compromised i.e., post-compromise security/future secrecy/backward
+secrecy with respect to the traffic secret. Indeed, an attacker who learns a
+traffic secret can compute all future traffic secrets.  Systems which want such
```
ekr: "all future traffic secrets on that connection"

draft-ietf-tls-tls13.md
katrielalex: Re 1: my emphasis was unclear---I meant to contrast KCI resistance (if _your_ key has been compromised) with PCS (if your _peer's_ key has been compromised). Let me reword it a bit.

katrielalex: Re 2: I think the division is between compromises which take place before the handshake and those which take place after (regardless of whether data have been sent). But yes, I agree that your scenario shows that the text isn't quite right.

draft-ietf-tls-tls13.md
```diff
@@ -4866,13 +4890,20 @@ Downgrade protection.
...
 : If the long-term keying material (in this case the signature keys in certificate-based
 authentication modes or the external/resumption PSK in PSK with (EC)DHE modes) are compromised after
 the handshake is complete, this does not compromise the security of the
 session key (See {{DOW92}}).  The forward secrecy property is not satisfied
 when PSK is used in the "psk_ke" PskKeyExchangeMode.
 
+Key Compromise Impersonation (KCI) resistance
+: In a mutually-authenticated connection with certificates, peer authentication
+should hold even if the local long-term secret was compromised before the
+connection was established (See {{HGFS15}}). For example, if a client's
```
martinthomson: nit s/See/see

draft-ietf-tls-tls13.md
```diff
@@ -4949,6 +4980,13 @@ server is not able to guarantee full uniqueness of the handshake
 state. See {{early-data-indication}} for one mechanism to limit
 the exposure to replay.
 
+TLS does not provide security for handshakes which take place after the peer's
+long-term secret (signature key or external PSK) is compromised. It therefore
+does not provide post-compromise security {{CCG16}}, sometimes also referred to
+as backwards or future security. This is in contrast to KCI resistance, which
+describes the security guarantees that a party may still have after its *own*
```
martinthomson: The emphasis on "own" here will probably render strangely.  You should remove it.

Also, "may" implies permissiveness.  "a party has after" would work equally well and is more assertive (I hope that being definitive rather than equivocal here is OK, I tend to think that it's OK as long as P=~0).

draft-ietf-tls-tls13.md
```diff
@@ -5010,11 +5048,12 @@ pseudorandom function (PRF).  In addition, as long as this function is truly
 one way, it is not possible to compute traffic keys from prior to a key change
 (forward secrecy).
 
-TLS does not provide security for data which is sent after a traffic secret is
-compromised (backward secrecy). Compromise of a traffic secret also compromises
-future confidentiality and integrity.  Systems which want backward secrecy need
-to do a fresh handshake and establish a new session key with an (EC)DHE
-exchange.
+TLS does not provide security for data which is communicated on a connection
+after a traffic secret of that connection is compromised i.e., post-compromise
```
martinthomson: Nit: I would s/compromised i.e., post-/compromised. That is, TLS does not provide post-/

draft-ietf-tls-tls13.md
```diff
@@ -4866,13 +4890,20 @@ Downgrade protection.
 should be the same as if the peers had been communicating in the
 absence of an attack (See {{BBFKZG16}}; defns 8 and 9}).
 
-Forward secret
+Forward secret with respect to long-term keys
 : If the long-term keying material (in this case the signature keys in certificate-based
 authentication modes or the external/resumption PSK in PSK with (EC)DHE modes) are compromised after
 the handshake is complete, this does not compromise the security of the
 session key (See {{DOW92}}).  The forward secrecy property is not satisfied
```
jensguballa: Shouldn't "session key" be replaced by "master secret"? Neither "session" nor "session key" is well defined at all. Refer to issue #847.

katrielalex: I've changed this, but "session key" appears elsewhere in the document. Should it be changed everywhere?

jensguballa: I think replacing it consistently throughout the document is the right way. Not sure if more consensus is needed.

katrielalex: I'll let others comment but I agree that it should either be defined or removed (and slightly prefer the latter)

mimoo: I'm not sure about that, the session key is defined and used all over the place. Also you could think that forward secrecy also affects the handshake secret which includes the master secret but the reverse is not true.

kaduk:
It also seems okay to me.  I was reading quickly, and did not come up with any editorial nits.

katrielalex:
It looks like this is stalled on the naming of "session keys" through the document. @ekr, shall I open an issue for that to unblock this?

ekr:
Please

katrielalex:
OK, moving discussion to #847 since there's already a related open issue; this PR doesn't add more uses of the term "session key" so hopefully it can go in without waiting for that one.

PR #870, opened 2017-02-02
Using both External PSK and (EC)DH in TLS 1.3.

seanturner:
The following text is a proposal from Russ Housley based on the following thread:
https://mailarchive.ietf.org/arch/msg/tls/1W_LRzLQHHH3iGS3zs-7LsMtx-8

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1498,15 +1498,21 @@ If the server successfully selects parameters and does not require a
 HelloRetryRequest, it indicates the selected parameters in the ServerHello as
 follows:
 
-- If PSK is being used, then the server will send a
-"pre_shared_key" extension indicating the selected key.
-- If PSK is not being used, then (EC)DHE and certificate-based
-authentication are always used.
-- When (EC)DHE is in use, the server will also provide a
-"key_share" extension.
-- When authenticating via a certificate (i.e., when a PSK is not
-in use), the server will send the Certificate ({{certificate}}) and
-CertificateVerify ({{certificate-verify}}) messages.
+- If PSK is not being used, then (EC)DHE and certificate-based authentication are always used, and the server will:
+
+    --  provide a "key_share" extension; and
+
+    --  send the Certificate (Section 4.4.1) and CertificateVerify (Section 4.4.2) messages.
+
+- If PSK (without DH or ECDH) is being used, then the server sends a "pre_shared_key" extension to indicate the selected key.
+
+-  If PSK and (EC)DH are being used together, then the server will:
+
+    --  sends a "pre_shared_key" extension to indicate the selected key;
+
+    --  provide a "key_share" extension; and
+
+    --  send the Certificate (Section 4.4.1) and CertificateVerify (Section 4.4.2) messages.
 
 If the server is unable to negotiate a supported set of parameters
 (i.e., there is no overlap between the client and server parameters),
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1508,7 +1508,7 @@ follows:
 
 -  If PSK and (EC)DH are being used together, then the server will:
 
-    --  sends a "pre_shared_key" extension to indicate the selected key;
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1498,15 +1498,21 @@ If the server successfully selects parameters and does not require a
...
+
+    --  provide a "key_share" extension; and
+
+    --  send the Certificate (Section 4.4.1) and CertificateVerify (Section 4.4.2) messages.
+
+- If PSK (without DH or ECDH) is being used, then the server sends a "pre_shared_key" extension to indicate the selected key.
+
+-  If PSK and (EC)DH are being used together, then the server will:
+
+    --  sends a "pre_shared_key" extension to indicate the selected key;
```
martinthomson: s/sends/send

draft-ietf-tls-tls13.md
```diff
@@ -1498,15 +1498,21 @@ If the server successfully selects parameters and does not require a
...
+
+    --  provide a "key_share" extension; and
+
+    --  send the Certificate (Section 4.4.1) and CertificateVerify (Section 4.4.2) messages.
```
martinthomson: I think that this is precisely the opposite of what the document currently specifies.  That is, PSK currently means that a certificate must not appear.

draft-ietf-tls-tls13.md
```diff
@@ -1498,15 +1498,21 @@ If the server successfully selects parameters and does not require a
...
-CertificateVerify ({{certificate-verify}}) messages.
+- If PSK is not being used, then (EC)DHE and certificate-based authentication are always used, and the server will:
+
+    --  provide a "key_share" extension; and
+
+    --  send the Certificate (Section 4.4.1) and CertificateVerify (Section 4.4.2) messages.
+
+- If PSK (without DH or ECDH) is being used, then the server sends a "pre_shared_key" extension to indicate the selected key.
+
+-  If PSK and (EC)DH are being used together, then the server will:
```
kaduk: Github is doing different syntax highlighting for the toplevel bullet points, and maybe they have inconsistent whitespace for them.

draft-ietf-tls-tls13.md
kaduk: I don't think PSK means that a certificate must not appear; I think it is not a specified behavior in the current document and is left to a future document to interpret.
It's still not clear to me that we need to paint ourselves into a corner at this time -- I never saw an explanation of which part of the existing text was problematic with respect to this issue.

draft-ietf-tls-tls13.md
```diff
@@ -1498,15 +1498,21 @@ If the server successfully selects parameters and does not require a
...
+
+    --  send a "pre_shared_key" extension to indicate the selected key;
+
+    --  provide a "key_share" extension; and
+
+    --  send the Certificate (Section 4.4.1) and CertificateVerify (Section 4.4.2) messages.
```
martinthomson: In response to @kaduk, since github has eaten his response in a way that I can't respond...

I think that this is a huge problem waiting to happen.  Current implementations assume that using PSK means no certificate.  We can't have ambiguity on this point or we won't get interoperability.  If you are saying that if PSK implies that the choice of whether to send a certificate is up to the server, then we need to explicitly say that.

ekr: I agree with MT. The current draft is intended to prohibit the server sending a certificate with PSK mode. I'm happy to take PRs that consolidate that prohibition so that if/when we want to relax the requirement with some extension it's easy, but I think that the prohibition needs to remain in the draft.

russhousley: Draft -18 already says this is possible at the bottom of page 17.  This PR was intended to say how to do what the document says can be done.  Page 17 says:

   PSKs can be used with (EC)DHE exchange in order to provide forward
   secrecy in combination with shared keys, or can be used alone, at the
   cost of losing forward secrecy.

ekr: Yes, PSK can be used with ECDHE. It cannot be used with server authentication via a signature.

Issue #872, opened 2017-02-06
Exporter label requirements


jsalowey:
refer the reader to exporter requirements in RFC 5705

"Requirements for the format of Exporter labels are defined in section 4 of RFC5705."

Creating pull request.

ekr:
@jsalowey: should we expect a PR?

ekr:
This appears to be done.

jsalowey:
yes I submitted PR #874 which has been merged already.

PR #873, opened 2017-02-06
Added reference to exporter format in RFC 5705

jsalowey:
see Issue 872 #872


jsalowey:
Need to update base a resubmit

Issue #1, opened 2014-04-17
SNI encryption


martinthomson:
Server Name Indication could be encrypted to protect it from being inspected by passive attackers.  There are some virtual-hosting type situations where hiding this information is considered desirable.  If we can encrypt SNI, then it is probably trivial to encrypt other extensions.

However, this complicates the handshake.  A lot.  It also complicates virtual-hosting scenarios.  Rich summarizes the issues pretty well here: http://www.ietf.org/mail-archive/web/tls/current/msg11823.html

leonklingele:
Is this already ruled-out? How could the SNI be encrypted without an addition RTT?

One could hash the SNI in the Client Hello which would at least help if the served host name is not publicly known (e.g. using a self-signed certificate).

kaduk:
It seems likely that there will eventually be an (optional?) extension that permits an encrypted "real" SNI, with some fake public SNI.  But maybe that is not what you are asking.

leonklingele:
@ekr why was this closed?

ekr:
Because we are not going to do anything about it in the base specification.

PR #875, opened 2017-02-09
Add an extra Derive-Secret stage prior to HKDF-Extract. This has two

ekr:
benefits:

(1) Restoring extract/expand parity, thus cleaning up the theoretical
    treatment wrt HKDF and future replacements.
(2) Futureproofing against someone being able to force the ECDHE
    value in the second HKDF-Extract to collide with one of the labels.
    This seems unlikely with current groups, but might be possible
    with some future algorithm. (Issue due to Trevor Perrin).

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4056,14 +4056,11 @@ In this diagram, the following formatting conventions apply:
                  +-----> Derive-Secret(., "early exporter master secret",
                  |                     ClientHello)
                  |                     = early_exporter_secret
-                 |
-           Derive-Secret(., "derived early secret", "")
+                 v
+           Derive-Secret(., "derived secret", "")
                  |
                  v
-(EC)DHE -> HKDF-Extract
-                 |
-                 v
-         Handshake Secret
+(EC)DHE -> HKDF-Extract = Handshake Secret
                  |
                  +-----> Derive-Secret(., "client handshake traffic secret",
                  |                     ClientHello...ServerHello)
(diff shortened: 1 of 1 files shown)


davidben:
This is mostly a cosmetic thing, and I'm not sure if I actually like all of my proposals below:

With the current scheme, the key schedule is very simple and neatly encompassed into a box with a shape like this:

- NewTLS13KeySchedule(hash) => creates a new schedule with base secret = 0
- AddEntropy(ikm) => HKDF-Extract ikm into the current base secret
- DeriveSecret(label) => take the current base secret and handshake transcript and derive a particular secret with the label

This breaks the symmetry in a few ways:

1. Incorporating a new IKM requires an extra label for the Derive-Secret operation in the middle.
2. Going from 0 to Early Secret does not involve a Derive-Secret.
3. We use Derive-Secret, Derive-Secret no longer always uses the current state of the handshake transcript

(1) could be resolved by just using the same label for both, like "prepare extract" or something. I don't believe this breaks anything? We're just trying to fix the problem where HKDF-Extract and HKDF-Expand can't be used on the same key.

(3) is actually kind of odd because Derive-Secret is defined as:

    Derive-Secret(Secret, Label, Messages) =
         HKDF-Expand-Label(Secret, Label,
                           Hash(Messages), Hash.Length)

So we're actually putting Hash("") into HKDF-Expand-Label, not the empty string. Then again, it feels odd making them ClientHello...ServerHello and ClientHello...Server Finished, respectively. We could use HKDF-Expand-Label directly which still gives us the separation, but it's not as obvious.

Adding an extra Derive-Secret just to fix (2) feels a little pointless but would increase symmetry.

(Alternatively, the answer could just be "deal with it, your key schedule box should have slightly richer methods on it". Like I said, this is just cosmetic.)

martinthomson:
I'm ambivalent on David's gripe.  This seems fine to me.

One editorial suggestion though.  This is now a very tall diagram.  You can save a few lines by changing to something like this

```
         |
         v
    Derive-Secret(., "derived handshake secret")
         |
         v
0 -> HKDF-Extract = handshake secret
         |
         +-----> Derive-Secret(., ...
```

mimoo:
There is a con: complicating the already complicated key schedule

kaduk:
> About this PR, there is a con: complicating the already complicated key schedule

It does not seem especially complicated compared to, e.g., Kerberos.  Or maybe that's just me.

mimoo:
well, you're comparing it to Kerberos :rofl:

ekr:
@davidben: I fixed #1. I think we can live without #2. We already have this "empty hash" construct for the binder keys, so while it's a bit goofy it's no more goofy than here.

PR #876, opened 2017-02-09
Convert the handshake hash into a hash of hashes. This enables

ekr:
HRR to just contain a hash value rather than some odd partial
hash state.


mimoo:
I'm late to the party, what's the point of that change?

kaduk:
See https://www.ietf.org/mail-archive/web/tls/current/msg22374.html

ekr:
I have been convinced that this particular construction is less good, so closing in favor of PR #901

PR #877, opened 2017-02-20
No fallthrough

### LINKED ISSUE #861: A broken syntax of  EarlyDataIndication

kazu-yamamoto:
To fix #861, let's add semi-colons to EarlyDataIndication. To make the presentation language simpler, let's remove fall-through which is not used in other places.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1334,9 +1334,8 @@ Fields and variables may be assigned a fixed value using "=", as in:
 
 Defined structures may have variants based on some knowledge that is
 available within the environment. The selector must be an enumerated
-type that defines the possible variants the structure defines. There
-must be a case arm for every element of the enumeration declared in
-the select. The body of the variant structure may be given a label
+type that defines the possible variants the structure defines.
+The body of the variant structure may be given a label
 for reference. The mechanism by which the variant is selected at
 runtime is not prescribed by the presentation language.
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2396,10 +2396,8 @@ The "extension_data" field of this extension contains an
            select (Handshake.msg_type) {
                case new_session_ticket:
                   uint32 max_early_data_size;
-
-               case client_hello:
-               case encrypted_extensions:
-                  // empty
+               case client_hello: ; // empty
+               case encrypted_extensions: ; // empty
            };
        } EarlyDataIndication;
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1371,7 +1371,7 @@ For example:
            select (VariantRecord.type) {
                case apple:  V1;
                case orange: V2;
-           } variant_body;       /* optional label on variant */
+           };
        } VariantRecord;
 
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1368,7 +1368,8 @@ For example:
        } V2;
 
        struct {
-           select (VariantTag) { /* value of selector is implicit */
+           VariantTag type;
+           select (VariantRecord.type) {
                case apple:  V1;  /* VariantBody, tag = apple */
                case orange: V2;  /* VariantBody, tag = orange */
            } variant_body;       /* optional label on variant */
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1370,8 +1370,8 @@ For example:
        struct {
            VariantTag type;
            select (VariantRecord.type) {
-               case apple:  V1;  /* VariantBody, tag = apple */
-               case orange: V2;  /* VariantBody, tag = orange */
+               case apple:  V1;
+               case orange: V2;
            } variant_body;       /* optional label on variant */
        } VariantRecord;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1375,12 +1368,9 @@ For example:
        } V2;
 
        struct {
-           select (VariantTag) {
-               case apple:
-                 V1;   /* VariantBody, tag = apple */
-               case orange:
-               case banana:
-                 V2;   /* VariantBody, tag = orange or banana */
+           select (VariantTag) { /* value of selector is implicit */
```
martinthomson: Not sure that I understand this comment.  I'd remove it.  I realize that the expression "VariantTag" isn't understood, but it isn't critical to understanding.  Or, you could change this to be more explicit:

```
struct {
    VariantTag type;
    select (VariantRecord.type) {
```

kazu-yamamoto: Good point!

draft-ietf-tls-tls13.md
```diff
@@ -1375,12 +1368,10 @@ For example:
...
        struct {
-           select (VariantTag) {
-               case apple:
-                 V1;   /* VariantBody, tag = apple */
-               case orange:
-               case banana:
-                 V2;   /* VariantBody, tag = orange or banana */
+           VariantTag type;
+           select (VariantRecord.type) {
+               case apple:  V1;  /* VariantBody, tag = apple */
```
mimoo: are these "tag" things used anywhere anyway? Can't we remove that as well?

mimoo:
LGTM

draft-ietf-tls-tls13.md
kazu-yamamoto: Good point.

We can replace it to `type == apple` or remove it. Which do you prefer?

kazu-yamamoto:
One more thing:

> There must be a case arm for every element of the enumeration declared in the select.

This is not true anymore. For instance `KeyShare` does not cover all handshake types. How can we improve this explanation?

ekr:
Hi folks. Sorry it took me so long to weigh in, but I'm not really excited about this set of changes.
As far as I can tell, the primary impetus for this is the need to specify an empty value for a case. 

The simplest thing to do here is just to define struct {} Empty; and then put Empty; in the relevant case arms. It's not new syntax, it's just a new type definition.

As for fallthrough, IMO it makes things easier to read.

I'm happy to remove the requirement to cover all types, because that also makes things easier to read.

kaduk:
I think I had managed to squint enough that I could convince myself that only all values of the enum being used as a selector *that are valid in the current context* need to be enumerated.  So, when talking about, e.g., KeyShare, it's okay to only list case arms for client_hello, hello_retry_request, and server_hello, since other handshake messages would not contain a KeyShare.  But, maybe that requires too much squinting.

draft-ietf-tls-tls13.md
mimoo: I'd remove it entirely as I don't see any of these VariantBody and tag things used anywhere in the spec.

mimoo:
yeah @kaduk is right, KeyShare does list all the possible options relevant to KeyShare. But yeah the quote is hard to read.

@ekr I think it's a matter of what you're most used to read, but I find the PR easier to understand while defining a new struct {} Empty or having the fall through would be cumbersome in my opinion.

kazu-yamamoto:
@ekr I pushed three more commits. I hope that you like these.

draft-ietf-tls-tls13.md
```diff
@@ -1370,8 +1369,8 @@ For example:
        struct {
            VariantTag type;
            select (VariantRecord.type) {
-               case apple:  V1;  /* VariantBody, tag = apple */
-               case orange: V2;  /* VariantBody, tag = orange */
+               case apple:  V1;
+               case orange: V2;
            } variant_body;       /* optional label on variant */
```
mimoo: what about this comment

draft-ietf-tls-tls13.md
```diff
@@ -2382,12 +2382,13 @@ The "extension_data" field of this extension contains an
 
 %%% Key Exchange Messages
 
+       struct {} Empty;
```
mimoo: Again, the specification does say that the presentation language is casually defined, I voice my concerns on this :)

(also I think there is another // empty somewhere in the spec)

draft-ietf-tls-tls13.md
kazu-yamamoto: If we remove the comment, I would like to remove `variant_body`, too.

kazu-yamamoto: The tag of `select` is only used in `Handshake` where the `body` tag is not referred at all.
We could remove the `body` tag of `Handshake`.

draft-ietf-tls-tls13.md
kazu-yamamoto: I doubly checked with my validator.

- `Empty`s are used only here.
- The syntax of this branch is valid.

kazu-yamamoto:
I have removed the tag and comment.

kazu-yamamoto:
@ekr I think this PR is ready to be merged.

PR #878, opened 2017-02-21
Security properties: note on awareness of authentication status

cascremers:
Added note in security properties section on the fact that the client cannot determine if the server considers the connection to be mutually authenticated.
    
This paragraph condenses the main takeaway from the discussion at
https://www.ietf.org/mail-archive/web/tls/current/msg22382.html
    
Cas Cremers, Marko Horvat, J. Hoyland, T. van der Merwe, and Sam Scott.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -224,6 +224,21 @@ informative:
          ins: T. van der Merwe
        -
          ins: S. Scott
+  CHHSV17:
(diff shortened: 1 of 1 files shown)


mimoo:
I think a mention of this problem could have been useful in the [receiving a certificate message](https://tlswg.github.io/tls13-spec/#receiving-a-certificate-message) section as well.

PR #879, opened 2017-02-22
Grammar fixes

Buddybenj:


draft-ietf-tls-tls13.md
```diff
@@ -4771,7 +4771,7 @@ with a "protocol_version" alert.
 
 Note that earlier versions of TLS did not clearly specify the record layer
 version number value in all cases (TLSPlaintext.legacy_record_version). Servers
-will receive various TLS 1.x versions in this field, however its value
+will receive various TLS 1.x versions in this field, but its value
```
ekr: however, seems fine to me here. the rest LGTM

Buddybenj: In the way it's being used here, "however" typically starts a new sentence (or continues a sentence by using a semicolon). I would also argue that "but" reads nicer in this sentence. However, I will change it if you want me to.

The reason this form of "however" typically starts a new sentence is because without a new sentence, it's usually used in this way: "The TLS standard, however, does not specify how protocols add security with TLS".

PR #886, opened 2017-03-03
Various updates to stuff which I didn't like / was incorrect / missing something

leonklingele:


draft-ietf-tls-tls13.md
```diff
@@ -3101,13 +3104,13 @@ The digital signature is then computed over the concatenation of:
 
 - A string that consists of octet 32 (0x20) repeated 64 times
 - The context string
-- A single 0 byte which serves as the separator
+- A single 0-byte which serves as the separator
```
kaduk: No, this is a byte which has value zero, not a thing that is zero bytes long.

leonklingele: You're right! Reverted with a1744b2686e21f1d525e852305ab77b8f78908fa

leonklingele:
Added more updates. This MR is no longer WIP and can be reviewed :)

ekr:
@leonklingele: this seems to consist of a number of systematic text changes, some of which I like and some of which I don't. Looking through these:

- adding periods where needed [OK]
- (See Section XXX) -> (see section XXX) [OK]
- capitalizing all message names [this seems like a case by case thing. I don't think "send an alert" needs to be capitalized]
- plural agreement for verbs [this is fine in principle, but not all of your changes appear correct.]

It would be helpful if you would break these up into separate PRs or at least separate commits, so I can handle them independently.

I see you also changed a bunch of stuff that's descriptive to normative. These are non-editorial changes and should be part of a separate PR.

leonklingele:
Updated this PR to only include cosmetic fixes.

leonklingele:
Non-cosmetic stuff is in #892

draft-ietf-tls-tls13.md
```diff
@@ -2471,7 +2473,7 @@ connection during which the ticket was established.
 
 Future extensions MUST define their interaction with 0-RTT.
 
-If any of these checks fail, the server MUST NOT respond
```
davegarrett: this is correct as-is

draft-ietf-tls-tls13.md
```diff
@@ -3517,10 +3520,10 @@ described in the following section.
 ## Record Payload Protection
 
 The record protection functions translate a TLSPlaintext structure into a
-TLSCiphertext. The deprotection functions reverse the process. In TLS 1.3
+TLSCiphertext. The deprotection functions reverse the process. In TLS 1.3,
 as opposed to previous versions of TLS, all ciphers are modeled as
 "Authenticated Encryption with Additional Data" (AEAD) {{RFC5116}}.
-AEAD functions provide a unified encryption and authentication
```
davegarrett: this is correct as-is
https://www.usingenglish.com/forum/threads/150770-a-unified

draft-ietf-tls-tls13.md
```diff
@@ -3687,7 +3690,7 @@ content type of the message.
 This padding scheme was selected because it allows padding of any encrypted
 TLS record by an arbitrary size (from zero up to TLS record size
 limits) without introducing new content types.  The design also
-enforces all-zero padding octets, which allows for quick detection of
```
davegarrett: I think the current text is preferable here. Both might be grammatically correct, but the former sounds better.

(design allows for; octets allow for)

draft-ietf-tls-tls13.md
```diff
@@ -3736,11 +3739,11 @@ Some alerts are sent to
 indicate orderly closure of the connection or the end of
 early data (see {{closure-alerts}}).
 Upon receiving such an alert, the TLS implementation SHOULD
-indicate end-of-data to the application and, if appropriate for
```
davegarrett: ditch the comma after "and"

draft-ietf-tls-tls13.md
```diff
@@ -4238,7 +4241,7 @@ ECDH functions are used as follows:
   input).
 * The ECDH shared secret is the result of applying the ECDH function to the
   secret key (into scalar input) and the peer's public key (into u-coordinate
-  point input). The output is used raw, with no processing.
```
davegarrett: It is correct as-is. I'm not convinced "rawly" is a word.

draft-ietf-tls-tls13.md
```diff
@@ -4302,7 +4305,7 @@ applicable features:
 
   * "supported_versions" is REQUIRED for all ClientHello messages.
   * "signature_algorithms" is REQUIRED for certificate authentication.
-  * "supported_groups" and "key_share" are REQUIRED for DHE or ECDHE key exchange.
```
davegarrett: This is correct as-is.
A && B are required for D || C
if D || C, must A && B
NOT just if D && C, must A && B

draft-ietf-tls-tls13.md
```diff
@@ -847,7 +849,7 @@ secret keying material. Once the handshake is complete, the peers
 use the established keys to protect application layer traffic.
 
 A failure of the handshake or other protocol error triggers the
-termination of the connection, optionally preceded by an alert message
+termination of the connection, optionally preceded by an Alert message
```
davegarrett: I don't think we need to go through here changing all these capitalization. Please fork all of the message type capitalization changes off into a separate PR for separate consideration to the typo and grammar changes. (and do a squash of all the little commits into one or a few)

ekr:
@leonklingele: I merged this by hand, dropping some of commits that I was unhappy about (or in some cases @davegarrett was unhappy about). I also edited a few of them to clean them up.

leonklingele:
👍 and sorry for nitpicking.

PR #887, opened 2017-03-03
DH params: Peers MUST validate each other's public key

leonklingele:
I wonder why this was only a `SHOULD` before.
Am I right to assume that this is because TLS 1.3 has removed support for static DH and leaking an ephemeral key is not considered to be a security problem?

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2315,7 +2315,7 @@ bytes.
 Note: For a given Diffie-Hellman group, the padding results in all public keys
 having the same length.
 
-Peers SHOULD validate each other's public key Y by ensuring that 1 < Y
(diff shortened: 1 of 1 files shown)


martinthomson:
Gotta say, I don't know why this was SHOULD.  That check is very easy to implement.

leonklingele:
[RFC 7919, section 5.1](https://tools.ietf.org/html/rfc7919#section-5.1) says `MUST`.

ekr:
This text was adapted from some old draft of 7919 which had should and never got updated (see for instance: https://tools.ietf.org/html/draft-ietf-tls-negotiated-ff-dhe-02). Good catch.

ekr:
@leonklingele Please provide a PR adding yourself to contributors.

leonklingele:
#891

PR #892, opened 2017-03-05
Keyword updates

leonklingele:


draft-ietf-tls-tls13.md
```diff
@@ -1825,10 +1825,10 @@ Extensions are generally structured in a request/response fashion, though
...
 its extension responses in the ServerHello, EncryptedExtensions
-and HelloRetryRequest messages. The server sends extension requests,
-in the CertificateRequest message which can be responded to with
-the client's Certificate message. The server can also send unsolicited
-extensions in the NewSessionTicket, though the client does not respond
+and HelloRetryRequest messages. The server sends extension requests
+in the CertificateRequest message which a client MAY respond to with
+a Certificate message. The server MAY also send unsolicited
+extensions in the NewSessionTicket, though the client MUST NOT respond
 directly to these.
```
ekr: This MUST NOT doesn't seem quite right. It's not normative.

draft-ietf-tls-tls13.md
```diff
@@ -2340,18 +2340,18 @@ curve.
 For the curves secp256r1, secp384r1 and secp521r1, the appropriate
 validation procedures are defined in Section 4.3.7 of {{X962}}
 and alternatively in Section 5.6.2.6  of {{KEYAGREEMENT}}.
-This process consists of three steps: (1) verify that Y is not the point at
+This process MUST consist of these three steps: (1) verify that Y is not the point at
```
ekr: I don't think this is necessary, because the MUST is somewhere else.

draft-ietf-tls-tls13.md
```diff
@@ -2340,18 +2340,18 @@ curve.
...
 infinity (O), (2) verify that for Y = (x, y) both integers are in the correct
 interval, (3) ensure that (x, y) is a correct solution to the elliptic curve equation.
-For these curves, implementers do not need to verify membership in the correct subgroup.
+For these curves, implementers MAY optionally verify membership in the correct subgroup.
```
ekr: This seems to sort of invert the sense. The "do not need to" is telling you that it's not ncessary.

draft-ietf-tls-tls13.md
```diff
@@ -3400,7 +3400,7 @@ the forward secrecy of data sent before the sender changed keys.
 
 If implementations independently send their own KeyUpdates with
 request_update set to "update_requested", and they cross in flight, then each side
-will also send a response, with the result that each side increments
+MUST also send a response, with the result that each side increments
```
ekr: I think we can remove the MUST here b/c the MUST requirement is above. This is telling you how things behave.

draft-ietf-tls-tls13.md
```diff
@@ -3818,20 +3818,20 @@ user_canceled
...
+received after a closure alert MUST be ignored. If a transport-level close is
 received prior to a "close_notify", the receiver cannot know that all the
 data that was sent has been received.
 
 Each party MUST send a "close_notify" alert before closing the write side
 of the connection, unless some other fatal alert has been transmitted. The
 other party MUST respond with a "close_notify" alert of its own and close down
 the connection immediately, discarding any pending writes. The initiator of the
-close need not wait for the responding "close_notify" alert before closing the
+close MUST NOT wait for the responding "close_notify" alert before closing the
```
ekr: This is not correct, because you *can* wait for it, you just need not.

ekr:
Merged the subset as noted above

Issue #836, opened 2016-12-28
PSK binder validation and fallback


kazu-yamamoto:
Section 4.2.8. says:

> Prior to accepting PSK key establishment, the server MUST validate the corresponding binder value (see Section 4.2.8.1 below). If this value is not present or does not validate, the server MUST abort the handshake. 

Are there any reasons to not allow fall back to full handshake if binder value is not present or does not validate?

kazu-yamamoto:
If "the handshake" means "selecting the PSK", not the entire handshake, I would suggest to change the wording.

ekr:
I'm not sure why you would want to allow it. There's no legitimate reason for a binder to be invalid.

kazu-yamamoto:
Suppose there is one pair of PSK and its binder. If the binder is invalid, I want to fallback to full handshake. But this sentence reads that the server must abort the entire handshake in that case.

davidben:
Why? Implementations aren't expected to fallback to full handshake if early data can't be decrypted, which is a similar sort of thing. If a PSK can't be resumed because some configuration or state changed and it's no longer valid, you want to fallback. If it can't be resumed because the client doesn't appear to have possession of the PSK, a hard error is reasonable.

Making this a fatal error means it's less likely that interop errors in computing the binder will get masked.

kazu-yamamoto:
OK. Make sense. It would be nice if such conditions for fallback are explicitly described in the draft.

ekr:
@kazu-yamamoto: PR welcome

kazu-yamamoto:
Let's close this issue in favor of #893.

PR #893, opened 2017-03-06
Fall back

kazu-yamamoto:
For #836, I would open this PR.

Please discuss whether or not we should use "SHOULD" here.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1813,7 +1813,17 @@ A number of TLS messages contain tag-length-value encoded extensions structures.
 
        struct {
            ExtensionType extension_type;
-           opaque extension_data<0..2^16-1>;
+           select (Extension.extension_type) {
+               case supported_groups:      NamedGroupList;
+               case signature_algorithms:  SignatureSchemeList;
+               case key_share:             KeyShare;
+               case pre_shared_key:        PreSharedKeyExtension;
+               case early_data:            EarlyDataIndication;
+               case supported_versions:    SupportedVersions;
+               case cookie:                Cookie;
+               case psk_key_exchange_modes: PskKeyExchangeModes;
+               case ticket_early_data_info: TicketEarlyDataInfo;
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2612,6 +2625,11 @@ facilitates implementation as described below). Servers MUST check
 that it is the last extension and otherwise fail the handshake with an
 "illegal_parameter" alert.
 
+When the syntax of this extension is correct but the server cannot
+verify at least one binder because of some configuration or state
+changes, the server SHOULD fall back to a full 1-RTT handshake if a
+"key_share" extension is supplied.
```
martinthomson: If the binder is present and the server believes that it has a key, then it MUST fail the handshake (in my view).  That's certainly what NSS does.  If the binder is present, but the server doesn't have a key, then it doesn't have a key and can't resume.

ekr: I agree with MT. The order of operations is:

- Determine whether you are going to PSK and select the PSK
- If you are resuming, check the binder.

If the binder checks fails, fail the handshake

kazu-yamamoto: Ah, I wrote this PR with confusion. The binder MUST be verified anyway.
What I should have talked about the encrypted PSK value itself.
I will push a fix.
Thank you for pointing out.

kazu-yamamoto:
I pushed a commit to fix the condition. What about this?

kazu-yamamoto:
I'm very sorry.
This branch accidentally contains my local commits.
So, I close this and resubmit a new PR.

kazu-yamamoto:
@martinthomson @ekr Please read #894.

davegarrett:
For future reference, you can simply force push a replacement set of commits to the same remote branch to change the contents of the PR, as needed. You don't have to refile completely.

kazu-yamamoto:
@davegarrett Thanks!

PR #894, opened 2017-03-06
a server should fall back to a full handshake

kazu-yamamoto:
For #836, I would open this PR (again).

Please discuss whether or not we should use "SHOULD" here.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2585,14 +2585,21 @@ Implementor's note: the most straightforward way to implement the
 PSK/cipher suite matching requirements is to negotiate the cipher
 suite first and then exclude any incompatible PSKs.
 
-Prior to accepting PSK key establishment, the server MUST validate the
-corresponding binder value (see {{psk-binder}} below). If this value is
+When a server can recover a PSK value from identity values, it MUST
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2612,6 +2612,12 @@ facilitates implementation as described below). Servers MUST check
 that it is the last extension and otherwise fail the handshake with an
 "illegal_parameter" alert.
 
+If a server fails to retrieve a PSK value from identity values
+(because of some configuration or state changes, for instance), the
+server cannot resume the previous session nor verify the binders. In
+this case, the server SHOULD fall back to a full 1-RTT handshake if a
+"key_share" extension is supplied.
```
martinthomson: It might be a good idea to note that if the keys CAN be recovered, failing to verify the binders is fatal.

kazu-yamamoto: OK. I will do. Thanks.

kazu-yamamoto: Done.

kazu-yamamoto:
The binder is validated with a key derived from the PSK value.
So, some readers might feel that "before using that PSK" is contradictory.

ekr:
I just pushed a commit which I think resolves the ambiguity satisfactorily.

PR #895, opened 2017-03-06
Remove the word 'session'

katrielalex:
The word "session" is used throughout the document but never really needed; as per https://github.com/tlswg/tls13-spec/issues/847 it only really came from the use of "session resumption". This PR removes all uses of the word "session", replacing them with terms that are more broadly used throughout the document.

I have not renamed NewSessionTicket but would also support `sed s/NewSessionTicket/NewPSKContextTicket` or similar.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -487,6 +487,8 @@ client: The endpoint initiating the TLS connection.
 
 connection: A transport-layer connection between two endpoints.
 
+context: A collection of parameters generated in a handshake (including secrets and negotiated options).
+
 endpoint: Either the client or server of the connection.
 
 handshake: An initial negotiation between client and server that establishes the parameters of their subsequent interactions.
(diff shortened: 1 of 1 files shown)


katrielalex:
The main changes are:

- "session state" -> "connection state"; similarly, things like "the version of TLS negotiated for this session" -> "for this connection"
- "session ticket" -> "ticket"
- "session key" -> "master secret"

draft-ietf-tls-tls13.md
```diff
@@ -2553,7 +2553,7 @@ obfuscated_ticket_age
   configuration that it is using, in milliseconds.  This value is
   added modulo 2^32 to the "ticket_age_add" value that was
   included with the ticket, see {{NewSessionTicket}}.  This addition
-  prevents passive observers from correlating sessions unless tickets
+  prevents passive observers from correlating connections unless tickets
```
katrielalex: I'm not sure about this one. Which things precisely are not meant to be correlated by passive observers?

ekr:
@katrielalex: please send a PR adding yourself to contributors.

PR #897, opened 2017-03-07
Tidy up singular vs. plural session keys.

davidben:
How's this look?

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4878,8 +4878,8 @@ is intended to provide both one-way authenticated (server-only) and
 mutually authenticated (client and server) functionality. At the completion
 of the handshake, each side outputs its view of the following values:
 
-- A set of "session keys" (collectively the various secrets derived from
-  the master secret) from which can be derived a set of working keys.
+- A set of "session keys" (the various secrets derived from the master secret)
+  from which can be derived a set of working keys.
 - A set of cryptographic parameters (algorithms, etc.)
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4878,7 +4878,7 @@ is intended to provide both one-way authenticated (server-only) and
 mutually authenticated (client and server) functionality. At the completion
 of the handshake, each side outputs its view of the following values:
 
-- A "session key" (collectively the various secrets derived from
+- A set of "session keys" (collectively the various secrets derived from
```
martinthomson: drop "collectively" now

draft-ietf-tls-tls13.md
```diff
@@ -4906,7 +4906,7 @@ Peer Authentication.
 identity. If the client is authenticated, the server's view of the
 peer identity should match the client's identity.
 
-Uniqueness of the session key:
+Uniqueness of the session keys:
```
martinthomson: The need for uniqueness applies not to the set, but to every individual key in that set.

draft-ietf-tls-tls13.md

draft-ietf-tls-tls13.md
davidben: Added an extra sentence.

davidben:
PS: I think this is obligatory. :-)

there-are-four-keys

martinthomson:
I was looking for that.

PR #901, opened 2017-03-08
Special-case the hash for CH1 when HRR is used. This allows the

ekr:
server to just store H(CH1) when doing HRR.

@davidben PTAL

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1450,6 +1450,7 @@ processed and transmitted as specified by the current active connection state.
            client_key_exchange_RESERVED(16),
            finished(20),
            key_update(24),
+           message_hash(254),
            (255)
        } HandshakeType;
 
@@ -1986,8 +1987,8 @@ Cookies serve two primary purposes:
 
 - Allowing the server to offload state to the client, thus allowing it to send
   a HelloRetryRequest without storing any state. The server does this by
-  storing the serialized hash state in the cookie (protected
-  with some suitable integrity algorithm).
+  storing the hash of the ClientHello in the HelloRetryRequest cookie
+  (protected with some suitable integrity algorithm).
 
 When sending a HelloRetryRequest, the server MAY provide a "cookie" extension to the
 client (this is an exception to the usual rule that the only extensions that
@@ -2621,7 +2622,7 @@ The PSK binder value forms a binding between a PSK and the current
 handshake, as well as between the handshake in which the PSK was
 generated (if via a NewSessionTicket message) and the handshake where
 it was used.  Each entry in the binders list is computed as an HMAC
-over the portion of the ClientHello (including the handshake header)
+over a transcript hash (see {{the-transcript-hash}}) containing a partial ClientHello
 up to and including the PreSharedKeyExtension.identities field. That
 is, it includes all of the ClientHello but not the binders list
 itself. The length fields for the message (including the overall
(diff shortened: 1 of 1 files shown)


ekr:
@martinthomson

draft-ietf-tls-tls13.md
```diff
@@ -3132,8 +3158,8 @@ and for a client signature is "TLS 1.3, client
 CertificateVerify". It is used to provide separation between signatures
 made in different contexts, helping against potential cross-protocol attacks.
 
-For example, if Hash(Handshake Context + Certificate) was 32 bytes of
-01 (this length would make sense for SHA-256), the content covered by
+For example, if the transcript hash was 32 bytes of
+01 (this length would make sense for SHA-256), the content covered by
```
martinthomson: added trailing space

draft-ietf-tls-tls13.md
```diff
@@ -2889,9 +2887,37 @@ for each scenario:
...
 
-In all cases, the handshake context is formed by concatenating the
-indicated handshake messages, including the handshake message type
-and length fields, but not including record layer headers.
+
+### The Transcript Hash
+
+Many of the cryptographic computations in TLS make use of a transcript
+hash. This value is computed by hashing the concatenation of the
+hashes of each included handshake message, including the handshake
```
martinthomson: hash of the concatenation of hashes isn't right, it's just the hash of the concatenated handshake messages

draft-ietf-tls-tls13.md
```diff
@@ -1450,6 +1450,7 @@ processed and transmitted as specified by the current active connection state.
            client_key_exchange_RESERVED(16),
            finished(20),
            key_update(24),
+           message_hash(254),
```
martinthomson: Is there an IANA section that needs updating?

draft-ietf-tls-tls13.md
```diff
@@ -2639,12 +2640,14 @@ and HelloRetryRequest are included in the transcript along with the
 new ClientHello.  For instance, if the client sends ClientHello1, its
 binder will be computed over:
 
-       ClientHello1[truncated]
+       Transcript-Hash(ClientHello1[truncated])
```
martinthomson: You need to explicitly call out that because this is a *partial* ClientHello1, it doesn't need to be turned into the special message_hash form.

draft-ietf-tls-tls13.md
ekr: Oops. Version skew

dvorak42:
Seems reasonable. Still feels like a bit of a hack, though is better than having to ship the entire hash state in the HRR.

draft-ietf-tls-tls13.md
```diff
@@ -2889,9 +2887,37 @@ for each scenario:
...
+
+     Transcript-Hash(M1, M2, ... MN) = Hash(M1 || M2 ... MN)
+
+As an exception to this general rule, when the server responds to a
+ClientHello with a HelloRetryRequest, the value of ClientHello1 is
+replaced with a special synthetic handshake message of handshake
+type "message_hash" containing Hash(ClientHello1). I.e.,
+
+     Transcript-Hash(ClientHello1, HelloRetryRequest, ... MN) =
+         Hash(254 ||                 // Handshake Type
```
davidben: Nit: 254 => message_hash?

draft-ietf-tls-tls13.md
```diff
@@ -2889,9 +2887,37 @@ for each scenario:
...
+              Hash(ClientHello1) ||  // Hash of ClientHello1
+              HelloRetryRequest ... MN)
+
+The reason for this construction is to allow the server to do a
+stateless HelloRetryRequest by storing just the hash of ClientHello1
+in the cookie, rather than requiring it to export the entire intermediate
+hash state (see {{cookie}}).
+
+In general, implementations can implement the transcript by keeping a
+running transcript hash value based on the negotiated hash. Note,
```
davidben: [This is not entirely true for the client due to the PSK binder issue, but we can defer that.]

ekr: "in general" covers a lot of sins

PR #902, opened 2017-03-08
Warn about eternal ticket extension. Fixes #871

### LINKED ISSUE #871: Guidance on ticket renewal cliff
This is somewhat application-dependent, but ticket-based PSKs with certificate-based auth are by far the most common use of TLS, and this is not immediately obvious: we should include guidance on ticket lifetimes. Specifically, a ticket issued from a resumed connection should take into account the freshness of any CertificateVerify signature being carried over. This will probably translate to a non-renewable lifetime on sessions.

ekr:
@davidben


davidben:
(Post-merge lgtm since I was summoned.)

PR #904, opened 2017-03-10
Add cipher suite to the HRR. This makes it slightly easier for the

ekr:
client to implement because it knows what hash the server will
select. Also clarify the language about HRR and Key Shares.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1807,17 +1807,18 @@ this specification are:
 - key_share (see {{key-share}})
 
 In addition, in its updated ClientHello, the client SHOULD NOT offer
-any pre-shared keys associated with a hash other than that of the selected
-cipher suite. This allows the client to avoid having to compute partial
-hash transcripts for multiple hashes in the second ClientHello.
-A client which receives a cipher suite that was not offered MUST abort the handshake.
-Servers MUST ensure that they negotiate the same cipher suite when
-receiving a conformant updated ClientHello (if the server selects the
-cipher suite first, then this will happen automatically). Upon
-receiving the ServerHello, clients MUST
-check that the cipher suite supplied in the ServerHello is the same
-as that in the HelloRetryRequest and otherwise abort the handshake
-with an "illegal_parameter" alert.
+any pre-shared keys associated with a hash other than that of the
+selected cipher suite. This allows the client to avoid having to
+compute partial hash transcripts for multiple hashes in the second
+ClientHello.  A client which receives a cipher suite that was not
+offered MUST abort the handshake.  Servers MUST ensure that they
+negotiate the same cipher suite when receiving a conformant updated
+ClientHello (if the server selects the cipher suite as the first step
+in the negotiation, then this will happen automatically). Upon
+receiving the ServerHello, clients MUST check that the cipher suite
+supplied in the ServerHello is the same as that in the
+HelloRetryRequest and otherwise abort the handshake with an
+"illegal_parameter" alert.
 
 ##  Extensions
 
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1775,12 +1775,13 @@ Structure of this message:
 
        struct {
            ProtocolVersion server_version;
+           CipherSuite cipher_suite;
            Extension extensions<2..2^16-1>;
        } HelloRetryRequest;
 
 {:br }
 
-The server_version and extensions fields have the
+The version, cipher_suite, and extensions fields have the
 same meanings as their corresponding values in the ServerHello.
 The server SHOULD send only the extensions necessary for the client to
 generate a correct ClientHello pair. As with ServerHello, a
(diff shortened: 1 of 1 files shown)


ekr:
@davidben is this what you had in mind?

draft-ietf-tls-tls13.md
```diff
@@ -2319,8 +2332,8 @@ Servers MUST NOT send a KeyShareEntry for any group not
 indicated in the "supported_groups" extension and
 MUST NOT send a KeyShareEntry when using the "psk_ke" PskKeyExchangeMode.
 If a HelloRetryRequest was received by the client, the client MUST verify that the
-selected NamedGroup matches one that the client sent in its supported_groups
-extension and MUST abort the handshake with an "illegal_parameter" alert otherwise.
+selected NamedGroup in the ServerHello is the same as that in the HelloRetryRequest. If this check
+fails, the client MUST abort the handshake with an "illegal_parameter" alert.
```
davidben: I'm not sure what's going on here. The old text requires that HRR's group is consistent with CH. The new text is requiring that SH's group is consistent with HRR. Both are true, but why the replacement?

ekr: I may be confused, but the graf directly above actually refers to the processing rules for the HRR itself and requires that it be consistent with the CH.

"Upon receipt of this extension in a HelloRetryRequest, the client MUST verify that (1) the selected_group field corresponds to a group which was provided in the "supported_groups" extension in the original ClientHello; and (2) the selected_group field does not correspond to a group which was provided in the "key_share" extension in the original ClientHello. If either of these checks fails, then the client MUST abort the handshake with an "illegal_parameter" alert. Otherwise, when sending the new ClientHello, the client MUST replace the original "key_share" extension with one containing only a new KeyShareEntry for the group indicated in the selected_group field of the triggering HelloRetryRequest."


This paragraph refers to the processing rules for the SH, so that's where we enforce consistency.

ekr:
I did consider bogus binders, but dropping seems more sensible, and it seems like the only impact of the bogus binder would be that if the server *didn't* follow a consistent rule, then you would get a handshake failure at the server. It's not like it could pick them in any case.

davidben:
(Agreed. This version feels way less weird.)

draft-ietf-tls-tls13.md
davidben: Ah, yes, you're right.

ekr:
@prosecco: comments welcome

draft-ietf-tls-tls13.md
```diff
@@ -1805,6 +1806,18 @@ this specification are:
...
+cipher suite. This allows the client to avoid having to compute partial
+hash transcripts for multiple hashes in the second ClientHello.
+A client which receives a cipher suite that was not offered MUST abort the handshake.
+Servers MUST ensure that they negotiate the same cipher suite when
+receiving a conformant updated ClientHello (if the server selects the
+cipher suite first, then this will happen automatically). Upon
+receiving the ServerHello, clients MUST
+check that the cipher suite supplied in the ServerHello is the same
+as that in the HelloRetryRequest and otherwise abort the handshake
+with an "illegal_parameter" alert.
```
martinthomson: Can the client drop cipher suites that the server didn't pick?  I can't see how that would be a problem.  That's not the same as it not *being* a problem, of course.

martinthomson: The checks on ServerHello+HelloRetryRequest consistency are now split between here and below.  Do you want to consolidate them at all?

ekr: I'd rather not do that at this stage in the game.

ekr: Below == KeyShare? No, I think it's better to leave them close to the relevant code point

draft-ietf-tls-tls13.md
```diff
@@ -1805,6 +1806,18 @@ this specification are:
...
 - key_share (see {{key-share}})
 
+In addition, in its updated ClientHello, the client SHOULD NOT offer
+any pre-shared keys associated with a hash other than that of the selected
+cipher suite. This allows the client to avoid having to compute partial
+hash transcripts for multiple hashes in the second ClientHello.
+A client which receives a cipher suite that was not offered MUST abort the handshake.
+Servers MUST ensure that they negotiate the same cipher suite when
+receiving a conformant updated ClientHello (if the server selects the
+cipher suite first, then this will happen automatically). Upon
```
kaduk: "selects the cipher suite first" seems a little informal; maybe "this happens automatically if the server selects a negotiated ciphersuite as the first step of processing a received ClientHello".  Though, hmm, aren't we supposed to look at some extensions before we know what ciphers we can use?

ekr: No, that's not needed any more with orthogonal negotiation

davidben: (Of course, an implementation could use whatever selection algorithm it likes but, yeah, in general it's expected that the server has the entire ClientHello available and potentially negotiates most things before generating HRR. The orthogonal negotiation helps a lot though and makes shortcuts possible if an implementation decides to go with a sane selection algorithm.)

PR #908, opened 2017-03-14
Require implementations to verify record boundary when a key change h…

knekritz:
…appens rather than on receipt of handshake messages.

It is only actually necessary to verify that a handshake message aligns with a key change when changing keys, not when receiving the message, and it can be simpler to implement the check in this way. This change requires senders to align all these messages with a record boundary as before, but only requires recipients to verify the record boundary when the key change actually happens.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3532,12 +3532,12 @@ that:
   types. That is, if a handshake message is split over two or more
   records, there MUST NOT be any other records between them.
 
-- Handshake messages MUST NOT span key changes. Because
-  the ClientHello, EndOfEarlyData, ServerHello, Finished, and KeyUpdate messages
-  can arrive immediately prior to a key change,
-  upon receiving these messages a receiver MUST verify that the end
-  of these messages aligns with a record boundary; if not, then it MUST
-  terminate the connection with an "unexpected_message" alert.
+- Handshake messages MUST NOT span key changes. Implementations MUST verify that
+  all messages immediately preceding a key change align with a record boundary;
+  if not, then they MUST terminate the connection with an "unexpected_message"
+  alert. Because the ClientHello, EndOfEarlyData, ServerHello, Finished, and
+  KeyUpdate messages can immediately precede a key change, implementations MUST
+  send these messages in alignment with a record boundary.
 
 Implementations MUST NOT send zero-length fragments of Handshake
 types, even if those fragments contain padding.


draft-ietf-tls-tls13.md
```diff
@@ -3524,12 +3524,11 @@ that:
...
-  the ClientHello, EndOfEarlyData, ServerHello, Finished, and KeyUpdate messages
-  can arrive immediately prior to a key change,
-  upon receiving these messages a receiver MUST verify that the end
-  of these messages aligns with a record boundary; if not, then it MUST
-  terminate the connection with an "unexpected_message" alert.
+- Handshake messages MUST NOT span key changes. Implementations MUST verify that
+  all messages immediately preceding a key change align with a record boundary.
+  Because the ClientHello, EndOfEarlyData, ServerHello, Finished, and KeyUpdate
+  messages can immediately precede a key change, implementations MUST send these
+  messages in alignment with a record boundary.
```
martinthomson: This loses the alert text, which I would add after the second sentence: "When applying a key change to remote keys, an implementation that has received a partial handshake message MUST generate a fatal "unexpected_message" alert.  

This also doesn't really fix the problem for a ClientHello that precedes a HelloRetryRequest.  I would change the last sentence here to be advisory (the MUST in the first two sentences suffices), as in "ClientHello, .... and KeyUpdate messages all immediate precede a key change."

knekritz: I'll add the alert text back.

I'm not sure I understand your second point. My intent is to require clients to always align these messages with a record boundary (but not necessarily a key change, as in the HRR case).

martinthomson: The title of your commit says "Require implementations to verify record boundary when a key change h...appens rather than on receipt of handshake messages", which I read as saying that you only wanted to enforce the transition on the key change, not at the end of each of these messages (because that it what it says).

I don't care about ClientHello, I care about the key transition that happens after ClientHello.  I don't think that NSS will ever generate a ClientHello that doesn't end with the record, but I don't see any point in adding extra checks.  Especially since they are very hard to add.  If you keep the MUST, there's going to be at least one implementation that is non-compliant (I think, my co-workers sometimes get inappropriately enthusiastic about these things).

davidben: It's not really plausible for a client to fail to end a ClientHello on a record boundary because it completes a flight. I suppose you could predict that you're going to get an HRR and send the first few bytes of CH2 header early, but that's kind of nuts. I'm not particularly offended by trying to enumerate sender obligations by handshake messages if that's the simplest phrasing, but there isn't much point beyond it.

I agree that a MUST-level requirement on the receiver to check them should be tight and exclude the HRR case.

PR #916, opened 2017-03-26
using ClientFinished and ServerFinished for consistency.

kazu-yamamoto:
Server Finished, server Finished, ServerFinished are used.
This patch ensures that only ServerFinished is used.
So as ClientFinished.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -572,7 +572,7 @@ draft-17
 
 - Explicitly allow non-offered extensions in NewSessionTicket
 
-- Explicitly allow predicting ClientFinished for NST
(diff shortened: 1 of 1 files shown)


ekr:
These are the same message, so I'm not comfortable changing the names like this.

I'm happy to have "{client, server} Finished", but not "{Client,Server}Finished'

kazu-yamamoto:
There is one word of ServerFinished in the draft currently.
Do you mean that we should do s/ServerFinished/Server Finished/ instead?

ekr:
Yes

kazu-yamamoto:
I've pushed a new commit with `-f`. Please merge.

PR #921, opened 2017-03-28
Add an extension to negotiate use of post-handshake client auth

martinthomson:
As discussed in Chicago.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2198,7 +2198,7 @@ purpose {{RFC6066}}, but is more complicated, is not used in TLS 1.3
 offering prior versions of TLS).
 
 
-### Post-Handshake Client Authentication (#post_handshake_auth}
+### Post-Handshake Client Authentication {#post_handshake_auth}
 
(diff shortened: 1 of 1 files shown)


ekr:
Hmm... I don't think that the server needs to confirm this, that just seems duplicative. I'd rather just have the client offer and the server knows it can then send. @martinthomson, open to being persuaded otherwise.

davidben:
One minor argument for echoing is the client knows whether it needs to stash the partial transcript, but that may not be compelling enough to bother here. (If we cared about that, it seems better to start the transcript off an exporter value or something...)

kaduk:
I guess there's enough bits to be worked out relating to the semantics of post-handshake server auth (presenting additional certificates) that we don't want to try to shoehorn that into the same extension codepoint.
My instinct would also have been to have the server echo back the signal, but I don't think I can provide any additional concrete argument in support of that.

ekr:
My argument against is: what happens if the client offers, the server refuses, and then requests in any case. Seems like another thing to go wrong.

davidben:
That's true. We'd need to worry about clients tracking that bit and actually enforcing, whereas there's no particular need for it.

PR #924, opened 2017-03-28
add "legacy_" prefix to alert level

davegarrett:
Not used anymore, so label it appropriately.

Pointed out by Ben Kaduk on list.

https://www.ietf.org/mail-archive/web/tls/current/msg22792.html


ekr:
I tend to agree with Kyle Nekritz's point that we don't name it differently elsewhere, so I'm going to leave as-is

PR #931, opened 2017-04-02
Change "Major Differences"

yoavnir:
Make that section actually list the major differences.


davegarrett:
I suggest adding this as a new section and renaming the old one explicitly as "Draft Changelog", as it's still being used for now. Just stick an "RFC EDITOR: PLEASE REMOVE THIS SECTION" at the top of the changelog section and we can keep it until we don't need it anymore.

yoavnir:
Yeah, I'll do this as soon as ekr merges his PR

PR #932, opened 2017-04-03
- Remove support for disallowed extensions (cert_type, user_mapping).

ekr:
- Explicitly define RFC 7250 certificate support.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2971,8 +2971,6 @@ Structure of this message:
 
 %%% Authentication Messages
 
-       opaque ASN1Cert<1..2^24-1>;
-
        struct {
            select(certificate_type){
                case RawPublicKey:
@@ -2980,7 +2978,7 @@ Structure of this message:
                  opaque ASN1_subjectPublicKeyInfo<1..2^24-1>;
 
                case X.509:
-                 ASN1Cert cert_data;
+                 opaque cert_data<1..2^24-1>;
            }
            Extension extensions<0..2^16-1>;
        } CertificateEntry;
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1895,10 +1895,7 @@ appears it MUST abort the handshake with an "illegal_parameter" alert.
 |:-----------------------------------------|------------:|
 | server_name [RFC6066]                    |      CH, EE |
 | max_fragment_length [RFC6066]            |      CH, EE |
-| client_certificate_url [RFC6066]         |      CH, EE |
 | status_request [RFC6066]                 |  CH, CR, CT |
-| user_mapping [RFC4681]                   |      CH, EE |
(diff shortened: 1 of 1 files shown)


kaduk:
Seems reasonable.  (I did not try to audit the rest of the text to see if there are other places that need updating.)

seanturner:
This captures what we talked about in Chicago.

draft-ietf-tls-tls13.md
```diff
@@ -2977,7 +2974,14 @@ Structure of this message:
...
 
        struct {
-           ASN1Cert cert_data;
+           select(certificate_type){
+               case RawPublicKey:
+                 // From RFC 7250 ASN.1_subjectPublicKeyInfo
+                 opaque ASN1_subjectPublicKeyInfo<1..2^24-1>;
+
+               case X.509:
+                 ASN1Cert cert_data;
```
martinthomson: Is this the only place that ASN1Cert is used?  Might pay to inline it here.

draft-ietf-tls-tls13.md
```diff
@@ -2977,7 +2974,14 @@ Structure of this message:
        opaque ASN1Cert<1..2^24-1>;
 
        struct {
-           ASN1Cert cert_data;
+           select(certificate_type){
+               case RawPublicKey:
+                 // From RFC 7250 ASN.1_subjectPublicKeyInfo
+                 opaque ASN1_subjectPublicKeyInfo<1..2^24-1>;
```
martinthomson: Does this really need to be 2^24-1 long?  I assume that we could do with 2^16-1 quite happily.

ekr: I assume so too, but I tend to think better to stick with the 7250 defn.

draft-ietf-tls-tls13.md
```diff
@@ -1895,10 +1895,7 @@ appears it MUST abort the handshake with an "illegal_parameter" alert.
 |:-----------------------------------------|------------:|
 | server_name [RFC6066]                    |      CH, EE |
 | max_fragment_length [RFC6066]            |      CH, EE |
-| client_certificate_url [RFC6066]         |      CH, EE |
```
hannestschofenig: We didn't discuss the client_certificate_url extension during the Chicago IETF meeting but I am wondering whether the considerations for it are somewhat similar than for the cached info extension. In an IoT context you most likely want to use TLS/DTLS for mutual authentication and to avoid sending the client certificate URI extension allows the client to avoid transmitting the full certificate. While it is not as efficient as cached info does for the server certificate it is still a significant improvement over sending the full certificate.

kaduk:
Shucks, this hasn't landed yet, so I'll have to remember to update my editorial changes pull request that puts the numbers for these into the enumerated type definition for extensions.

draft-ietf-tls-tls13.md
ekr: The problem is it largely unused and is also hardwired to a weak hash.

PR #933, opened 2017-04-04
Pha ext

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -3427,23 +3445,27 @@ and the time since the peer's online CertificateVerify signature.
...
 
-Note: Because client authentication may require prompting the user,
-servers MUST be prepared for some delay, including receiving an
-arbitrary number of other messages between sending the
-CertificateRequest and receiving a response. In addition, clients which receive multiple
-CertificateRequests in close succession MAY respond to them in a
-different order than they were received (the
-certificate_request_context value allows the server to disambiguate
-the responses).
+A client that receives a CertificateRequest message without having sent
```
kaduk: Maybe "without" is slightly unclear (but only very slightly); "but did not send" could be better.

seanturner:
lgtm

draft-ietf-tls-tls13.md
```diff
@@ -3427,23 +3445,27 @@ and the time since the peer's online CertificateVerify signature.
...
 ### Post-Handshake Authentication
 
-The server is permitted to request client authentication at any time
-after the handshake has completed by sending a CertificateRequest
-message. The client SHOULD respond with the appropriate Authentication
-messages. If the client chooses to authenticate, it MUST send
-Certificate, CertificateVerify, and Finished. If it declines, it
-MUST send a Certificate message containing no certificates followed by Finished.
+When the client has sent the "post_handshake_auth" extension (see
+{{post_handshake_auth}}), a server can request client authentication at any time
```
davegarrett: "can" probably should be "MAY", here

draft-ietf-tls-tls13.md
ekr: I think I prefer the current version

PR #935, opened 2017-04-10
IANA considerations section tweak.

seanturner:
Need to be explicit say that it's the ExtensionType registry from RFC 4366.


kaduk:
I literally was making the same change in my editor when the notification came in!
(Should it be the "TLS Extensions ExtensionType registry"?)

PR #936, opened 2017-04-10
Mostly editorial

kaduk:
with some other editorial-ish bits split out as separate commits to be looked at separately.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1159,6 +1159,7 @@ as with a 1-RTT handshake with PSK resumption.
                                                     + pre_shared_key
                                                         + key_share*
                                                {EncryptedExtensions}
+                                                       + early_data*
                                                           {Finished}
                                  <--------       [Application Data*]
          (EndOfEarlyData)
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2519,10 +2519,12 @@ fresh.
 0-RTT messages sent in the first flight have the same (encrypted) content types
 as their corresponding messages sent in other flights (handshake and
 application_data) but are protected under
-different keys.  After receiving the server's Finished message, if the
-server has accepted early data, an EndOfEarlyData message
+different keys.  After receiving the server's Finished message,
+an EndOfEarlyData message
 will be sent to indicate the key change. This message will be encrypted
-with the 0-RTT traffic keys.
+with the 0-RTT traffic keys, so the server will be unable to decrypt it
+if the server rejected early data because it did not have tne necessary
+cryptographic key material.
 
 A server which receives an "early_data" extension
 MUST behave in one of three ways:
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3612,7 +3612,7 @@ useful as a traffic analysis countermeasure.
 %%% Record Layer
 
        enum {
-           invalid_RESERVED(0),
+           invalid(0),
            change_cipher_spec_RESERVED(20),
            alert(21),
            handshake(22),
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -903,6 +903,7 @@ TLS supports three basic key exchange modes:
 
 Key  ^ ClientHello
 Exch | + key_share*
+     | + signature_algorithms*
      | + psk_key_exchange_modes*
      v + pre_shared_key*         -------->
                                                        ServerHello  ^ Key
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -865,6 +865,10 @@ TLS 1.2:
 
 * RSASSA-PSS signature schemes are defined in {{signature-algorithms}}.
 
+* The "supported_versions" ClientHello extension can be used for negotiating
+  the version of TLS to use, in preference to the legacy_version field of
+  the ClientHello.
+
 An implementation of TLS 1.3 that also supports TLS 1.2 might need to include
 changes to support these changes even when TLS 1.3 is not in use.  See the
 referenced sections for more details.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3131,7 +3185,8 @@ aspect of the certificate chain was unacceptable (e.g., it was not signed by a
 known, trusted CA), the server MAY at its discretion either continue the
 handshake (considering the client unauthenticated) or abort the handshake.
 
-Any endpoint receiving any certificate signed using any signature algorithm
+Any endpoint receiving any certificate that must be validated and is
+signed using any signature algorithm
```
davegarrett: This is not an editorial nor trivial change. We had consensus for total rejection of MD5 in TLS 1.3+, without the nuance/exceptions/risks we negotiated for SHA1. Something archaic enough to rely on MD5, still, even if it doesn't need to be validated, can be stuck with TLS 1.2 until it gets its act together. This is not to say that some implementation can't ignore the spec and do as you propose (again, there are no IETF police), but we should not be entertaining any legitimate acceptance of MD5 in new specification documents, at this point. If you wish to argue for reopening negotiations on the language here, please propose it on the mailing list with some evidence as to why it might be needed. (If you've got a compelling scenario that can't be fixed otherwise, this is not an unlivable change, but it's not something to stick into a big PR of other fixes/cleanups.)

draft-ietf-tls-tls13.md
```diff
@@ -3333,7 +3388,7 @@ TLS also allows other messages to be sent after the main handshake.
 These messages use a handshake content type and are encrypted under the
 appropriate application traffic key.
 
-### New Session Ticket Message {#NewSessionTicket}
+### New Session Ticket Message {#NSTMessage}
```
davegarrett: Why?

draft-ietf-tls-tls13.md
```diff
@@ -1844,8 +1863,21 @@ A number of TLS messages contain tag-length-value encoded extensions structures.
        } Extension;
 
        enum {
+           server_name(0),
```
kaduk: The motivation here was to match the ones listed in the table that permits them in the various fields.

draft-ietf-tls-tls13.md
kaduk: The links go to the appendix entry otherwise?

draft-ietf-tls-tls13.md
```diff
@@ -1135,7 +1140,7 @@ via a previous handshake), TLS 1.3 allows clients to send data on the
 first flight ("early data"). The client uses the PSK to authenticate
 the server and to encrypt the early data.
 
-When clients use a PSK obtained externally then the following
+When clients use a PSK obtained externally to send early data then the following
```
kaduk: Done.

draft-ietf-tls-tls13.md
```diff
@@ -1159,13 +1164,17 @@ as with a 1-RTT handshake with PSK resumption.
...
+                                                       + early_data*
                                                           {Finished}
                                  <--------       [Application Data*]
          (EndOfEarlyData)
          {Finished}              -------->
 
          [Application Data]      <------->        [Application Data]
 
+              +  Indicates noteworthy extensions sent in the
+                 previously noted message.
```
kaduk: heh, it was vim yank-and-put, so it is the surrounding text that is inconstent between figures.
But, I will go for local consistency, thanks for pointing it out.

draft-ietf-tls-tls13.md
```diff
@@ -2305,7 +2344,9 @@ client_shares
   Each KeyShareEntry value MUST correspond to a group offered in the
   "supported_groups" extension and MUST appear in the same order.  However, the
   values MAY be a non-contiguous subset of the "supported_groups" extension and
-  MAY omit the most preferred groups.
+  MAY omit the most preferred groups, as could happen if the most preferred groups
```
kaduk: I'll make the explanation of potential usage its own sentence.

draft-ietf-tls-tls13.md
```diff
@@ -2750,7 +2800,7 @@ ServerHello message. This is the first message that is encrypted
 under keys derived from the server_handshake_traffic_secret.
 
 The EncryptedExtensions message contains extensions
-which should be protected, i.e., any which are not needed to
+that could be protected, i.e., any which are not needed to
```
kaduk: Sure.

draft-ietf-tls-tls13.md
kaduk: I was semi-remembering discussion about roots that were self-signed with md5, but do not wish to reopen the discussion.
@seanturner do you remember what I might be thinking about?

ekr:
@kaduk: I just merged this. I have one action item you might want to do about annotating the extension code points with the RFC.

draft-ietf-tls-tls13.md
davegarrett: This section here is on how to handle the receipt of a certificate message and its contents, not how to handle the entirety of a certificate chain, including root store. A self-signed root using MD5 would generally not need to be included in the certificate chain sent to the client, and thus would not actually hit the existing text.
https://tools.ietf.org/html/draft-ietf-tls-tls13-19#page-59

I don't really want clients accepting use of MD5 in any scenario, but adding a note restating that these checks do not apply to a trust anchor not sent in this message would be fine. This would not change the existing requirements; just clarify them. Yeah, there's a valid argument to be made that the same checks should apply to all certs regardless of where the client gets them, but the difference here is that this is something that goes through the TLS protocol, specifically (and I don't really want to reopen the discussion at all, though, please do so on list if you disagree).

(Just because it's not clear looking at GitHub here, this PR was merged without this specific changeset.)

kaduk: Yes, they usually would not need to be sent, but they could be sent.
A connection could also be using opportunistic encryption and not plan on validating anything at all.
It seems silly to impose a requirement that "even if you're not going to do any other validation of any certificates, you still need to check for MD5 and reject it".
But, if I want to keep arguing this, I should take it to the list, and I'm not sure that I feel strongly enough to do so.

PR #938, opened 2017-04-11
IDNits reports unused references.

seanturner:
I removed some of the references that IDnits reports as unused.

@ekr Can we delete the AES and GCM references they're really picked up by the cipher suite referenes? We'll need to refer whatever RFC5289 @ PS ends up being for the AES_GCM suites.


ekr:
Hmm.... I think we probably do want them. I'll merge this in the meantime.

PR #946, opened 2017-04-14
Annotate extension code points with RFC number

kaduk:
Show inline which document defines the meaning of that extension,
in addition to listing it in the table of extensions.

Per ekr's comment on  #936 (hopefully)


ekr:
Note to self. Merged locally.

PR #950, opened 2017-04-16
Require (2119 SHOULD) that the certificate context for post-handshake

ekr:
be unpredictable in order to prevent pre-computation of
CertificateVerify. Maybe this should actually be a MUST?


ekr:
Note: an attacker can of course forge a complete handshake but maybe for some quirk of the environment that's not as good.

PR #952, opened 2017-04-16
Move text about PSK interaction with certificate-based client authent…

### LINKED ISSUE #934: CertificateVerify "client's signature with non-certificate-based handshakes"
Section 4.4.3 (CertificateVerify) has:

    Note: When used with non-certificate-based handshakes (e.g., PSK),
    the client's signature does not cover the server's certificate
    directly.  When the PSK was established through a NewSessionTicket,
    the client's signature transitively covers the server's certificate
    through the PSK binder.  [PSK-FINISHED] describes a concrete attack
    on constructions that do not bind to the server's certificate.  It is
    unsafe to use certificate-based client authentication when the client
    might potentially share the same PSK/key-id pair with two different
    endpoints and implementations MUST NOT combine external PSKs with
    certificate-based authentication.

I call attention to the first sentence, which appears to be describing a scenario that cannot happen in the current spec, if I am understanding things correctly.  That is, the server can only send CertificateRequest to the client if the server is authenticating with a certificate, and we don't specify a way to do both PSK and certificate auth yet (i.e., the resumption flow just has a Finished, with no room for Certificate/CertificateVerify).

I conclude, then, that this text is a preemptive warning to people who might want to add such a thing in the future; if so, we should probably have text to more clearly indicate that that is what is going on.

ekr:
…ication. Fixes #934.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3307,17 +3307,6 @@ verification process takes as input:
 If the verification fails, the receiver MUST terminate the handshake
 with a "decrypt_error" alert.
 
-Note: When used with non-certificate-based handshakes (e.g., PSK), the
-client's signature does not cover the server's certificate directly.
-When the PSK was established through a NewSessionTicket, the client's
-signature transitively covers the server's certificate through
-the PSK binder. {{PSK-FINISHED}}
-describes a concrete attack on constructions that do not bind to
-the server's certificate. It is unsafe to use certificate-based client
-authentication when the client might potentially share the same
-PSK/key-id pair with two different endpoints and implementations
-MUST NOT combine external PSKs with certificate-based authentication.
-
 ###  Finished
 
 The Finished message is the final message in the authentication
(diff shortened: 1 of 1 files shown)


ekr:
@kaduk: ^^

draft-ietf-tls-tls13.md
```diff
@@ -5111,6 +5100,19 @@ also provide collision resistance.
...
+message with non-certificate-based handshakes (e.g., PSK).
+If this restriction were to be relaxed in future, the
+client's signature would not cover the server's certificate directly.
+However, if the PSK was established through a NewSessionTicket, the client's
+signature would transitively cover the server's certificate through
+the PSK binder. {{PSK-FINISHED}}
+describes a concrete attack on constructions that do not bind to
+the server's certificate. It is unsafe to use certificate-based client
+authentication when the client might potentially share the same
+PSK/key-id pair with two different endpoints and implementations
```
kaduk: consider a comma before "and"

draft-ietf-tls-tls13.md
```diff
@@ -5111,6 +5100,19 @@ also provide collision resistance.
...
+MUST NOT combine external PSKs with certificate-based authentication.
```
kaduk: I guess I've forgotten whether Russ just wanted to be able to do cert auth with NST-provisioned PSKs or with external PSKs as well.  If the latter, to double-check, we're happy with some future document overriding this MUST NOT?

ekr: I'm not sure what Russ was looking for.

PR #958, opened 2017-04-17
Extensions may provide additional InputSecrets

jschanck:
- Specify that extensions may provide additional InputSecret
  values for incorporation into the key schedule.
- Specify ordering of secrets provided by non-MTI extensions.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4250,15 +4250,21 @@ and length fields, but not including record layer headers. Note that
 in some cases a zero-length HashValue (indicated by "") is passed to
 HKDF-Expand-Label.
 
+An extension may produce an InputSecret value for incorporation into the
+key derivation schedule.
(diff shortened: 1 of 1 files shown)


kaduk:
I am not convinced that we need to build this extension point in right now.  Do you have plans for a specific extension that would use it?

jschanck:
Yes, Douglas and I just posted an ID for an AdditionalKeyShare extension. There's a thread on the mailing list. https://tools.ietf.org/html/draft-schanck-tls-additional-keyshare-00

davegarrett:
Adding this here just confuses things. A new extension can specify whatever changes it needs after the fact, without the need for this here.

jschanck:
The draft says what to do "Given a set of n input secrets" without specifying how one would come to have n input secrets. This just makes the existing functionality more explicit.

ekr:
Moving to closed.

PR #967, opened 2017-04-19
Enhanced the list of TLS 1.3 features

hannestschofenig:


draft-ietf-tls-tls13.md
```diff
@@ -964,27 +964,42 @@ are may minor differences.
 
 - The list of supported ciphersuites has been pruned of all algorithms that
   are considered legacy. Those that remain are all Authenticated Encryption
-  with Associated Data (AEAD) and all offer Perfect Forward Secrecy (PFS).
+  with Associated Data (AEAD). The ciphersuite concept has been changed to 
+  separate the authentication and key exchange mechanisms from the record 
+  protection algorithm (including secret key length) and a hash to be used 
+  with key derivation function.
```
martinthomson: "key derivation function and HMAC."  We still use that function for Finished and Binders.

PR #974, opened 2017-04-21
Added text on determining the validity of OCSP responses

nmav:
The OCSP protocol (RFC6960), defines the validity period of an
OCSP response based on the presence of the thisUpdate and nextUpdate
values. However, the nextUpdate value is optional, and may be omitted
in an OCSP response. In that case it is implied by the OCSP issuer
that an updated OCSP response is available at all time, however,
that does not give a hint to verifier for how long to treat the
previous response valid. Set a maximum value to 15 days.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -39,6 +39,7 @@ normative:
   RFC5869:
   RFC6066:
   RFC6655:
+  RFC6960:
   RFC7539:
   RFC7748:
   RFC7919:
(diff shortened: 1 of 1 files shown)


kaduk:
I think this is more text than is needed (and also one paragraph too late); I would rather just see a minimal reference to RFC 6960 for all OCSP handling.

ekr:
@nmav: can you post this to the list?

nmav:
@ekr I have already. The reason there more text than the RFC6960 reference is the fact that RFC6960 says nothing about the validity time when the nextUpdate field is missing.

kaduk:
> RFC6960 says nothing about the validity time when the nextUpdate field is missing.

Oh.  Ugh.  Maybe this amount of text will be appropriate, then; let's see what happens on the list.

draft-ietf-tls-tls13.md
```diff
@@ -3310,9 +3311,15 @@ CertificateRequest message. If the client opts to send an OCSP response, the
...
-Signed Certificate Timestamp (SCT) as an extension in the ServerHello
-in TLS 1.2 and below.
+Note that the validity period of OCSP responses should be, as per 4.2.2.1
+of {{RFC6960}}, the {thisUpdate, nextUpdate} interval of the OCSP response
+values. For OCSP responses which do not set the nextUpdate field, the validity
+period for the purposes of TLS is set to be 15 days after the time set in the
+'thisUpdate' OCSP response field.
+
+Similarly with the OCSP response extension, {{!RFC6962}} provides a mechanism
+for a server to send aSigned Certificate Timestamp (SCT) as an extension in
```
kroeckx: missing space between "a" and "Signed"

nmav: fixed; thanks

PR #975, opened 2017-04-21
Flesh out discussion of side channels

kaduk:
Add side channel resistance as a desired (though not guaranteed)
property in the introductory material, and provide a dedicated
subsection in the security analysis appendix for further discussion.


kaduk:
Putting side channels in the list of security properties alongside authentication, confidentiality, and integrity still feels a little awkward, as side-channel-resistance is not quite that first class of a goal, but it is still something we pay attention to.

kaduk:
(travis seems to be complaining about "{{ticket-establishment}}", which is not modified by this change.)

ekr:
Hi Ben,

I think this is

PR #977, opened 2017-04-25
Shorter HKDF labels. Fixes #964.

### LINKED ISSUE #964: Labels push HKDF-Expand-Label into two hash blocks
I notice that some of the application traffic secret labels are long enough to push us into a second hash block. I don't think this is a big deal, but I thought I would mention it because now is the time to change.

@adl @martinthomson @davidben

ekr:
Per mailing list discussion, this allows us to have every HKDF-Expand just
have one hash block of info.

@ilaril would love your feedback.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -604,6 +604,9 @@ draft-20
 - Add "post_handshake_auth" extension to negotiate post-handshake authentication
   (*).
 
+- Shorten labels for HKDF-Expand-Label so that we can fit within one
+  compression block (*).
+
 draft-19
 
 - Hash context_value input to Exporters (*)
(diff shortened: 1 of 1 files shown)


ilaril:
Shouldn't "TLS 1.3, " be shortened to "tls13 "? Without that, several
of the labels would seem to blow blocks.

Also, "derived secret" seems unchanged, and is too long to fit into
a block, even with "tls13 " prefix.

ekr:
Oops. Good catch. Script fail.

ekr:
@ilaril: fixed?

draft-ietf-tls-tls13.md
```diff
@@ -4376,7 +4376,7 @@ defined below:
 
     struct {
         uint16 length = Length;
-        opaque label<10..255> = "TLS 1.3, " + Label;
+        opaque label<10..255> = "tls13 " + Label;
```
davegarrett: Maybe also change the max to 18? At minimum, sticking a short note somewhere on the reason for the label lengths would probably be worth doing. This affects interop, so this also needs a changelog entry.

ekr:
Yeah, I'll add a note. I still have an issue for the ChangeLog so I will
add this as well

ilaril:
Also, minimum label size looks to be still 10, presumably should be 7
with 6-byte prefix (as it 10 with 9-byte prefix).

PR #981, opened 2017-04-27
Adding missing "no_application_protocol" alert

hannestschofenig:
RFC 7301 defines the ALPN extension and defined a new alert "no_application_protocol". TLS 1.3 uses ALPN but currently misses the alert in Section 6.


davegarrett:
https://tlswg.github.io/tls13-spec/#error-alerts

You also need to add it to the Error Alerts section with a description.

PR #990, opened 2017-04-27
Reserved confusing

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -2301,14 +2303,22 @@ EdDSA algorithms
...
   {{RFC8032}} or its successors. Note that these correspond to the
   "PureEdDSA" algorithms and not the "prehash" variants.
+
+Legacy algorithms
+: Indicates algorithms which are being deprecated because they use
+  SHA-1 but are still permitted for backward compatibility,
+  specifically RSA using RSASSA-PKCS1-v1_5 and ECDSA.  These values
+  refer solely to signatures which appear in certificates (see
+  {{server-certificate-selection}}) and are not defined for use in
+  signed TLS handshake messages.  lients offering these values (e.g.,
```
kaduk: 'C'lients.

draft-ietf-tls-tls13.md
```diff
@@ -2301,14 +2303,22 @@ EdDSA algorithms
 : Indicates a signature algorithm using EdDSA as defined in
   {{RFC8032}} or its successors. Note that these correspond to the
   "PureEdDSA" algorithms and not the "prehash" variants.
+
+Legacy algorithms
+: Indicates algorithms which are being deprecated because they use
+  SHA-1 but are still permitted for backward compatibility,
```
kaduk: I'm not sure that we specifically need to call out SHA-1 as the only cause.
Maybe something about "because they use weak cryptography"?

draft-ietf-tls-tls13.md
```diff
@@ -2301,14 +2303,22 @@ EdDSA algorithms
...
+  specifically RSA using RSASSA-PKCS1-v1_5 and ECDSA.  These values
```
kaduk: I don't think this statement as written is accurate.
You can use either RSA or ECDSA without SHA-1, which is allowed in the real list of defined algorithms.

draft-ietf-tls-tls13.md
kaduk: Hmm, looks like we are losing the 2119-SHOULD NOT be used, which I would prefer to retain.

ekr: Well, SHA-1 is the weak link here. So I feel like clarify is more accurate.

draft-ietf-tls-tls13.md
ekr: Ah, I was trying to rely on the SHA-1 above but maybe I got too clever.

draft-ietf-tls-tls13.md
ekr: I'll restore a SHOULD NOT

draft-ietf-tls-tls13.md
```diff
@@ -2301,14 +2303,22 @@ EdDSA algorithms
 : Indicates a signature algorithm using EdDSA as defined in
   {{RFC8032}} or its successors. Note that these correspond to the
   "PureEdDSA" algorithms and not the "prehash" variants.
+
+Legacy algorithms
+: Indicates algorithms which are being deprecated because they use
+  algorithms with known weaknesses, specifically SHA-1 when used
+  either with RSA using RSASSA-PKCS1-v1_5 or ECDSA.  These values
```
kaduk: How about "specifically SHA-1, which is used in this context with either RSASSA-PKCS1-v1_5 or ECDSA"?

draft-ietf-tls-tls13.md
```diff
@@ -2301,14 +2303,22 @@ EdDSA algorithms
...
+  refer solely to signatures which appear in certificates (see
+  {{server-certificate-selection}}) and are not defined for use in
+  signed TLS handshake messages. Endpoints SHOULD NOT these algorithms
+  but are permitted to do so solely for backward compatibility. Clients
+  offering these values (e.g., for backwards compatibility) MUST list
```
kaduk: The now-doubled "for backward[s] compatibility" may be overkill.

draft-ietf-tls-tls13.md
ekr: Sure.

draft-ietf-tls-tls13.md
ekr: Good catch.

draft-ietf-tls-tls13.md
kaduk: Looks like "used" got dropped in the merge, FWIW.

PR #995, opened 2017-04-27
Add Joe Salowey as contributor

ekr:


ekr:
@jsalowey: ?

seanturner:
We can fix this later but Salowey.net is better.

ekr:
Fixed

PR #999, opened 2017-05-01
fixing syntax errors of ID 20.

kazu-yamamoto:
My parser detected these.


ekr:
@kazu-yamamoto: is your parser online? We could run it as part of CI...

kazu-yamamoto:
My parser is here:

https://github.com/kazu-yamamoto/tls13-spec-validator

But I think it's hard to run it on CI because it is written in Haskell and I modified the draft a bit to validate coverage:

https://github.com/kazu-yamamoto/tls13-spec/tree/validation

PR #1004, opened 2017-05-02
cite RFC for alert

davegarrett:
Follow up to PR #981. Add a link for the ALPN RFC to the relevant alert's description. (same as with other extension alerts above)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4382,7 +4382,8 @@ certificate_required
 no_application_protocol
 : Sent by servers when a client
   "application_layer_protocol_negotiation" extension advertises
-  protocols that the server does not support.
+  protocols that the server does not support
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4382,7 +4382,8 @@ certificate_required
 no_application_protocol
 : Sent by servers when a client
   "application_layer_protocol_negotiation" extension advertises
-  protocols that the server does not support.
+  protocols that the server does not support
+  (see {{RFC7301}}).
```
martinthomson: 7301 is informative currently, should it be normative?

martinthomson:
There's a bunch of text that specifically mentions ALPN, some of it with normative statements.  If those normative statements can't be understood, optional feature or not, then it's normative.

PR #1005, opened 2017-05-04
Anti replay

ekr:


ekr:
@colmmacc please review.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+storing and retrieving the received ClientHello messages.
+
+### Stateless Time-Based Anti-Replay {#stateless-anti-replay}
+
+Finally, the server can implement a very rough anti-replay mechanism
+merely by measuring the mismatch between client and server views of
+time. The server can determine its view of the age of the ticket by
+subtracting the the time the ticket was issued from the current
+time. If the client and server clocks were running at the same rate,
+the client's view of would be shorter than the actual time elapsed on
```
vdukhovni: Something missing after `client's view of`...

ekr: oops "ticket_age"

draft-ietf-tls-tls13.md
```diff
@@ -3593,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+mechanisms which allow a server to implement a number of limited
+server-side anti-replay defenses. Servers SHOULD implement
+either Single-Use Tickets {{single-use-tickets}} or
+Client Hello Recording {{client-hello-recording}}
+as described
+below, and if not, SHOULD implement the stateless mechanism
+described in {{stateless-anti-replay}}.
+See {{replay-0rtt}} for more information on the limitations
+of these mechanisms.
+
```
enygren: Perhaps mention that clients and application protocols must assume in their security profile (and in-terms of what they are willing to send via 0-RTT) that servers are only implementing stateless-anti-replay?  (When the "server" spans multiple clusters with non-trivial latency between them, stateless-anti-replay is the only one that works and it has the weakest guarantees.)

draft-ietf-tls-tls13.md
```diff
@@ -5594,6 +5670,45 @@ application protocols separately ensuring that confidential
...
+Replayable 0-RTT data presents a number of security threats to
+TLS-using applications. Specifically, if applications are not
+engineered to be idempotent, then duplication of requests
+may cause side effects (e.g., purchasing an item or transferring
+money) to be duplicated, thus harming the site or the user.
+In addition, if data can be replayed a large number of times,
+this enables a variety of attacks via side channels such
+as cache timing or measuring the speed of cryptographic
+operations {{Mac17}}.
+
```
enygren: Perhaps say "side-effect free" rather than "idempotent"?  DELETE and PUT are idempotent but do have side-effects and without additional application layer controls an attacker doing 0-RTT replay could reorder them.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+and these PSKs are deleted upon use, then connections established
+using one PSK enjoy forward security with respect to other PSKs
+established on the same connection. This is a security advantage for
+all 0-RTT data and for PSK usage when PSK is used without DH.
+
+Because this mechanism requires sharing the session database between
+server nodes, it may be hard to achieve high rates of PSK and and
+0-RTT success when compared with self-encrypted tickets which do not
+require consistent server-side storage for basic functionality but
+only for 0-RTT anti-replay.
```
vdukhovni: Perhaps replace `but only for 0-RTT anti-replay` with `, but may require such storage for 0-RTT anti-replay` or something else that makes the sense more clear.

draft-ietf-tls-tls13.md
```diff
@@ -5594,6 +5670,45 @@ application protocols separately ensuring that confidential
...
+{{client-hello-recording}}, prevent a
+replayed ClientHello and its associated 0-RTT data from being accepted
+multiple times by any cluster with consistent state.  However, if
+state is not completely consistent, then an attacker might be able to
+have multiple copies of the data be accepted during the replication
+window. The stateless mechanism described in
+{{stateless-anti-replay}} only prevents replay outside the
+time window.
+
+
```
enygren: Add something like this which covers the fundamental requirement and responsibilities:

"The onus is on clients not to send messages in 0-RTT data which are not safe to have replayed and which they would not be willing to retry across multiple 1-RTT connections.  The onus is on servers to protect themselves against attacks employing 0-RTT data replication."

(or "___ have responsibility to" instead of "the onus is on"?)

draft-ietf-tls-tls13.md
enygren: (This is mentioned below, but repeating it here as well may be worthwhile.)

knekritz: It's also possible to use an instance of Client Hello Recording per cluster in a distributed setup, effectively reducing the # of possible replay for a "Large Number" to 1 per cluster.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+For a given storage window, the server implements anti-replay as
+follows:
+
+1. If the creation time is outside of the window, then accept the PSK but
+   reject 0-RTT.
+
+2. If the ClientHello matches an existing ClientHello, then
+   abort the handshake using an "illegal_parameter" alert
+   (this should never happen in a functional system).
```
knekritz: To make this more efficient implementations can use some hashing/bloom filter rather than storing the entire client hello. This will have false-positives, in which case zero RTT data should just be rejected, not the connection aborted.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+
+3. Otherwise, store the ClientHello during the window
+   and accept 0-RTT.
```
knekritz: It might be wise to explicitly recommend storing the PSK binder rather than the entire ClientHello. This has the benefit of essentially being a compact token cryptographically tied to the 0-RTT key (also preventing someone from polluting the replace cache with random data).

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+
```
colmmacc: It might be adding something here ala ...

It is also critical to the sure that the record of ClientHellos that led to accepted 0-RTT sections from a given window is complete, before accepting any new 0-RTT sections for that same window. For example, if the system recording ClientHellos crashes with no durable record of the ClientHellos previously accepted, then the system needs to wait at least one full window before accepting any ClientHellos.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+Because this mechanism does not require storing all outstanding
+tickets, it may be easier to implement in distributed systems with
+high rates of resumption and 0-RTT, at the cost of potentially
+weaker anti-replay defense because of the difficulty reliably
+storing and retrieving the received ClientHello messages.
+
+### Stateless Time-Based Anti-Replay {#stateless-anti-replay}
```
colmmacc: I think stateless mitigation is pointless, as it doesn't bound the number of replays, but some notes anyway ...

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+therefore delayed, which might be hidden by TCP. For browser clients
+on the Internet, this implies that an
+allowance on the order of ten seconds to account for errors in clocks and
+variations in measurements is advisable; other deployment scenarios
+may have different needs. Outside the selected range, the
+server SHOULD reject early data and fall back to a full 1-RTT
+handshake. Clock skew distributions are not
+symmetric, so the optimal tradeoff may involve an asymmetric range
+of permissible mismatch values.
+
```
colmmacc: Here I would suggest adding  something that says:

Note that while stateless anti-replay can bound over how long in time a packet may be replayed, the total amount of replays tolerated is bounded by bandwidth and system capacity. This can be thousands to billions of replays in real-world settings. 

And I'd argue for adding this too:

Stateless anti-replay SHOULD NOT be used in environments without strong assurance of application and system behavior and MUST NOT be used in environments that must interoperate with third-party systems and applications.

draft-ietf-tls-tls13.md
colmmacc: +1 to side-effect free.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+mechanisms which allow a server to implement a number of limited
+server-side anti-replay defenses. Servers SHOULD implement
+either Single-Use Tickets {{single-use-tickets}} or
+Client Hello Recording {{client-hello-recording}}
+as described
+below, and if not, SHOULD implement the stateless mechanism
+described in {{stateless-anti-replay}}.
+See {{replay-0rtt}} for more information on the limitations
+of these mechanisms.
+
```
colmmacc: Here I would argue for something stronger:

"0-RTT server implementations that must interoperate with third party systems and applications MUST implement a robust anti-replay mechanism". 

My reasoning here is that these a CDN or TLS accelerator that enables 0-RTT without robust anti-replay, will break other downstream systems. (For example upstream 0-RTT leading to throttle exhaustion down stream). I want clear and strong language so that there can be no ambiguity when a CVE is requested against the upstream component. It's not ok imo to break a basic assumption about the internet like that.

draft-ietf-tls-tls13.md
```diff
@@ -489,6 +489,14 @@ informative:
        author:
        -
          ins: H. Krawczyk
+  Mac17:
+       title: "Security Review of TLS1.3 0-RTT"
+       date: 2017
+       target: https://github.com/tlswg/tls13-spec/issues/1001
+       author:
+       -
+         ins: C. MacCarthaigh
```
ekr: Correct, ASCII only. Feel free to supply some other flattening :)

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+  which the ticket was established.
+
+The server can determine the client's view of the age of the ticket by
+subtracting the ticket's "ticket_age_add value" from the
+"obfuscated_ticket_age" parameter in the client's "pre_shared_key"
+extension. The server can determine the approximate time that the
+client sent the ClientHello as:
+
+~~~~
+    creation time + (client's view - RTT estimate/2)
```
ekr: I thought that initially, but I believe that's wrong, because we're interested in the client's claimed sending time.

Consider the case where clocks are globally synchronized and we have a 200 ms RTT.

- We send NST at 0
- The client receives NST at 100ms
- The client waits 1000ms and then sends CH (at 1100)
- The server receives CH at 1200

Now, if the client had put the absolute time in CH, it would have been 1100, but it puts in relative time so that's 1000. When we add 1/2 RTT, we get 1100. If we were to add RTT, we would get 1200, which is wrong, because that's when the *server* got it.

This is different from below where we are interested in the mismatch.

colmmacc: Oh you're right! Wait a second, no. I'm still wrong. But you're not totally right either. If we want to get very pedantic about it (pedantic on a crypto spec? no!) , all we know is that the client is some portion of the RTT behind. We don't know that it's 1/2. The RTT might be asymmetric.  Imagine it's 20ms server -> client, but 180ms client to server. 

* We send NST at 0
* The client receives NST at 20ms
* The client waits 1000ms and then sends CH (at 1020)
* The server receives at 1200

ekr: Oh, totally. It's just the best we can do.

Remember that we're not trusting this value, we're just using it to have the best chance of getting the time within the window we are saving CH for.

ekr: Hmm... Well, we are *sort of* trusting this, because we are using it to distinguish between hard fail and forced 1-RTT. OTOH, the attacker can always force us into that posture by just delaying the packet until its out of window, so I don't think that it's an issue. But it might be easiest to just use "issue time + obfuscated ticket age"....

draft-ietf-tls-tls13.md
martinthomson: We're *almost* able to put people's names in RFCs, but the road is a long one (for reasons that I won't burden you with).

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+
+
+### Client Hello Recording
+
+An alternative form of anti-replay is to record each ClientHello or a
+unique value derived from the ClientHello and reject
+duplicates. However, recording all ClientHellos causes state to grow
+without bound, so in practice the server must instead record
+ClientHellos within a given time window based on the
+"obfuscated_ticket_age" value provided by the client.
```
martinthomson: I would instead say "Recording all ClientHellos causes state to grow without bound.  A server can instead record ClientHellos within a given time window and use the "obfuscated_ticket_age" to ensure that tickets aren't reused outside that window."

draft-ietf-tls-tls13.md
martinthomson: It's half an RTT for sure, but I'm not clear on why you would want to bucket based on when the client creates the ClientHello as opposed to when you receive it.  For any given ticket, the RTT value is a constant, so all you are doing is adding work.

BTW, there's a simpler model for handling ticket ages in general.  The server stores the value `now()+RTT-ticket_age_add` in the ticket (in ekr's example, this is 200-taa).  Then when the server receives the ticket, it extracts that, adds the obfuscated elapsed time value (1000+taa) and compares that to the current time (1200) with whatever allowance for slop you want.

draft-ietf-tls-tls13.md
martinthomson: Ahh hipster crypto to the rescue.

Efficient storage isn't so much the problem as global synchronization within reasonable time frames.

draft-ietf-tls-tls13.md
martinthomson: If the ClientHello needs to be valid, then polluting the cache is as trivial as just creating new and different ClientHello values.  In a way, the binder is just a way of having the other side calculate your hash for you.

draft-ietf-tls-tls13.md
martinthomson: That's probably not going to fly.  The window is large and waiting that long would kill all the benefits that 0-RTT provides.  I would instead recognize that challenges exist in synchronizing state across participating nodes.  See Erik Nygren's comments on the list that amount to basically "not gonna happen", which I agree with.

It's fine to recommend this design, but the need to have globally consistent state is a massive hurdle.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+
+In order to implement this mechanism, a server needs to store the
+following values, either locally or by encoding them in the ticket:
+
+- The time that the server generated the session ticket.
+- The estimated round trip time between the client and server;
+  this can be estimated by measuring the time between sending
+  the Finished message and receiving the first message in the
+  client's second flight, or potentially using information
+  from the operating system.
```
knekritz: I think it would make more sense to just say the server should store one thing:
- The time that the server generated the session ticket, offset by an estimate of the round trip time between client and server.

That also makes it more clear that only information in the ticket can be used for these calculations (any RTT information in the resumption handshake can not be trusted).

draft-ietf-tls-tls13.md
knekritz: I think it is the full round trip you want to add. The server is comparing the timestamp to the server time, at the point that the server has received the ClientHello, so it should be 1200 in your example (right?).

draft-ietf-tls-tls13.md
martinthomson: First of all, I would talk about "actions" rather than requests.  The things that the server does in response to receiving 0-RTT are what will be exploited.

Side-effect free is useful, but forbidding that doesn't really cover it.  In @enygren's example, the side effects are relevant, but the primary effect (creation/update of a resource vs. removal) is what we're really concerned with.  The only way to ensure that this is perfectly safe is to use the "safe" definition in HTTP - that is the request does nothing but generate a response.  And even then, it's rare that such a request is ever free from side effects or side channels.

This risks us defining something that is very HTTP-centric.  I would prefer that we instead say that idempotency is desirable for the actions that the server takes, but that idempotency could be insufficient.  That is more or less what the text here is getting at.

draft-ietf-tls-tls13.md
```diff
@@ -5568,6 +5670,45 @@ application protocols separately ensuring that confidential
...
+client, e.g., because it is in a different cluster which does not
+share state or because the ticket has been deleted as described in
+{{single-use-tickets}}. If the application layer protocol retransmits
+data in this setting, then it is possible for an attacker to induce a
+replay attack by sending the ClientHello to both the original cluster
+(which processes the data immediately) and another cluster which will
+fall back to 1-RTT and process the data upon application layer
+replay. The scale of this attack is limited by the client's
+willingness to replay and therefore only allows a small number of
+replays, which will also use different encryption keys.
```
martinthomson: I would remove the "therefore only allows a small number of replays".  That's all up to the client.  I don't consider 10 to be small, which is where we are at in Firefox.

draft-ietf-tls-tls13.md
martinthomson: This is pretty obvious, but worth stating, I think.

martinthomson:
Hmm, it seems like some of my replies look like new comments.  I clearly fail at GitHib.

ekr:
@martinthomson: I guess I'm writing badly, but you're misunderstanding "store". You accept 0-RTT and then you process the 0-RTT data, but you store a copy of the CH (or the hash) during the window so you don't accept a replay. This doesn't involve delaying the processing of the CH or the EarlyData at all.

martinthomson:
re: store, I was responding to @colmmacc, where he suggested that you have to wait for propagation of storage attempts from all nodes in the cluster.

draft-ietf-tls-tls13.md
ekr: Kyle is right that you need to validate the binder, because it changes the cost of polluting the cache to the cost of getting a new PSK (and you can use PSK-specific filtering to blacklist bad actors).

Maybe it's obvious to others, but we also can't use a hash of the packet because if the CH contains two PSKs, then the attacker can corrupt the second binder without detection and potentially pollute the cache. So, I think you want either CH.Random or the binder.

draft-ietf-tls-tls13.md
knekritz: The window here is just the clock skew tolerance window (ie 10s of seconds). Waiting 1 window before starting to accept  0-RTT data sounds very reasonable to me.

I believe the intent of these is to have 1 of these client hello record caches per cluster, rather than a global state.

draft-ietf-tls-tls13.md
knekritz: Sure the entire client hello can be stored, but why not do it more efficiently?

I don't think this state is global (ie one client hello cache per cluster).

draft-ietf-tls-tls13.md
colmmacc: Sorry, my wording confused things. All I was referring to is that when you restart a strike register with a clean slate, you need to wait the window period of time before accepting any new entries. The reason is because of this race:

T1: Strike register accepts key K
T2: Strike register crashes, loses all memory of key K
T3: Strike register restarts
T4: Strike register accepts K again

To avoid that, the register needs to have a pause on start-up. Or it can record everything durably, but that's very slow. The register can still respond with microseconds during ordinary operation. 
Clearly, I need better wording.

draft-ietf-tls-tls13.md
ekr: We seem to be having trouble reasoning about how this mechanism works, which is probably my fault for bad writing. In any case, the idea here is supposed to be that when you receive a CH, you need to distinguish three cases:

1. I know it's not a replay, because if I saw it I would still have remembered it.
2. I know it's a replay, in which case I should hard fail.
3. I'm not sure if it's a replay, in which case I force you back to 1-RTT.

Let's pretend that the client didn't have an age value in the CH and I have a window of 10 seconds. So, if I receive a CH at T=0, then an attacker can replay it by waiting 11 seconds, by which time I will have flushed it and I will accept the replay. So, the time of receipt is not itself enough.

Conversely say that the client stuffs its view of age in absolute time in the CH and I have a window that is "I keep any packet which claims to be within 10 seconds of the current time either way". If I receive a CH[T=0] at T=0, then I accept it because my window is [-10,10]. Now, if that packet is replayed at T=11, then my window is now [1, 21] and so I have no information about whether this is a replay or not.

Unless I'm badly confused, if the client's clock is crazily wrong here, then what happens is that they get forced into 1-RTT, but it won't cause me to accept a CH as if it had not been replayed when it had, because I'm always retaining any CH with a claimed time within my current window.

The situation is more subtle when the client is reporting relative time, so before discussing that, perhaps we should make sure we all agree that the logic above is right.

knekritz: Yes, that sounds correct (although I think the value of distinguishing 2 and 3 are debatable).

I'm pretty sure what it boils down to is
1. Server stores `<time when generating the NST> + X`
2. Client calculates `<time when sending ClientHello> - <time when NST received>` and sends it as the ticket_age value
3. Server checks that `<time when received ClientHello> - <ticket_age value client provided>` is within a `<clock skew tolerance window>` (note that no RTT information on this resumed connection can be trusted)
And we are trying to figure out what the best value of X is. I'm pretty sure X is the server's estimate of the full RTT.

martinthomson: @knekritz has it right.  The only thing I was pushing on was the fact that you had to bucket based on when the client sent the packet as opposed to when the server received it.  I think that either work equally well because RTT is a constant for any given connection.  (As Kyle points out, you have to use a genuine handshake to estimate it.)

Bucketing on client send time makes things much more complicated to implement.

colmmacc: Estimating the RTT doesn't seem useful. It can vary by quite a lot across connections; on mobile it's all over the place. And if the user is routed between locations, that'll change it too. And clock skew can be more significant. 

I'm not sure how any of this aids security much; the attacker can still generate a lot of accepted replays with this technique.

ekr: > I'm not sure how any of this aids security much; the attacker can still generate a lot of accepted replays with this technique.

How? And specifically, how is it any different in that respect from the single-use session ID approach.

colmmacc: Sorry @ekr , the comments are so deep I got lost on which section we're in. I thought this was about the stateless anti-replay.

enygren: It may be that you always want to force 1-RTT and never hard-fail on detecting replays.  There may be some legitimate scenarios to receive a replay.  For example, TCP FastOpen (or QUIC) combined with 0RTT and packet duplication in the network.  Hard-fail could result in some weird race conditions here.  Having hard-fail (fatal alert, I assume?) as distinct from forcing 1-RTT could also just give an attacker more information.

colmmacc: TCP FO is still TCP - a duplicate packet will be rejected by the TCP state machine and shouldn't make it as far as TLS.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+provided, the server falls back to a full handshake as normal.
+
+If the tickets are not self-contained but rather are database keys,
+and these PSKs are deleted upon use, then connections established
+using one PSK enjoy forward security with respect to other PSKs
+established on the same connection. This is a security advantage for
+all 0-RTT data and for PSK usage when PSK is used without DH.
+
+Because this mechanism requires sharing the session database between
+server nodes, it may be hard to achieve high rates of PSK and and
```
kaduk: "between server nodes in environments with multiple servers acting as endpoints for the same service"

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+allow each session ticket once. In order to implement this, the server
+maintains a database of all outstanding valid tickets; deleting each
+ticket from the database as it is used. If an unknown ticket is
+provided, the server falls back to a full handshake as normal.
+
+If the tickets are not self-contained but rather are database keys,
+and these PSKs are deleted upon use, then connections established
+using one PSK enjoy forward security with respect to other PSKs
+established on the same connection. This is a security advantage for
+all 0-RTT data and for PSK usage when PSK is used without DH.
```
kaduk: I'm not sure this part really fits in the current exposition as-is.  (Also, doesn't the addition of a per-ticket nonce into the PSk ticket derivation give self-contained tickets the same forward secrecy property?)

draft-ietf-tls-tls13.md
kaduk: There should probably be some more clarification that 0-RTT success and PSK success are partially independent, and that tickets can still be used for PSK even if the single-use property cannot be guaranteed; that is, PSK can succeed even in cases where 0-RTT must be rejected for safety.  (Unless I misunderstand?)

draft-ietf-tls-tls13.md
kaduk: MT's version is better, yes.

draft-ietf-tls-tls13.md
kaduk: I tend to agree with Nygren about not giving the attacker more information -- that is, don't hard-fail and just fall back to 1-RTT [unless you're under attack and need to shed load].  The 1-RTT will fail for the attacker's replays, of course.

draft-ietf-tls-tls13.md
kaduk: "Limited" should probably be in the subsection heading.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+NewSessionTicket and ClientHello messages might be retransmitted and
+therefore delayed, which might be hidden by TCP. For browser clients
+on the Internet, this implies that an
+allowance on the order of ten seconds to account for errors in clocks and
+variations in measurements is advisable; other deployment scenarios
+may have different needs. Outside the selected range, the
+server SHOULD reject early data and fall back to a full 1-RTT
+handshake. Clock skew distributions are not
+symmetric, so the optimal tradeoff may involve an asymmetric range
+of permissible mismatch values.
```
kaduk: Once the discussion above about storing received ClientHello(-related stuff) settles, we should probably normalize this text with what we end up with there.

Also, a server could rate-limit how often it accepts 0-RTT, to provide some reduction in the amount of replay possible.  The amount of reduction gained probably is not enough to make it worth doing, but I'll toss it out there.

draft-ietf-tls-tls13.md
```diff
@@ -5568,6 +5670,45 @@ application protocols separately ensuring that confidential
 information is not inadvertently leaked.
 
 
+## Replay Attacks on 0-RTT {#replay-0rtt}
+
+Replayable 0-RTT data presents a number of security threats to
+TLS-using applications. Specifically, if applications are not
+engineered to be idempotent, then duplication of requests
+may cause side effects (e.g., purchasing an item or transferring
+money) to be duplicated, thus harming the site or the user.
```
kaduk: I think this is some HTTPS mindset sneaking on.  (Which is not necessarily wrong, just something to be aware of.)

draft-ietf-tls-tls13.md
kaduk: Again I agree with MT

draft-ietf-tls-tls13.md
```diff
@@ -5568,6 +5670,45 @@ application protocols separately ensuring that confidential
...
+If implemented correctly,
+the mechanisms described in {{single-use-tickets}} and
+{{client-hello-recording}}, prevent a
+replayed ClientHello and its associated 0-RTT data from being accepted
+multiple times by any cluster with consistent state.  However, if
+state is not completely consistent, then an attacker might be able to
+have multiple copies of the data be accepted during the replication
+window. The stateless mechanism described in
+{{stateless-anti-replay}} only prevents replay outside the
+time window.
```
kaduk: probably should reiterate that this can be tens or hundreds of thousands of replays.

draft-ietf-tls-tls13.md
kaduk: Yeah, worth stating again.  Maybe also that the application profile should tell the client to do so.

draft-ietf-tls-tls13.md
kaduk: Ah, the 8th fallacy of naming!  (Not all names are representable in Unicode.)  ((Number probably wrong; I made it up.))

draft-ietf-tls-tls13.md
ekr: No, because compromise of the STEK leads to compromise of all tickets.

kaduk: Oh, database-key tickets certainly have better forward security properties than self-contained tickets.  I'm just not sure about what the "with respect to other PSKs established on the same connection" means.

ekr: I'll see what I can do to rewrite it.

draft-ietf-tls-tls13.md
```diff
@@ -3593,6 +3548,127 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
 In particular, this includes any alerts sent by the
 server in response to client Certificate and CertificateVerify messages.
 
+## 0-RTT and Anti-Replay {#replay-time}
```
isonmad: Not that it matters much, but should this section be labeled `{#anti-replay}` or something else instead? The old `{#replay-time}` section was named that when it only described the stateless time-mismatch mechanism.

draft-ietf-tls-tls13.md
ekr: 1. I'm persuaded by knekrit's argument
2. MT, cute trick, but it doesn't work if you want the window to be on arrival time, so only for the stateless technique, not strike registers.
3. enygren: I was more ambivalent about whether to abort in this round

draft-ietf-tls-tls13.md
russhousley: Aren't the three alternatives in the SHOULD all equal?

Servers need not permit 0-RTT at all, but those which do SHOULD implement either Single-Use Tickets {{single-use-tickets}}, Client Hello Recording {{client-hello-recording}}, or the stateless mechanism described in {{stateless-anti-replay}}.

ekr: This is a WG decision ultimately, but IMO they are not. the first two clearly are stronger.

colmmacc: I'd go further: the third method doesn't work - it doesn't prevent replays. It permits thousands to billions of replays, depending on the amount of bandwidth and hosts available, and it doesn't mitigate several of the attacks. It should be taken out - it's insecure.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+client's view of the age of the ticket by subtracting the ticket's
+"ticket_age_add value" from the "obfuscated_ticket_age" parameter in
+the client's "pre_shared_key" extension. The server can determine the
+"expected arrival time" of the ClientHello as:
+
+~~~~
+    expected_arrival_time = adjusted_creation_time + client's ticket age
+~~~~
+
+For a given storage window, the server implements anti-replay as
```
kaduk: Do we need to introduce the phrase "storage window" somehow (or reword)?

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+follows.
+
+1. Verify the PSK binder.
```
kaduk: "What do I do if the verification fails?"

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+
+2. If the expected_arrival_time is outside the window or the ClientHello
+   matches a known ClientHello then accept the PSK but
+   reject 0-RTT.
+
+3. If the ClientHello matches a known ClientHello then
+   either abort the handshake with an "illegal_parameter" alert
+   or accept the PSK but reject 0-RTT.
```
kaduk: This seems to contradict the last half of (2), which probably should just be "If the expected_arrival_time is outside the window then accept the PSK but reject 0-RTT".

I'd prefer to not give multiple options of behavior without some guidance on when to choose which one.  Ideally there would only be one choice, of course.

Also, "known ClientHello" should be clarified to "a ClientHello that has previously been recorded and entered into the cache" or similar.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+
+4. Otherwise, store the ClientHello as long as its
+   expected_arrival_time is inside the the window and
+   accept 0-RTT.
```
kaduk: comma before "and"?

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+Because this mechanism does not require storing all outstanding
+tickets, it may be easier to implement in distributed systems with
+high rates of resumption and 0-RTT, at the cost of potentially
+weaker anti-replay defense because of the difficulty reliably
+storing and retrieving the received ClientHello messages. In addition,
+servers MAY implement a separate cache for each cluster, thus
+limiting replay to once per cluster. Servers MAY also implement
+data stores with false positives, such as Bloom filters, in
+which case they MUST respond to apparent replay by rejecting
+0-RTT but SHOULD NOT abort the handshake.
```
kaduk: I would be happy to make this a MUST NOT instead of SHOULD NOT.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+
+Note: When implementations are freshly started, they SHOULD
+reject 0-RTT as long as any portion of their storage window overlaps
```
kaduk: ("storage window" also occurs here)

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+the startup time. Otherwise, they run the risk of accepting
+replays which were originally sent during that period.
+
+
+### Stateless Time-Based Anti-Replay {#stateless-anti-replay}
+
+Finally, the server can implement a very rough anti-replay mechanism
```
kaduk: I'd almost call this a "replay reduction" or "replay limitation" mechanism instead of "anti-replay", as "anti-replay" could be read as meaning it is a stronger mechanism than it actually is.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+merely by measuring the mismatch between client and server views of
+time. The server can determine its view of the age of the ticket by
+subtracting the the time the ticket was issued from the current
```
kaduk: "the the"

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+time. If the client and server clocks were running at the same rate,
+the client's view of would be shorter than the actual time elapsed on
+the server by a single round trip time.  This difference is comprised
+of the delay in sending the NewSessionTicket message to the client,
+plus the time taken to send the ClientHello to the server.
```
kaduk: It's a little unfortunate to have the discussion of the time calculations both here and in Client Hello Recording, though I don't have an alternative proposal.

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+There are several potential sources of error that make an exact
+measurement of time difficult. Variations in client and server clock
+rates are likely to be minimal, though potentially with gross time
+corrections.  Network propagation delays are the most likely causes of
+a mismatch in legitimate values for elapsed time.  Both the
+NewSessionTicket and ClientHello messages might be retransmitted and
+therefore delayed, which might be hidden by TCP. For browser clients
+on the Internet, this implies that an
+allowance on the order of ten seconds to account for errors in clocks and
+variations in measurements is advisable; other deployment scenarios
```
kaduk: The more we talk about this I kind of want to drop down to less than 10 seconds, like maybe 5 or even 1.  (Yes, I know the text here is just talking order of magnitude.)

draft-ietf-tls-tls13.md
```diff
@@ -3588,6 +3554,160 @@ appropriate application traffic key as described in {{updating-traffic-keys}}.
...
+may have different needs. Outside the selected range, the
+server SHOULD reject early data and fall back to a full 1-RTT
+handshake. Clock skew distributions are not
+symmetric, so the optimal tradeoff may involve an asymmetric range
+of permissible mismatch values.
+
+Note that while stateless anti-replay can bound over how long in time
+a packet may be replayed, the total amount of replays tolerated is
```
kaduk: "amount" is not pluralizable; use just "amount of replay tolerated"

draft-ietf-tls-tls13.md
```diff
@@ -3611,9 +3731,10 @@ appropriate application traffic key.
 
 ### New Session Ticket Message {#NSTMessage}
 
-At any time after the server has received the client Finished message, it MAY send
-a NewSessionTicket message. This message creates a pre-shared key
-(PSK) binding between the ticket value and the resumption master secret.
+At any time after the server has received the client Finished message,
+it MAY send a NewSessionTicket message. This message creates a
+pre-shared key (PSK) binding between the ticket value and a secret
```
kaduk: "PSK binding" is perhaps a confusing term to use, given that we have "PSK binders" that are different.
Maybe "association" or "relationship"?

draft-ietf-tls-tls13.md
```diff
@@ -5568,6 +5709,48 @@ application protocols separately ensuring that confidential
...
+in some cases may require even stronger conditions),
+then duplication of actions
+may cause side effects (e.g., purchasing an item or transferring
+money) to be duplicated, thus harming the site or the user.
+In addition, if data can be replayed a large number of times,
+this enables a variety of attacks via side channels such
+as cache timing or measuring the speed of cryptographic
+operations {{Mac17}}.
+
+The limited anti-replay described in {{anti-replay}} are intended to
```
kaduk: "anti-replay" is singular and would take "is", but I think this is supposed to be "anti-replay mechanisms".

draft-ietf-tls-tls13.md
```diff
@@ -5568,6 +5709,48 @@ application protocols separately ensuring that confidential
...
+handshake when the server does not have any information about the
+client, e.g., because it is in a different cluster which does not
+share state or because the ticket has been deleted as described in
+{{single-use-tickets}}. If the application layer protocol retransmits
+data in this setting, then it is possible for an attacker to induce a
+replay attack by sending the ClientHello to both the original cluster
+(which processes the data immediately) and another cluster which will
+fall back to 1-RTT and process the data upon application layer
+replay. The scale of this attack is limited by the client's
+willingness to replay and therefore only allows a limited number of
```
kaduk: Is the different-cluster attack limited by the client's willingness to replay?
I guess it relies on the fallback to 1-RTT when the 0-RTT attempt is blackholed, so that's a "yes".

draft-ietf-tls-tls13.md
kaduk: In the main body text, not the security considerations, if I'm reading the line numbers correctly.
The security considerations might do well to reiterate the consequences of billions of replays (e.g., Colm's analysis).

draft-ietf-tls-tls13.md
```diff
@@ -1370,21 +1386,27 @@ keys derived using the offered PSK.
...
 
-Protocols MUST NOT use 0-RTT data without a profile that defines its
-use. That profile needs to identify which messages or interactions are
-safe to use with 0-RTT. In addition, to avoid accidental misuse,
-implementations SHOULD NOT enable 0-RTT unless specifically
-requested. Implementations SHOULD provide special functions for 0-RTT data to ensure
-that an application is always aware that it is sending or receiving
-data that might be replayed.
+
+Clients MUST NOT send messages in early data which are not safe to
```
mnot: Is the target for this requirement really clients? How will they know? I think this is a requirement being placed upon protocols (as below) in how they specify client behaviour.

draft-ietf-tls-tls13.md
davegarrett: Actually, "amount" is pluralizable, just quirky in how it's applied.
A quick search result:  https://english.stackexchange.com/a/254372

I agree that "amounts" isn't valid here, however with the potential to use "replays" to refer to each replayed message and "replay" to refer to the concept of replay, both the current text and this suggestion are valid. English can be weird. I sorta lean slightly towards kaduk's suggested wording, but whatever ekr prefers or a coin flip to chose is fine by me. ;)

draft-ietf-tls-tls13.md
mimoo: Two examples:

* a client browser decides to only send 0-RTT messages for GET requests to the domain root
* a client facebook mobile application decides to only send 0-RTT messages to request new timeline changes

I don't see how you could do that outside the client.

kaduk: I think that this should bind both how protocols specify client behavior and actual client behavior.

draft-ietf-tls-tls13.md
ekr: No, I don't think that's correct. This is a case where you know you have a replay.

kaduk: Sorry, which part do you think is uncorrect?
I so far still hold my preference to not give multiple options without guidance.

ekr: I don't think it contradicts the last half of #2. That is a case of uncertainty, this is not.

kaduk: I would propose, then, to have this text just say "then abort the hanshake with an illegal_parameter alert".  Are there reasons to leave it as is?

PR #1007, opened 2017-05-07
Clarify that EOED is sent iff server accepts early data

bifurcation:


kaduk:
This follows naturally from the inclusion of EndOfEarlyData in the handshake transcript, but it doesn't hurt to mention it explicitly.  See also, uh, 7511e2a259.

bifurcation:
@kaduk I mention it explicitly because it tripped me up in implementation :)

PR #1008, opened 2017-05-07
Add rekey events to the state machine

bifurcation:
In updating [mint](https://github.com/bifurcation/mint) to use this state machine, I found that it was tricky to align re-key events with state transitions, since they happen in between the messages output by a given state transition (e.g., rekeying to the handshake keys in between `ServerHello` and `EncryptedExtensions` on the way out of `NEGOTIATED`).

This PR adds to the state machine diagrams notations of when the keys should be changed on the inbound and outbound sides, in the order in which they need to be sequenced with handshake messages.  It also makes a few more minor edits:

* Adds some missing handshake message sends
* Notes when the server needs to read past decrypt errors due to rejected early data
* Moves "Recv early data" to after finished, since it doesn't affect the state

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4802,21 +4802,23 @@ by IANA:
 This section provides a summary of the legal state transitions for the
 client and server handshakes.  State names (in all capitals, e.g.,
 START) have no formal meaning but are provided for ease of
-comprehension.  Messages which are sent only sometimes are indicated
-in [].
+comprehension.  Actions which are taken only in certain circumstances are
+indicated in [].
 
 ## Client
 
 ~~~~
                            START <----+
             Send ClientHello |        | Recv HelloRetryRequest
-         /                   v        |
+   [Rekey out to early keys] |        | Send ClientHello
+         /                   v        | 
         |                  WAIT_SH ---+
-    Can |                    | Recv ServerHello
-   send |                    V
-  early |                 WAIT_EE
-   data |                    | Recv EncryptedExtensions
-        |           +--------+--------+
+        |                    | Recv ServerHello
+        |                    | Rekey in to handshake keys
+    Can |                    V
+   send |                 WAIT_EE
+  early |                    | Recv EncryptedExtensions
+   data |           +--------+--------+
         |     Using |                 | Using certificate
         |       PSK |                 v
         |           |            WAIT_CERT_CR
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4802,21 +4802,23 @@ by IANA:
...
 
 ## Client
 
 ~~~~
                            START <----+
             Send ClientHello |        | Recv HelloRetryRequest
-         /                   v        |
+   [Rekey out to early keys] |        | Send ClientHello
+         /                   v        | 
         |                  WAIT_SH ---+
```
martinthomson: Having this line attach to the "[Rekey out...]" line appears to have some significance, can you move the '/' down one line?  It's a fault in the original diagram as well...

draft-ietf-tls-tls13.md
```diff
@@ -4829,14 +4831,14 @@ in [].
         |           |                 | Recv CertificateVerify
         |           +> WAIT_FINISHED <+
         |                  | Recv Finished
-        \                  |
+         \                 | Rekey in to application keys
+                           | Rekey out to handshake keys
                            | [Send EndOfEarlyData]
```
martinthomson: IMPORTANT: EOED is sent using 0-RTT keys.  Swap the above two lines.

draft-ietf-tls-tls13.md
```diff
@@ -4850,33 +4852,38 @@ in [].
...
+app data   -->                 | Rekey out to application keys
+after here                 +---+-------------+
+                           |                 |
+                  No 0-RTT |                 | 0-RTT
+Rekey in to handshake keys |                 | Rekey in to early keys
+[Read past decrypt errors] |                 | Read early data
+                           |                 v
+                           |             WAIT_EOED
+                           |                 | Recv EndOfEarlyData
+                           |                 | Rekey in to hs keys
```
martinthomson: maybe be consistent between "hs" and "handshake"

bifurcation: 72-column limitation is a bitch

draft-ietf-tls-tls13.md
bifurcation: Yep, that's a bug.  Will fix.

draft-ietf-tls-tls13.md
```diff
@@ -4802,21 +4802,23 @@ by IANA:
...
                            START <----+
             Send ClientHello |        | Recv HelloRetryRequest
-         /                   v        |
-        |                  WAIT_SH ---+
-    Can |                    | Recv ServerHello
-   send |                    V
-  early |                 WAIT_EE
-   data |                    | Recv EncryptedExtensions
-        |           +--------+--------+
+   [Rekey out to early keys] |        | Send ClientHello
```
isonmad: Does this extra 'Send ClientHello' before returning to START belong here, when it's already going to retry with a second ClientHello after returning to START?

PR #1009, opened 2017-05-07
Add post-handshake state transitions

bifurcation:
If an implementation is using a state machine structure to process incoming handshake messages, then it could be useful to have a similar diagram for state transitions that result from handshake messages that show up after the initial handshake.  For the most part, these are trivial (`CONNECTED -> CONNECTED`), but post-handshake authentication needs the sub-graph with `WAIT_CERT` etc.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4799,11 +4799,11 @@ by IANA:
 
 # State Machine
 
-This section provides a summary of the legal state transitions for the
-client and server handshakes.  State names (in all capitals, e.g.,
-START) have no formal meaning but are provided for ease of
-comprehension.  Messages which are sent only sometimes are indicated
-in [].
+This section provides a summary of the legal state transitions for the client
+and server handshakes and handshake-level interactions that occur after the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4879,6 +4879,28 @@ here         No 0-RTT |                 | 0-RTT
...
+                -- or -- |   |  Certificate |    WAIT_CV
+   Send NewSessionTicket |   |              |       | Recv 
+                -- or -- |   |              |       | CertificateVerify
+   Recv NewSessionTicket |   |              v       V 
+                -- or -- |   |            WAIT_FINISHED
+ Recv CertificateRequest |   |                  | Recv Finished
+        Send Certificate |   |                  |
+[Send CertificateVerify] |   |                  |
+           Send Finished |   |                  |
+                         +-->+<-----------------+
```
martinthomson: A little bit of text here explaining that most of these transitions atomically transition back to CONNECTED would help.  This isn't immediately obvious.

kaduk:
The post-handshake state machine differs for clients and servers (only servers send NST, CR), so I'm not entirely sure that it's helpful to squash all of them together.

PR #1020, opened 2017-05-17
EndOfEarlyData comes after ServerFinished

kaduk:
Remove redundant bits in the table of base key computations.


davidben:
"Later of" is a slightly odd phrasing, but I think it's written this way because EndOfEarlyData isn't always present. Similar with "later of EncryptedExtensions/CertificateRequest".

davegarrett:
Yes, I think this is correct as-is.

ekr:
I also think this is correct.

@kaduk: objections to closing this?

kaduk:
No real objection, though it sort of seems like there is room for increased clarity here.
I just filed this because I made an observation in a posting on the list and didn't want it to get lost.

PR #1021, opened 2017-05-17
Encourage logging alerts, per Kathleen's AD review.

ekr:


briansmith:
I disagree with this. Equally valid advice is to do as little as possible except abort the connection when receiving an alert. Some implementations will not send alerts at all.

ekr:
@briansmith: can you please raise this on the list in response to Kathleen's comments? Thanks

PR #1024, opened 2017-05-24
Move OID Filters extension

martinthomson:
This was the one extension that wasn't in the extensions section, which made it hard to find.

Note I've also removed the indent on the trailing paragraphs to avoid the walloftext problem with the existing text.


isonmad:
Does the part that says

> - If the certificate_extensions list in the CertificateRequest message
>  was non-empty, the end-entity certificate MUST match the extension OIDs
>  recognized by the client, as described in {{certificate-request}}.

need to be updated to point to this new location instead?

martinthomson:
Good catch.  Fixed.

seanturner:
lgtm

PR #1026, opened 2017-06-02
A simpler method to calculate AEAD nonce.

brainhub:
This is a "random start" method with an increment to the lower 64 bits. It is not necessary for the AEAD encryption layer to maintain TLS record counter in the big-endian format and perform an XOR per each record to produce a unique IV.

The security properties of the new nonce calculation method remain the same.

The re-key must happen at the TLS layer, and that's the the layer that should watch for data limits. In any case, the encryption layer can maintain its own counter; however, this counter would be a simple integer that never needs to be serialized to the wire format.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ draft-*.xml
 draft-*.html
 diff-*.html
 lib
+.targets.mk
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ addons:
   apt:
     packages:
      - python-pip
+dist: trusty
 
 install:
  - gem install kramdown-rfc2629
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Group](https://datatracker.ietf.org/wg/tls/documents/) draft of [TLS 1.3]
 
 TLS 1.3 specification:
 * [Editor's copy](https://tlswg.github.io/tls13-spec/)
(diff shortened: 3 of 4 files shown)


ekr:
@brainhub: something is wrong with your diff. Can you rebase....

brainhub:
The content is a single 2ed3733 change. Does it look OK in principle?

Will re-do.

davegarrett:
@brainhub: GitHub apparently sends out email notifications on mention of usernames in commit messages, even on unwatched branches (got one prior to PR submit here). I would recommend avoiding merged mega-summaries in the future. Rebase/reset is generally the simpler way to do things, anyway.

https://github.com/tlswg/tls13-spec/pull/1026/commits/2ed37337d2dc017133efa3ce08cb52ac1f809328?diff=split
As to the actual proposed change: we have very different definitions of "simpler". If you're going to propose a low-level change like this, you really should state why you think a pile of increments and adds is "simpler" than one increment and a XOR. Is this all just to save the 8 bytes of memory for the sequence number that is already used in older versions of TLS? Also, you should do it on the mailing list first, not here. I and others are going to be against any new complexity here, so please provide a solid argument if you think this is actually something worth changing during the WGLC.

Also, typo: "ithe" in second to last sentence in the meaningful commit.

Lastly, dear lord, no pointer math in an example, no matter how benign you think it may be.

brainhub:
I have re-created #1027. I have added additional points to the comments. Sorry about the mess. Please re-review there.

PR #1027, opened 2017-06-02
A simpler method to calculate AEAD nonce

brainhub:
This is a "random start" method with an increment to the lower 64 bits. It is unnecessary for the AEAD encryption layer to maintain TLS record counter in the big-endian format and perform an XOR per each record to produce a unique IV.

The security properties of the new nonce calculation method remain the same. Security arguments regarding the nonce space are now a bit cleaner (nonce now is essentially a counter).

The re-key must happen at the TLS layer, and that's the the layer that should watch for data limits. In any case, the encryption layer can optionally maintain its own counter. However, this counter would be a simple integer that never needs to be serialized to the wire format.

Before this change the encryption layer:
- must maintain a sequence counter;
- must serialize it;
- must XOR 8 bytes with the mask.

After this change the encryption layer:
- increments a single byte in 255 out of 256 cases;
- has minimal state (12 bytes for AES-GCM).
Other implementation techniques are possible, e.g. in 3-intrinsics swap/increment/swap on x86.

Some of this work is influenced by the FIPS 140-2 mandate to make IV management internal, per sec. A.5 http://csrc.nist.gov/groups/STM/cmvp/documents/fips140-2/FIPS1402IG.pdf . Getting rid of TLS sequence number is arguably more "FIPSable".

The method is fully compatible with TLS 1.2.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4011,37 +4011,48 @@ maximum AEAD expansion of 255 octets.
 
 ## Per-Record Nonce {#nonce}
 
-A 64-bit sequence number is maintained separately for reading and writing
-records.  Each sequence number is set to zero at the beginning of a connection
-and whenever the key is changed.
+The length of the per-record nonce (iv_length) is set to max(8 bytes,
+N_MIN) for the AEAD algorithm (see {{RFC5116}} Section 4). An AEAD
+algorithm, where N_MAX is less than 8 bytes, MUST NOT be used with TLS.
 
-The appropriate sequence number is incremented by one after reading or writing each record.
-The first record transmitted under a particular set of traffic keys
-MUST use sequence number 0.
+The TLS session maintains a variable of iv_lenth bytes, called here an iv,
+separately for reading and writing records. One iv is initialized with client_write_iv
+and another -- with server_write_iv, depending on the role. The initializion
+is done once per TLS session.
 
-Because the size of sequence numbers is 64-bit, they should not
-wrap. If a TLS implementation would need to
-wrap a sequence number, it MUST either re-key ({{key-update}}) or
-terminate the connection.
+The per-record nonce for the AEAD construction is managed as follows for each
+of the two ivs:
 
-Each AEAD algorithm will specify a range of possible lengths for the
-per-record nonce, from N_MIN bytes to N_MAX bytes of input ({{RFC5116}}).
-The length of the TLS per-record nonce (iv_length) is set to the larger of
-8 bytes and N_MIN for the AEAD algorithm (see {{RFC5116}} Section 4). An AEAD
-algorithm where N_MAX is less than 8 bytes MUST NOT be used with TLS.
-The per-record nonce for the AEAD construction is formed as follows:
-
-  1. The 64-bit record sequence number is encoded in network byte order
-     and padded to the left with zeros to iv_length.
-
-  2. The padded sequence number is XORed with the static client_write_iv
-     or server_write_iv, depending on the role.
-
-The resulting quantity (of length iv_length) is used as the per-record
+   1. The iv is incremented as described later in this session. 
+   2. The resulting iv (of length iv_length) is used as the per-record
 nonce.
 
-Note: This is a different construction from that in TLS 1.2, which
-specified a partially explicit nonce.
+The following steps are performed per each record.
+
+In mathematical notation, treating iv as an integer < 2^(iv_length*7):
+
+       t0 = iv mod 2^64
+       t1 = t0 + 1 mod 2^64
+       iv = iv - t0 + t1
+
+In C code:
+
+       void inc64_be( uint8_t *p ) {
+          int i;
+          for (i = 7; i >= 0 && !(p[i]++); i--);
+       }
+       ...
+       uint8_t iv[iv_length];   /* session state */
+       ...
+      /* generate the next nonce for each record */
+      inc64_be( iv + iv_length - 8 );  
+
+This is a different construction from that in TLS 1.2, which
+specified a partially explicit nonce. 
+
+The method assumes that the limits on key usage are followed. The increment
+occures before the very first encrypted record, and continues for each
+record.
 
 
 ## Record Padding
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4027,12 +4027,13 @@ of the two ivs:
    2. The resulting iv (of length iv_length) is used as the per-record
 nonce.
 
-The following steps are performed per each record.
+Once the initial iv is set, the trailing 64 bits of iv are incremented by 1
+as a big-endian integer, per each record.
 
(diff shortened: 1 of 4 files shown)


davegarrett:
@brainhub: You don't need to post replacement PRs. Just edit the branch and (force) push to GitHub. The PR will be updated as needed. (takes a bit to learn the best way to manage these things) Ideally, a PR should contain only the commits that do changes relevant to that PR; there shouldn't be any merge commits here. One of the ways to do this:
0) Use git via the command line instead of GitHub's web interface. (you can of course do these things with it from the start, but cleaning up an existing PR requires full functionality)
1) Create a new branch off of a fully updated master upstream (the current state of the official repo)
2) Apply just your patch on top of that as one commit. ("git cherry-pick" the old one or manually make a new one or save a patch and apply it to the new branch)
3) Use diff to check that the branch contains exactly the intended changes and use log to check that you have only one commit beyond the last merged one up here.
4) Delete your old branch for this PR, and rename this new branch to the old name ("random_start_nonce" or "nonce_calculation", whichever PR you're updating).
5) Force push it to GitHub to replace the PR's contents.
6) Close whichever PR you didn't update.

Also, automated tests have failed for some reason, so there's something broken in the markup somewhere. I don't immediately see where the error is, though.

Again, however, this needs to be taken to the mailing list.

draft-ietf-tls-tls13.md
```diff
@@ -4011,37 +4011,48 @@ maximum AEAD expansion of 255 octets.
...
+      /* generate the next nonce for each record */
+      inc64_be( iv + iv_length - 8 );  
+
+This is a different construction from that in TLS 1.2, which
+specified a partially explicit nonce. 
 
-Note: This is a different construction from that in TLS 1.2, which
-specified a partially explicit nonce.
+The method assumes that the limits on key usage are followed. The increment
+occures before the very first encrypted record, and continues for each
```
martinthomson: occurs

draft-ietf-tls-tls13.md
```diff
@@ -4011,37 +4011,48 @@ maximum AEAD expansion of 255 octets.
...
-nonce.
+       t0 = iv mod 2^64
+       t1 = t0 + 1 mod 2^64
+       iv = iv - t0 + t1
+
+In C code:
+
+       void inc64_be( uint8_t *p ) {
+          int i;
+          for (i = 7; i >= 0 && !(p[i]++); i--);
```
martinthomson: If this is correct, then see the above comment.  Otherwise, if `iv[4..11]` have all bits set, then this will not increment `iv[3]`.

draft-ietf-tls-tls13.md
```diff
@@ -4011,37 +4011,48 @@ maximum AEAD expansion of 255 octets.
...
 
-  2. The padded sequence number is XORed with the static client_write_iv
-     or server_write_iv, depending on the role.
+In mathematical notation, treating iv as an integer < 2^(iv_length*7):
 
-The resulting quantity (of length iv_length) is used as the per-record
-nonce.
+       t0 = iv mod 2^64
+       t1 = t0 + 1 mod 2^64
+       iv = iv - t0 + t1
```
martinthomson: Isn't this just a convoluted way of saying `iv += 1` ?  Unless you intend to discard the overflow from the addition on the second line.  If that is your intent, then you need to take `t1 mod 2^64` somewhere explicitly, or add parentheses around `t0 + 1`.

davidben:
We also already standardized the XOR scheme in RFC 7905, with the intention that TLS 1.3 would match.

draft-ietf-tls-tls13.md
brainhub: It's 

t1 = t0 + 1 (mod 2^64) or
t1 = (t0 + 1) mod 2^64

In C-like language:

uint128_t iv = XXX;
uint128_t m = (1<64)-1;
iv = (iv & ~m) | ((iv + 1) & m);

draft-ietf-tls-tls13.md
brainhub: The intent is not to propagate the carry into iv[3], iv[2], ... iv[0].

For TLS 1.3 we could have done simply client_write_iv++, server_write_iv++ and be done. 

All above the complexity is entirely for the TLS 1.2 compatibility, which requires preservation of the 4 byte implicit part. I think that it would be bad to make the encryption layer behave differently depending on the TLS protocol version.

brainhub:
@davidben: I wasn't aware of [rfc7905](https://tools.ietf.org/html/rfc7905). 

Ideally the TLS 1.3 spec should be the authoritative document on TLS 1.3 nonce creation and rfc7905 refer to it by saying "use the nonce as formed per the TLS 1.3 spec" with minimal management.

With my proposal the interface between layers is straightforward. AEAD cipher layer does "iv++, preserving the top bytes".  The initial iv is set once, along with the key. There is no padding, XORing, no concept of TLS records to explain.

martinthomson:
We can't exactly change RFC 7905 now.

The layering isn't exactly complex right now.  The cipher layer maintains a counter and a baseline IV.  There's a ++ and an XOR on every encryption, which I can agree is a teensy bit more work, but that's a sunk cost given RFC 7905.

agl:
The full sequence number already has to be maintained, so there's no savings there. Indeed, incrementing a single byte and checking whether the next needs to be incremented will be slower.

XORing with the mask and writing out is likewise trivial, and all of this is immeasurably small compared to the work of the AEAD itself.

Thus not inventing another nonce scheme is the overriding interest here.

ekr:
I agree with @agl and @martinthomson. As @davegarrett says, if you want to pursue this your next avenue is the list.

brainhub:
The sequence doesn't have to be maintained at the encryption layer for nonce calculation with this method. We are discussing this in the context of a brand new scheme for TLS 1.3.

agl:
But we don't want a brand new scheme for TLS 1.3. RFC 7905 was done in order to avoid having another nonce construction for TLS 1.3

> The sequence doesn't have to be maintained at the encryption layer for nonce calculation with this method. 

I'm afraid that I don't understand your point here. The sequence already isn't maintained at the encryption layer in typical TLS designs, rather they specify it for each record using the [standard AEAD interface](https://tools.ietf.org/html/rfc5116#section-2.1).

brainhub:
@agl: 

FIPS 140-2 IG A.5 tells that the encryption layer must manage the nonces itself. That's one design for the interface. In this case the doing iv++ (and ignoring the carry into top bytes) inside the encryption  is simpler.

What you mean is that the higher layer, TLS layer in this case, can build the nonce for each record and pass it to encryption layer. That's another design. Yes, there is record number in this case for nonce construction, but then again, the iv++ is simpler for this case as well.

ekr:
@brainhub: as noted in contributing.md "Note that github issues are not for substantial discussions; the only appropriate place to discuss design issues is on the mailing list itself."

This is clearly a substantive issue, so if you want it considered, you need to take it to the list and get consensus.

ekr:
Closing. Feel free to reopen on list

PR #1034, opened 2017-06-23
Add extensive discussion of 0-RTT anti-replay as suggested by

ekr:
Colm MacCarthaigh. Specifically:

- Describe both one time tickets and client hello storage
  ("strike register") mechanisms and SHOULD-level require
  people to do them.

- Provide a security considerations section describing the
  threats.

This is version 2 of this PR after extensive discussion on the
list.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5801,7 +5801,7 @@ client, e.g., because it is in a different cluster which does not
 share state or because the ticket has been deleted as described in
 {{single-use-tickets}}. If the application layer protocol retransmits
 data in this setting, then it is possible for an attacker to induce
-message duplicatio nby sending the ClientHello to both the original cluster
+message duplication by sending the ClientHello to both the original cluster
 (which processes the data immediately) and another cluster which will
 fall back to 1-RTT and process the data upon application layer
 replay. The scale of this attack is limited by the client's
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5840,6 +5840,21 @@ allow applications to implement this kind of processing, TLS
 implementations MUST provide a way for the application to determine if
 the handshake has completed.
 
+### Replay and Exporters
+
+Replays of the ClientHello produce the same early exporter, thus
+requiring additional care by applications which use these exporters.
+In particular, if these exporters are used as an authentication
+channel binding (e.g., by signing the output of the exporter)
+an attacker who compromises the PSK can transplant authenticators
+between connections without compromising the authentication key.
+
+In addition, the early exporter SHOULD NOT be used to generate
+server-to-client encryption keys because that would entail
+the reuse of those keys. This parallels the use of the early
+application traffic keys only in the client-to-server direction.
+
+
 # Working Group Information
 
 The discussion list for the IETF TLS working group is located at the e-mail
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4753,7 +4753,7 @@ storing and retrieving the received ClientHello messages.
 In many such systems, it is impractical to have globally
 consistent storage of all the received ClientHellos.
 In this case, the best anti-replay protection is provided by
-having single storage zone be
+having a single storage zone be
 authoritative for a given ticket and refusing 0-RTT for that
 ticket in any other zone. This approach prevents simple
 replay by the attacker because only one zone will accept
@@ -5835,7 +5835,7 @@ data if it is rejected by the server unless instructed by the
 application. Server-side applications may wish to implement special
 processing for 0-RTT data for some kinds of application traffic (e.g.,
 abort the connection, request that data be resent at the application
-layer, or delay processing until the handshake completes) In order to
+layer, or delay processing until the handshake completes). In order to
 allow applications to implement this kind of processing, TLS
 implementations MUST provide a way for the application to determine if
 the handshake has completed.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4675,7 +4675,7 @@ sections.
 
 The second class of attack cannot be prevented at the TLS layer and
 MUST be dealt with by any application. Note that any application whose
-clients implement any kind of retry behavior already needed to
+clients implement any kind of retry behavior already needs to
 implement some sort of anti-replay defense.
 
 In normal operation, clients will not know which, if any, of these
@@ -4687,15 +4687,15 @@ described in {{replay-0rtt}}.
 ## Single-Use Tickets
 
 The simplest form of anti-replay defense is for the server to only
-allow each session ticket to be used once. In order to implement this, the server
-maintains a database of all outstanding valid tickets; deleting each
+allow each session ticket to be used once. For instance, the server
+can maintain a database of all outstanding valid tickets; deleting each
 ticket from the database as it is used. If an unknown ticket is
-provided, the server falls back to a full handshake.
+provided, the server would then fall back to a full handshake.
 
 If the tickets are not self-contained but rather are database keys,
 and the corresponding PSKs are deleted upon use, then connections established
 using one PSK enjoy forward security. This improves security for
-all 0-RTT data and PSK usage when PSK is used without DH.
+all 0-RTT data and PSK usage when PSK is used without (EC)DHE.
 
 Because this mechanism requires sharing the session database between
 server nodes in environments with multiple distributed servers,
@@ -4719,7 +4719,7 @@ tickets aren't reused outside that window.
 In order to implement this, when a ClientHello is received, the server
 first verifies the PSK binder as described
 {{pre-shared-key-extension}}. It then computes the
-expected_arrival_time as described in the next section and and rejects
+expected_arrival_time as described in the next section and rejects
 0-RTT if it is outside the recording window, falling back to the
 1-RTT handshake.
 
@@ -4727,8 +4727,8 @@ If the expected arrival time is in the window, then the server
 checks to see if it has recorded a matching ClientHello. It
 either aborts the handshake with an "illegal_parameter" alert
 or accepts the PSK but reject 0-RTT. If no matching ClientHello
-is found, then it stores the ClientHello as long as the
-expected_arrival_time is inside the window and accepts 0-RTT.
+is found, then it accepts 0-RTT and then stores the ClientHello for
+as long as the expected_arrival_time is inside the window.
 Servers MAY also implement data stores with false positives, such as
 Bloom filters, in which case they MUST respond to apparent replay by
 rejecting 0-RTT but MUST NOT abort the handshake.
@@ -4737,11 +4737,13 @@ The server MUST derive the storage key only from validated sections
 of the ClientHello. If the ClientHello contains multiple
 PSK identities, then an attacker can create multiple ClientHellos
 with different binder values for the less-preferred identity on the
-assumption that the server will not verify it. I.e., if the
+assumption that the server will not verify it, as recommended
+by {{pre-shared-key-extension}}.
+I.e., if the
 client sends PSKs A and B but the server prefers A, then the
 attacker can change the binder for B without affecting the binder
 for A. If the validated binder or the ClientHello.random
-are used for this purpose, then this attack is not possible.
+are used as the storage key, then this attack is not possible.
 
 Because this mechanism does not require storing all outstanding
 tickets, it may be easier to implement in distributed systems with
@@ -4750,8 +4752,9 @@ weaker anti-replay defense because of the difficulty of reliably
 storing and retrieving the received ClientHello messages.
 In many such systems, it is impractical to have globally
 consistent storage of all the received ClientHellos.
-In this case, the stronger design is to have a single storage zone be
-authoritative for a given ticket and refuse 0-RTT for that
+In this case, the best anti-replay protection is provided by
+having single storage zone be
+authoritative for a given ticket and refusing 0-RTT for that
 ticket in any other zone. This approach prevents simple
 replay by the attacker because only one zone will accept
 0-RTT data. A weaker design is to implement separate storage for
@@ -4772,11 +4775,11 @@ the second form of attack described above.
 
 
 
-## Relevance Checks
+## Freshness Checks
 
 Because the ClientHello indicates the time at which the client sent
 it, it is possible to efficiently determine whether a ClientHello was
-likely sent reasonably recently and only accept 0-RTT for such
+likely sent reasonably recently and only accept 0-RTT for such a
 ClientHello, otherwise falling back to a 1-RTT handshake.
 This is necessary for the ClientHello storage mechanism
 described in {{client-hello-recording}} because otherwise the server
@@ -4823,7 +4826,7 @@ may have different needs. Clock skew distributions are not
 symmetric, so the optimal tradeoff may involve an asymmetric range
 of permissible mismatch values.
 
-Note that relevance checking alone is not sufficient to prevent replays
+Note that freshness checking alone is not sufficient to prevent replays
 because it does not detect them during the error window, which,
 depending on bandwidth and system capacity could include
 billions of replays in real-world settings.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,193 @@ a context in all exporter computations, though the value could be empty.
...
+single-use tickets because it allows efficient rejection of ClientHellos
+which cannot be used for 0-RTT.
+
+In order to implement this mechanism, a server needs to store the time
+that the server generated the session ticket, offset by an estimate of
+the round trip time between client and server. I.e.,
+
+~~~~
+    adjusted_creation_time = creation_time + estimated_RTT
+~~~~
```
colmmacc: I think this is really is the "estimated_max_RTT". If a client happens to have a 200ms RTT when the ticket is issued, that doesn't mean they'll show up with a 200ms RTT the next time. For Satellite and Mobile users,  the variance can be quite a lot; ranging between 50ms and 500ms isn't unusual on some networks. 

It would probably be too strict to reject 0-RTT just because the user got an unlucky connection. Especially since the slow connections benefit the most from 0-RTT.

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+1-RTT connections.
+
+Application protocols MUST NOT use 0-RTT data without a profile that
+defines its use. That profile needs to identify which messages or
+interactions are safe to use with 0-RTT and how to handle the
+situation when the server rejects 0-RTT and falls back to 1-RTT.
+
+In addition, to avoid accidental misuse, TLS implementations MUST NOT
+enable 0-RTT (either sending or receiving) unless specifically
+requested by the application and SHOULD NOT automatically resend 0-RTT
```
vasilvv: So, I think this is less a matter of should vs should not, and more of the matter of "impossible while making a working TLS implementation".  When the client retries the connection, exporter secrets are guaranteed to change, and other TLS-negotiated parameters can change too, so retrying a request can result in incorrect application-layer payload unless the implementation is minimal enough to not provide any of those facilities.

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+data if it is rejected by the server unless instructed by the
+application.  In order to allow applications to implement special
+processing for 0-RTT data, when accepting 0-RTT as a server, a TLS
```
vasilvv: Would it be helpful to spell out the possible options for "special processing"?  Regardless of protocol, I see three possible options: delay processing until liveness confirmation, request the data to be resent, and abort the connection with a fatal error.

draft-ietf-tls-tls13.md
martinthomson: Actually, it's better if the server estimates low here, so the minimum is actually better.   Besides, the maximum RTT is essentially unbounded.

As we discovered, if you are using a system with false positives is vulnerable to a one-time replay if the client clock is super fast.  Basically, you reject 0-RTT for a ClientHello that is purportedly from the future, but then that ClientHello can be replayed (and 0-RTT accepted) once the time enters the window. Increasing your RTT estimate causes that particular error to be more likely.

draft-ietf-tls-tls13.md
martinthomson: This seem to be good choices and probably worth enumerating.

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,193 @@ a context in all exporter computations, though the value could be empty.
...
+from the ClientHello (generally either the random value or the PSK
+binder) and reject duplicates. Recording all ClientHellos causes state
+to grow without bound, but a server can instead record ClientHellos within
+a given time window and use the "obfuscated_ticket_age" to ensure that
+tickets aren't reused outside that window.
+
+In order to implement this, when a ClientHello is received, the server
+first verifies the PSK binder as described
+{{pre-shared-key-extension}}. It then computes the
+expected_arrival_time as described in the next section and and rejects
```
chris-wood: "and and"

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+
+If data can be replayed a large number of times, additional attacks
+become possible. Specifically, attackers can use multiple replays to
+exploit information leakage via side channels such as timing network
+caches or measuring the speed of cryptographic operations. In addition, they may
+be able to overload rate-limiting systems. For further description of
+these attacks, see {{Mac17}}.
+
+Ultimately, servers have the responsibility to protect themselves
+against attacks employing 0-RTT data replication. The mechanism
```
chris-wood: "mechanism" -> "mechanisms"

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,193 @@ a context in all exporter computations, though the value could be empty.
...
+described in {{replay-0rtt}}.
+
+
+## Single-Use Tickets
+
+The simplest form of anti-replay defense is for the server to only
+allow each session ticket to be used once. In order to implement this, the server
+maintains a database of all outstanding valid tickets; deleting each
+ticket from the database as it is used. If an unknown ticket is
+provided, the server falls back to a full handshake.
```
chris-wood: This is one way to implement the single-use ticket strategy. Perhaps this paragraph could be rephrased so as to not sound so prescriptive?

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,193 @@ a context in all exporter computations, though the value could be empty.
...
+
+Note: If the client's clock is running much faster than the server's
+then a ClientHello may be received that is outside the window in the
+future, in which case it might be accepted for 1-RTT, causing a client retry,
+and then acceptable later for 0-RTT. This is another variant of
+the second form of attack described above.
+
+
+
+## Relevance Checks
```
chris-wood: "Relevant" doesn't seem to capture the intent here. Perhaps this could be renamed to "Freshness Checks" or something similar?

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
 information is not inadvertently leaked.
 
 
+## Replay Attacks on 0-RTT {#replay-0rtt}
+
+Replayable 0-RTT data presents a number of security threats to
+TLS-using applications. Specifically, if applications are not
+engineered to be replay safe (minimally, this means idempotent, but in
+some cases may there might be even stronger conditions), then duplication of
```
kaduk: Past discussions seem to indicate that there are stronger conditions in most cases, not just some.
I think we should consider adding an enumeration of potential other issues, like side channel leakage, DoS/rate-limiting, reordering, etc.  Which actually looks a lot like the next few paragraphs, and...

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+actions will cause side effects (e.g., purchasing an item or
+transferring money) to be duplicated, thus harming the site or the
+user.  Attackers can also store and replay 0-RTT messages in order to
+re-order them with respect to other messages, which may harm
+unsuspecting applications.
```
kaduk: ... which would then render this sentence redundant.

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+
+If data can be replayed a large number of times, additional attacks
+become possible. Specifically, attackers can use multiple replays to
+exploit information leakage via side channels such as timing network
+caches or measuring the speed of cryptographic operations. In addition, they may
```
kaduk: (Just noting on the PR that Ilari sent mail noting that some of these should move to the previous paragraph)

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+described in {{anti-replay}} are intended to
+prevent replay at the TLS layer do not provide complete protection
+against receiving multiple copies of client data.
+TLS 1.3 falls back to the 1-RTT
+handshake when the server does not have any information about the
+client, e.g., because it is in a different cluster which does not
+share state or because the ticket has been deleted as described in
+{{single-use-tickets}}. If the application layer protocol retransmits
+data in this setting, then it is possible for an attacker to induce a
+replay attack by sending the ClientHello to both the original cluster
```
kaduk: If we want to maintain a strict replay/retransmit separation, this text should change.

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+If implemented correctly, the mechanisms described in
+{{single-use-tickets}} and {{client-hello-recording}} prevent a
+replayed ClientHello and its associated 0-RTT data from being accepted
+multiple times by any cluster with consistent state; for servers
+which limit the use of 0-RTT to one cluster for a single ticket, then a given
+ClientHello and its associated 0-RTT data will only be accepted once.
+However, if state is not completely consistent,
+then an attacker might be able to have multiple copies of the data be
+accepted during the replication window.
+Because they do not know the exact details of server behavior, clients
```
kaduk: "they" is used before the referent ("client") here, which is perhaps poor writing style.

draft-ietf-tls-tls13.md
```diff
@@ -5636,6 +5759,72 @@ application protocols separately ensuring that confidential
...
+replayed and which they would not be willing to retry across multiple
+1-RTT connections.
+
+Application protocols MUST NOT use 0-RTT data without a profile that
+defines its use. That profile needs to identify which messages or
+interactions are safe to use with 0-RTT and how to handle the
+situation when the server rejects 0-RTT and falls back to 1-RTT.
+
+In addition, to avoid accidental misuse, TLS implementations MUST NOT
+enable 0-RTT (either sending or receiving) unless specifically
```
kaduk: I guess technically it is "accepting" (the negotiation of) rather than "receiving" that is configured, but that's a pretty fine nit to be picking.

draft-ietf-tls-tls13.md
kaduk: So ... MUST NOT?

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,193 @@ a context in all exporter computations, though the value could be empty.
...
+The first class of attack can be prevented by the mechanism described
+in this section.  Servers need not permit 0-RTT at all, but those
+which do SHOULD implement either the single-use tickets or
+ClientHello recording techniques described in the following two
+sections.
+
+The second class of attack cannot be prevented at the TLS layer and
+MUST be dealt with by any application. Note that any application whose
+clients implement any kind of retry behavior already needed to
+implement some sort of anti-replay defense.
```
ekr: Done.

draft-ietf-tls-tls13.md
```diff
@@ -4750,8 +4752,9 @@ weaker anti-replay defense because of the difficulty of reliably
 storing and retrieving the received ClientHello messages.
 In many such systems, it is impractical to have globally
 consistent storage of all the received ClientHellos.
-In this case, the stronger design is to have a single storage zone be
-authoritative for a given ticket and refuse 0-RTT for that
+In this case, the best anti-replay protection is provided by
+having single storage zone be
```
kaduk: "a single"

draft-ietf-tls-tls13.md
```diff
@@ -5788,8 +5800,8 @@ handshake when the server does not have any information about the
 client, e.g., because it is in a different cluster which does not
 share state or because the ticket has been deleted as described in
 {{single-use-tickets}}. If the application layer protocol retransmits
-data in this setting, then it is possible for an attacker to induce a
-replay attack by sending the ClientHello to both the original cluster
+data in this setting, then it is possible for an attacker to induce
+message duplicatio nby sending the ClientHello to both the original cluster
```
kaduk: s/o nb/on b/

draft-ietf-tls-tls13.md
```diff
@@ -5817,14 +5829,32 @@ interactions are safe to use with 0-RTT and how to handle the
...
+enable 0-RTT (either sending or accepting) unless specifically
+requested by the application and MUST NOT automatically resend 0-RTT
 data if it is rejected by the server unless instructed by the
-application.  In order to allow applications to implement special
-processing for 0-RTT data, when accepting 0-RTT as a server, a TLS
-implementation MUST provide a way for the application to determine if
+application. Server-side applications may wish to implement special
+processing for 0-RTT data for some kinds of application traffic (e.g.,
+abort the connection, request that data be resent at the application
+layer, or delay processing until the handshake completes) In order to
```
kaduk: full stop

draft-ietf-tls-tls13.md
chris-wood: "duplication by"

ekr: yeah, I caught this on too :)

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,193 @@ a context in all exporter computations, though the value could be empty.
...
+high rates of resumption and 0-RTT, at the cost of potentially
+weaker anti-replay defense because of the difficulty of reliably
+storing and retrieving the received ClientHello messages.
+In many such systems, it is impractical to have globally
+consistent storage of all the received ClientHellos.
+In this case, the stronger design is to have a single storage zone be
+authoritative for a given ticket and refuse 0-RTT for that
+ticket in any other zone. This approach prevents simple
+replay by the attacker because only one zone will accept
+0-RTT data. A weaker design is to implement separate storage for
```
kaduk: Er, I don't remember any (recent) comments from Victor Vasiliev relevant to having only one zone accept tickets for 0-RTT vs letting all zones do so, either on this pull request or on the WG list.  Did I miss something?

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,196 @@ a context in all exporter computations, though the value could be empty.
...
+In order to implement this, when a ClientHello is received, the server
+first verifies the PSK binder as described
+{{pre-shared-key-extension}}. It then computes the
+expected_arrival_time as described in the next section and rejects
+0-RTT if it is outside the recording window, falling back to the
+1-RTT handshake.
+
+If the expected arrival time is in the window, then the server
+checks to see if it has recorded a matching ClientHello. It
+either aborts the handshake with an "illegal_parameter" alert
```
kaduk: I think we want to say "If a matching ClientHello is found" (or similar) at the start of this sentence, for clarity.

draft-ietf-tls-tls13.md
ekr: He said none of this should be mandatory.

In any case, I think describing multiple options is within editor discretion here.

kaduk: Oh, describing multiple options is totally within editor discretion.
The main thing I wanted to comment about in this space was that starting a sentence "In this case, the stronger design" is a strange construction (in that it uses the definite article before it's clear what's being referred to), which it looks like you changed already.  (It's fine to say "a weaker design" without a preceeding "stronger design".)
So, I am belatedly realizing that you are only responding to my "mixed feelings" remark, which was not intended to be anywhere close to blocking, so I'll just shut up now; sorry for the noise.

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,196 @@ a context in all exporter computations, though the value could be empty.
...
+on the order of ten seconds to account for errors in clocks and
+variations in measurements; other deployment scenarios
+may have different needs. Clock skew distributions are not
+symmetric, so the optimal tradeoff may involve an asymmetric range
+of permissible mismatch values.
+
+Note that freshness checking alone is not sufficient to prevent replays
+because it does not detect them during the error window, which,
+depending on bandwidth and system capacity could include
+billions of replays in real-world settings.
```
knekritz: I would also add something along the lines of:
"Also note that this freshness checking is only done at the time the ClientHello is received, and not when later early application data records are received. After early data is accepted, records may continue to be streamed to the server over a longer time period."
I believe there was some previous discussion on the list about using the time window to determine when it is safe to resend rejected 0-RTT data, which is not actually sufficient by itself.

draft-ietf-tls-tls13.md
```diff
@@ -4696,6 +4641,196 @@ a context in all exporter computations, though the value could be empty.
...
+PSK identities, then an attacker can create multiple ClientHellos
+with different binder values for the less-preferred identity on the
+assumption that the server will not verify it, as recommended
+by {{pre-shared-key-extension}}.
+I.e., if the
+client sends PSKs A and B but the server prefers A, then the
+attacker can change the binder for B without affecting the binder
+for A. If the validated binder or the ClientHello.random
+are used as the storage key, then this attack is not possible.
+
```
ekr: At minimum it allows pollution of the table.

davegarrett:
@ekr: This was closed, but nothing appears to have been merged.

ekr:
I haven't pushed yet.

draft-ietf-tls-tls13.md
ekr: I agree.

draft-ietf-tls-tls13.md
ekr: Cleaned up a bit.

PR #1036, opened 2017-07-03
Annotate state machine with key changes, original due to Richard Barnes

ekr:


ekr:
@martinthomson @davegarrett I would appreciate a sanity check here.

draft-ietf-tls-tls13.md
```diff
@@ -614,6 +614,11 @@ draft-21
 - Add a per-ticket nonce so that each ticket is associated with a
   different PSK (*).
 
+- Clarify that clients should change to sending with handshake keys
+  immediately upon reading ServerHello. Arguably a wire format change. (*).
```
martinthomson: It either is, or it isn't.  I think that it isn't, and this is merely a clarification.

Also, this formulation doesn't work for 0-RTT at all.  The fact is that the client's second flight uses handshake keys, except for EOED.

draft-ietf-tls-tls13.md
```diff
@@ -4995,22 +5000,24 @@ by IANA:
...
             Send ClientHello |        | Recv HelloRetryRequest
+       [K_send = early data] |        |
          /                   v        |
-        |                  WAIT_SH ---+
-    Can |                    | Recv ServerHello
-   send |                    V
-  early |                 WAIT_EE
-   data |                    | Recv EncryptedExtensions
-        |           +--------+--------+
+        |                 WAIT_SH ----+
```
martinthomson: Can you drop the "/" down one line?  It now looks like the "[K_send = early data]" is connected to the line in some way, which doesn't make a lot of sense if you are thinking, but...

draft-ietf-tls-tls13.md
```diff
@@ -4995,22 +5000,24 @@ by IANA:
...
+        |                    | Recv ServerHello
+        |                    | K_send = K_recv = handshake
```
martinthomson: Here my previous point holds.  You set K_send after receiving the Finished, not here.

draft-ietf-tls-tls13.md
```diff
@@ -5023,14 +5030,13 @@ in \[].
         |           |                 | Recv CertificateVerify
         |           +> WAIT_FINISHED <+
         |                  | Recv Finished
-        \                  |
+         \                 | K_recv = application
                            | [Send EndOfEarlyData]
                            | [Send Certificate [+ CertificateVerify]]
```
martinthomson: As above, this goes:

* Recv Finished
* [K_send = early data]
* [Send EndOfEarlyData]
* K_send = handshake
* [Send Certificate [+ CertificateVerify]]
* Send Finished
* K_send = application
* K_recv = application (sure, whatever)

draft-ietf-tls-tls13.md
```diff
@@ -5045,17 +5051,20 @@ in \[].
...
                                | Send EncryptedExtensions
                                | [Send CertificateRequest]
 Can send                       | [Send Certificate + CertificateVerify]
 app data -->                   | Send Finished
-after                 +--------+--------+
-here         No 0-RTT |                 | 0-RTT
-                      |                 v
-                      |             WAIT_EOED <---+
+after                          | K_send = application
+here                  +--------+--------+
```
martinthomson: Now that you have K_send = application, you could just say that after installing application keys in K_send, then you can send application data.  Same goes for receiving it.

At a minimum, I would move the arrow to the line where you install the keys.

draft-ietf-tls-tls13.md
```diff
@@ -5045,17 +5051,20 @@ in \[].
...
+             No 0-RTT |                 | 0-RTT
+   K_recv = handshake |                 | K_recv = early_data
+[Skip decrypt errors] |             WAIT_EOED <---+
                       |            Recv |   |     | Recv
-                      |  EndOfEarlyData |   |     | early data
-                      |                 |   +-----+
+                      |  EndOfEarlyData |   |     | Early data
+                      |        K_recv = |   +-----+
+                      |       handshake |
```
martinthomson: I don't know how to improve this, but this is a little cramped.  Maybe...
```
  K_recv = handshake  |                 |                           
[Skip decrypt errors] |    +------> WAIT_EOED --+                    
                      |    |       Recv |       | Recv EndOfEarlyData
                      |    | early data |       | K_recv = handshake
                      |    +------------+       |
                      |                         |
                      +> WAIT_FLIGHT2 <---------+
```
That's 69 columns exactly, though you can trim one extra.

draft-ietf-tls-tls13.md
```diff
@@ -5070,7 +5079,7 @@ here         No 0-RTT |                 | 0-RTT
                       |             v       | CertificateVerify
                       +-> WAIT_FINISHED <---+
                                | Recv Finished
-                               v
+                               v K_recv = application
```
martinthomson: I'd add another line here.

draft-ietf-tls-tls13.md
ekr: Done

PR #1039, opened 2017-07-04
Unclutter the server state machine

martinthomson:
This isn't necessarily any better, but it's an option.  I didn't want to lose this from my edit buffer, so you can have it.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5064,14 +5064,14 @@ app data                       | Send Finished
 after   -->                    | K_send = application
 here                  +--------+--------+
              No 0-RTT |                 | 0-RTT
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -5064,14 +5064,14 @@ app data                       | Send Finished
 after   -->                    | K_send = application
 here                  +--------+--------+
              No 0-RTT |                 | 0-RTT
-   K_recv = handshake |                 | K_recv = early_data
```
beurdouche: @martinthomson Was the `K_recv = early_data` removed by mistake ?

martinthomson: Yes.

beurdouche: Personally, I find this style much better BTW...

PR #1049, opened 2017-07-11
Document the static key share attack

huitema:
As discussed on the mailing list, here is a short proposal to document the "static (EC) DH private key" sharing attack.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3,8 +3,8 @@ title: The Transport Layer Security (TLS) Protocol Version 1.3
 abbrev: TLS
 docname: draft-ietf-tls-tls13-latest
 category: std
-updates: 4492
-obsoletes: 5077, 5246, 5746
+updates: 4492, 5705, 6066, 6961
+obsoletes: 5077, 5246
 
 ipr: pre5378Trust200902
 area: General
@@ -33,28 +33,16 @@ author:
 normative:
   RFC2104:
   RFC2119:
-  RFC2434:
-  RFC3447:
+  RFC5226:
   RFC5280:
-  RFC5288:
-  RFC5289:
   RFC5869:
   RFC6066:
-  RFC6209:
-  RFC6367:
   RFC6655:
-  RFC7251:
-  I-D.ietf-tls-chacha20-poly1305:
-  I-D.irtf-cfrg-curves:
-  I-D.irtf-cfrg-eddsa:
-
-  AES:
-       title: Specification for the Advanced Encryption Standard (AES)
-       date: 2001-11-26
-       author:
-         org: National Institute of Standards and Technology
-       seriesinfo:
-         NIST: FIPS 197
+  RFC7539:
+  RFC7748:
+  RFC7919:
+  RFC8032:
+  RFC8017:
   SHS:
        title: Secure Hash Standard
        date: 2012-03
@@ -83,44 +71,34 @@ normative:
           - ins: M. Hellman
         date: 1977-06
         seriesinfo: IEEE Transactions on Information Theory, V.IT-22 n.6
-
+  GCM:
+        title: "Recommendation for Block Cipher Modes of Operation: Galois/Counter Mode (GCM) and GMAC"
+        date: 2007-11
+        author:
+          - ins: M. Dworkin
+        seriesinfo:
+          NIST: Special Publication 800-38D
 
 informative:
-  RFC0793:
-  RFC1948:
   RFC4086:
-  RFC4279:
-  RFC4302:
-  RFC4303:
   RFC4346:
   RFC4366:
   RFC4492:
-  RFC4506:
-  RFC4507:
   RFC4681:
-  RFC5054:
   RFC5077:
-  RFC5081:
   RFC5116:
   RFC5246:
-  RFC5746:
-  RFC5763:
   RFC5764:
-  RFC5878:
   RFC5929:
   RFC6176:
   RFC6091:
   RFC6520:
-  RFC6961:
-  RFC6962:
-  RFC7301:
+  RFC7230:
   RFC7250:
-  RFC7366:
   RFC7465:
   RFC7568:
   RFC7627:
   RFC7685:
-  I-D.ietf-tls-negotiated-ff-dhe:
 
   DSS:
        title: "Digital Signature Standard, version 4"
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1,4 +1,4 @@
----
+﻿---
```
martinthomson: This change is why travis is complaining.  Removing the hypen would be fine, but adding a byte order mark isn't.

martinthomson:
That pernicious BOM remains.  I assume that if this PR is OK, that's an easy editorial fix, so I wouldn't stress about it.

huitema:
I am really stumped. I just tried integrating the recent changes in the master copy, in case there would be some kind of incompatibility.

huitema:
Finally fixed. I have no idea where the extra dash came from.

davegarrett:
There's a bunch of merge commits that shouldn't be in here in addition to the churn from fixing the build failure (which was likely a result of the former, somehow). Ideally, you should have concise commits of just what you're changing in the PR. Can be fixed via git commands, easily enough:

Please do a git squash of these changesets down to a single commit, then rebase it on top of the current upstream master. Do a diff to upstream master to make sure it matches the cumulative diff shown here, clean up the commit message if needed, then force push to this PR branch to update it. For this PR, exactly one commit is sufficient.

huitema:
@davegarrett : I am of course willing to do that, but it appears to be a notch above my level of expertise in git. Do you have a pointer to a "how to" doc? Or, I can give you access to my fork if that helps.

davegarrett:
@huitema: First, taking a longer look at this PR technically, it seems you're pushing this PR from your master branch. That's why we have merge commits and stuff listed from 2015. PRs should be submitted from separate branches made for each PR. (this'll probably still work, so don't start over; just for future reference, to avoid similar pitfalls) You can create new branches via the command line or just use the GitHub web interface to edit, both of which can be synced and then edited with the other. It's much easier to avoid/fix test fails if you've got a clean PR branch with just your new work.

Here's a quick link for rebasing with squash:
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

An simpler way to do this, because this changeset is just a paragraph, is to just save the block of new text separately, do a git reset on your PR branch (your fork's master in this case here) to make it match upstream master exactly, then make a new clean commit with the new text. When you force push your local version to GitHub, it will overwrite this PR (but leave most of the comments, generally).

Resetting a branch to match upstream exactly is just "git reset --hard upstream/master" whilst on your branch (assuming you've mapped upstream as such, and have fetched latest). Of course, make sure you've already backed-up your change to re-commit and always check a diff to upstream master before force pushing.

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
 the reuse of those keys. This parallels the use of the early
 application traffic keys only in the client-to-server direction.
 
+## Static (EC) DH Shares
+
+For various reasons, some implementations may be tempted to use static
+(EC) DH private key. Using such keys lowers the security guarantees of
```
davegarrett: nit: do "(EC)DH" without the space, to match usage below

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
...
+TLS 1.3. Adversaries that get access to the static (EC) DH private key
+can now get access to the content of the communication. Adversaries that
```
davegarrett: nit: drop "now"

brainhub:
If one replaces "**static (EC)DH private keys**" with **PSK** (static or for session resumption), the consequences of this _attack_ are about the same.  It's a reasonable assumptions that the compromise of a cached ECDH share should imply compromise of all sessions that the server can resume. The corresponding state is expected to live in the same RAM of the server. 

This compromise is mitigated by managing the lifetime of PSKs or ECDH shares. To put it differently, if a server operator takes this text as a serious concern, the server operator must disable session resumption and any use of PSK, unless the ECDH with PSK is the only option.

Should this text at least mention a similar _attack_ on resumption and PSK?

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
 the reuse of those keys. This parallels the use of the early
 application traffic keys only in the client-to-server direction.
 
+## Static (EC)DH Shares
+
+For various reasons, some implementations may be tempted to use static
```
kaduk: Implementations probably should not be assumed to have motive, whereas implementors would.

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
...
+(EC)DH private keys. Using such keys lowers the security guarantees of
+TLS 1.3. Adversaries that get access to the static (EC)DH private key
```
kaduk: Lowering a guarantee feels like an odd phrasing to me, as a guarantee is supposed to be a fixed thing.  So, one might reduce the level of security that is guaranteed, or break a guarantee, or similar, instead.

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
...
+can get access to the content of the communication. Adversaries that
+acquire the key in real-time can compromise the confidentiality of the
+conversation. Adversaries that acquire the key later can use it to
+access the content of recorded sessions, thus breaking the forward
```
kaduk: This text doesn't really do much to distinguish post-handhsake compromise from contemporary compromise, in that they can "compromise confidentiality" or "access the content of", which is basically synonymous.  If we are to cover these cases separately, it would be good to be more clear about what is different between them.

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
...
+secrecy promise of the protocol. TLS 1.3 clients should detect the use
```
kaduk: I don't know that I'm very happy about "should" (vs. "SHOULD") for this.

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
...
+of static (EC)DH keys by corresponding servers, and should treat servers
+using such keys as compromised. Clients can perform this detection by
```
kaduk: "treat as compromised" is somewhat vague; it might be better to also add that clients should assume that confidentiality is not protected on connections to such servers, etc.

draft-ietf-tls-tls13.md
```diff
@@ -5924,6 +5924,21 @@ server-to-client encryption keys because that would entail
...
+comparing keys proposed by the same server at different times, or by
+cooperating with other clients to compare the keys proposed to them by
+sets of servers.
```
kaduk: This sort of cooperating with other clients seems to suffer from the same client fingerprinting issue that has plagued attempts to introduce gossiping in CT.

ekr:
I'm not detecting a lot of consensus for this.

PR #1051, opened 2017-07-13
Send server plaintext records with a different content type.

knekritz:
We've (Facebook) been experimenting with some tweaks to the record layer to try to work around middlebox interference issues when negotiating TLS 1.3 on the internet. Initial results using a different content type for server plaintext handshake messages show significant improvement, in particular against the two most common issues we see (on mobile clients):
1) access_denied alert received on the client
2) missing server hello (but the rest of the connection continues)
We're still collecting data, but assuming this continues to show deployment improvement I think this change makes a lot of sense with minimal complication added to the protocol.

Note that I picked the value at random and that is what we've been experimenting with, but I assume any other value would work too.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1640,6 +1640,10 @@ of a connection. Handshake messages are supplied to the TLS record layer, where
 they are encapsulated within one or more TLSPlaintext or TLSCiphertext structures, which are
 processed and transmitted as specified by the current active connection state.
 
+Implementations MUST send these messages with the "handshake" ContentType, with the exception of
+server plaintext messages (ServerHello and HelloRetryRequest) which MUST be sent using the
+"server_plaintext_handshake" ContentType.
+
 %%% Handshake Protocol
 
        enum {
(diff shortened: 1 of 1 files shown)


martinthomson:
@knekritz, is this something that works more generally?  What if we changed to a completely new content type for all application data/handshake records?  (and could be also remove the extra 3 octets from the header when we did that?)

ilaril:
Did you also try modifying ServerHello to look more like TLS 1.2 ServerHello, and seeing what that does to intolerance rates? Specifically, adding one zero byte after ServerRandom (where TLS 1.2 would have session ID) and one zero byte after Ciphersuite (where TLS 1.2 would have compression)?

Also, I don't suppose totally alien framing starting directly after ServerHello would do good things to intolerance rates... What are these middleboxes trying to do? Why would they scan ServerHello, but then accept handshake seemingly cutting short?

Also what kind of connections were tested? I would guess various mobile and WiFi connections (as those tend to be the ones mobile devices connect to). Also, is there difference between IPv4 and IPv6 when both are available from the same source?

ekr:
I think we concluded this didn't help

PR #1056, opened 2017-07-16
Resumption/PSK diagram: add PSK KE extension to initial handshake

nharper:


kaduk:
Can you say a bit more about why you want this?  If the client is not offering any PSKs, would it want to offer psk_key_exchange_modes?

nharper:
This is clarifying the diagram - it does not change the behavior specified.

The psk_key_exchange_modes extension is used both to indicate what modes the PSK on that handshake can be used for, and what modes a PSK supplied by an NST can be used for.

From section 4.2.9: "The semantics of this extension are that the client only supports the use of PSKs with these modes, which restricts both the use of PSKs offered in this ClientHello and those which the server might supply via NewSessionTicket."

Also (from the same section): "This extension also restricts the modes for use with PSK resumption; servers SHOULD NOT send NewSessionTicket with tickets that are not compatible with the advertised modes; however, if a server does so, the impact will just be that the client’s attempts at resumption fail."

kaduk:
Ah, right.  Thanks!

davidben:
Another option would be to omit it from both diagrams, to reduce clutter, since it's now like supported_groups where it is a capability advertisement relevant to the handshake, but not tied to any particular shape.

Issue #1043, opened 2017-07-05
Early data with external PSK


mattcaswell:
The Early Data Indication section contains this sentence:

    The parameters for the 0-RTT data (symmetric cipher suite, ALPN
    protocol, etc.) are the same as those which were negotiated in the
    connection which established the PSK.

This implies to me that you can only do 0-RTT where the PSK was established via an earlier connection. In external PSK this is not the case. I don't think this interpretation was the intention.

kaduk:
This is stale text from before we allowed 0-RTT with external PSK, IIRC.

Issue #1040, opened 2017-07-04
resumption validation


kazu-yamamoto:
2.3.  Zero-RTT Data says:

> -  The TLS version number for use with this PSK
> 
> -  The cipher suite for use with this PSK
> 
> -  The Application-Layer Protocol Negotiation (ALPN) protocol [RFC7301], if any is to be used
> 
> -  The Server Name Indication (SNI), if any is to be used

But 4.6.1.  New Session Ticket Message says:

> Any ticket MUST only be resumed with a cipher suite that has the same
> KDF hash algorithm as that used to establish the original connection,
> and only if the client provides the same SNI value as in the original
> connection, as described in Section 3 of [RFC6066].

For consistency, I would like to have the latter in Section 2.2 with the same style.

Moreover, I would like to know where this differences of requirements come from.

kaduk:
I think 4.2.10 may be more appropriate than 2.2, as it is closer to where the server is doing the relevant processing.

kazu-yamamoto:
Now 4.2.10 has clear requirements. So, why don't we remove the same requirement from 2.3?

> When clients use a PSK obtained externally to send early data, then the following
additional information MUST be provisioned to both parties:
>
> -  The TLS version number for use with this PSK
> 
> -  The cipher suite for use with this PSK
> 
> -  The Application-Layer Protocol Negotiation (ALPN) protocol [RFC7301], if any is to be used
> 
> -  The Server Name Indication (SNI), if any is to be used

PR #1061, opened 2017-07-21
Be more explicit about PSK requirements

kaduk:
Both for 0-RTT and 1-RTT.

Give NST- and externally-provisioned PSKs more uniform treatment.

Closes: #1040
Closes: #1043

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2823,18 +2823,18 @@ In order to accept early data, the server MUST have accepted a
 PSK cipher suite and selected the first key offered in the
 client's "pre_shared_key" extension. In addition, it MUST verify that
 the following values are consistent with those associated with the selected
-PSK.  For externally established PSKs, the associated values are those
-provisioned along with the key.  For PSKs established via a NewSessionTicket
-message, the associated values are those negotiated in the
-connection during which the ticket was established.
+PSK:
 
 - The TLS version number
-- The selected cipher suite.
-- The selected ALPN {{RFC7301}} protocol, if any.
-- The selected SNI (Section 3 of {{RFC6066}}) value, if any.
+- The selected cipher suite
+- The selected ALPN {{RFC7301}} protocol, if any
+- The selected SNI (Section 3 of {{RFC6066}}) value, if any
 
 These requirements are a superset of those needed to perform a 1-RTT
-handshake using the PSK in question.
+handshake using the PSK in question.  For externally established PSKs, the
+associated values are those provisioned along with the key.  For PSKs
+established via a NewSessionTicket message, the associated values are those
+negotiated in the connection during which the ticket was established.
 
 Future extensions MUST define their interaction with 0-RTT.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2819,11 +2822,19 @@ MUST behave in one of three ways:
...
+PSK.  For externally established PSKs, the associated values are those
+provisioned along with the key.  For PSKs established via a NewSessionTicket
+message, the associated values are those negotiated in the
 connection during which the ticket was established.
 
-- The TLS version number and cipher suite.
+- The TLS version number
+- The selected cipher suite.
 - The selected ALPN {{RFC7301}} protocol, if any.
+- The selected SNI (Section 3 of {{RFC6066}}) value, if any.
```
martinthomson: list construction: remove the period

draft-ietf-tls-tls13.md
```diff
@@ -3607,10 +3627,13 @@ authentication state. Clients SHOULD attempt to use each
...
 
-Any ticket MUST only be resumed with a cipher suite that has the
-same KDF hash algorithm as that used to establish the original connection,
-and only if the client provides the same SNI value as in the original
-connection, as described in Section 3 of {{RFC6066}}.
+Any ticket MUST only be resumed on connections where the following values
+match the values negotiated in the connection during which the ticket
+was established:
+
+
```
martinthomson: extra line

draft-ietf-tls-tls13.md
martinthomson: I would move the text on external vs. ticket after the list so that you can use a colon.  e.g.,

```
In addition, it MUST verify that
the following values are consistent with those associated with the selected
PSK:

- ...

These requirements are a superset of those needed to perform a 1-RTT
handshake using the PSK in question.  For externally established PSKs,
the associated values are those provisioned along with the key.  For
PSKs established via a NewSessionTicket message, the associated
values are those negotiated in the connection during which the ticket
was established.
```

kaduk:
@ekr ping?  It would be good to get some resolution in this space, as it's somewhat relevant to https://github.com/openssl/openssl/pull/3926

kaduk:
I guess we could perhaps recommend that external PSKs not be provisioned with a SNI value, since the PSK value itself ought to serve as enough of a lookup key for what the intended communication semantics are.

kaduk:
@ekr is there a path forward for this (e.g., does it need to go to the list)?

ekr:
@kaduk: I think this is generally good except for the SNI stuff which conflicts with PR #1080, which we also need to resolve one way or another. What say we land this without those pieces and then when we resolve #1080, we will know what to do

kaduk:
>  What say we land this without those pieces and then when we resolve #1080, we will know what to do

SGTM.  Do you need me to adjust the patch?

ekr:
Nah, I'll do it.

ekr:
These actually just seem to make the SNI requirements clearer but no sharper, so merging.

PR #1063, opened 2017-07-26
Use the enum value.

stevecheckoway:


davegarrett:
Why?

stevecheckoway:
Because the description language says you can assign enumerated elements to enumerated types. It does not say you can assign integers.

From Section 3.5.

```
      Color color = Color.blue;     /* overspecified, legal */
      Color color = blue;           /* correct, type implicit */
```

ekr:
@stevecheckoway : this doesn't seem to solve the problem because the next line is an integer assignment and we don't want to assign a code point for that, so I think we should probably adjust the language.

stevecheckoway:
Adjusting the language is certainly doable, but I don't think `ProtocolVersion` is an issue since it's not an enumerated type, but a `uint16`.

PR #1066, opened 2017-07-26
Give each select arm an optional field label.

stevecheckoway:
Each select arm consists of a single type with an optional field label
(for non-vector types).

    select (E) {
        case e1: Te1 [[fe1]];
        case e2: Te2 [[fe2];
        ....
        case en: Ten [[fen]];
    };

PreSharedKeyExtension had an arm with two fields, a new structure,
OfferedPsks` was created to hold it. This is purely a notational change.


ekr:
Merged by hand

PR #1070, opened 2017-07-28
Use /* comments */ rather than //

stevecheckoway:
The presentation language only specifies `/* ... */` as a comment. It
doesn't really apply to the `Transcript-Hash`, but it seems good to be
consistent.


richsalz:
I'd prefer to go the other way; describe and only use // comments.

stevecheckoway:
I have no problem with that. I don't think multiline comments are used anywhere. This was the minimal change.

ekr:
I'd prefer to stay consistent with the previous syntax.

PR #1073, opened 2017-08-15
How to handle padding so that TLS implementation is constant time

tomato42:
add information on handling the padding in constant time


ekr:
I think it would be fine to sharpen the point about padding leaking information and I'd take a short PR for that.  I don't believe it's necessary either to require that it be constant time (for the reasons I indicated on-list and already documented in the spec) or to describe a specific algorithm, especially at this point on the document life cycle.

tomato42:
obsoleted by #1074

PR #1075, opened 2017-08-16
Make the deprecation of status_request_v2 extension more explicit

tomato42:


draft-ietf-tls-tls13.md
```diff
@@ -3304,6 +3304,13 @@ The body of the "status_request" extension
 from the server MUST be a CertificateStatus structure as defined
 in {{RFC6066}}, which is interpreted as defined in {{!RFC6960}}.
 
+Note: status_request_v2 extension ({{!RFC6961}}) is deprecated. TLS 1.3 servers
+MUST NOT act upon its presence or information in it when processing Client
+Hello, in particular they MUST NOT send status_request_v2 extension in the
```
kaduk: I think a "the" is needed as "the status_request_v2 extension".

tomato42: that should be fixed now

PR #1076, opened 2017-08-29
Updating second ClientHello based on HelloRetryRequest

dcooper16:
Section 4.1.2 lists the specific items that may be changed when sending a second ClientHello in response to a HelloRetryRequest. Since the HelloRetryRequest specifies a server_version and cipher_suite, then it seems that there is no reason for the second ClientHello to offer any version or cipher suite other than the ones in the HelloRetryRequest.


ekr:
We explicitly decided very early that we weren't going to do this, and in any case, I think it's way too late to change it now. Feel free to try to get consensus on the list, but for now I'm closing this.

PR #1080, opened 2017-09-13
Allow resumption across multiple domains

vasilvv:
There is no reason from cryptographical point of view to not resume
across multiple domains.  The primary concern comes from
interoperability concerns and potential waste of tickets, hence the
default should be to not resume.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1353,7 +1353,6 @@ additional information MUST be provisioned to both parties:
   * The cipher suite for use with this PSK
   * The Application-Layer Protocol Negotiation (ALPN) protocol {{!RFC7301}}, if
     any is to be used
-  * The Server Name Indication (SNI), if any is to be used
 
 As shown in {{tls-0-rtt}}, the 0-RTT data is just added to the 1-RTT
 handshake in the first flight. The rest of the handshake uses the same messages
@@ -2840,7 +2839,6 @@ PSK:
 - The TLS version number
 - The selected cipher suite
 - The selected ALPN {{RFC7301}} protocol, if any
-- The selected SNI (Section 3 of {{RFC6066}}) value, if any
 
 These requirements are a superset of those needed to perform a 1-RTT
 handshake using the PSK in question.  For externally established PSKs, the
@@ -2949,14 +2947,15 @@ PSK is established, or default to SHA-256 if no such algorithm
 is defined. The server MUST ensure that it selects a compatible
 PSK (if any) and cipher suite.
 
-Each PSK is also associated with at most one Server Name Identification
-(SNI) value (Section 3 of {{RFC6066}}).  For PSKs established via the
-ticket mechanism ({{NSTMessage}}), this is the SNI value (if any) on
-the connection where the ticket was established.  For externally established
-PSKs, the associated SNI value (or lack of value) is set when the PSK
-is established.  The server MUST ensure that it selects a compatible
-PSK (if any) and SNI value (if any).  SNI MUST NOT be negotiated when
-using a PSK not associated with an SNI value.
+In TLS versions prior to TLS 1.3, the Server Name Identification (SNI) value was
+intended to be associated with the session (Section 3 of {{RFC6066}}), with the
+server being required to enforce that the SNI value associated with the session
+matches the one specified in the resumption handshake.  However, in reality the
+implementations were not consistent on which of two supplied SNI values they
+would use, leading to the consistency requirement being de-facto enforced by the
+clients.  In TLS 1.3, the SNI value is always explicitly specified in the
+resumption handshake, and there is no need to associate an SNI value with the
+ticket anymore.
 
 Implementor's note: the most straightforward way to implement the
 PSK/cipher suite matching requirements is to negotiate the cipher
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3616,9 +3616,16 @@ ticket no more than once, with more recent tickets being used
 first.
 
 Any ticket MUST only be resumed with a cipher suite that has the
-same KDF hash algorithm as that used to establish the original connection,
-and only if the client provides the same SNI value as in the original
-connection, as described in Section 3 of {{RFC6066}}.
+same KDF hash algorithm as that used to establish the original connection.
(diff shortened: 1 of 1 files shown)


ekr:
@vasilvv This looks plausible. Can you please email the list?

kaduk:
It's perhaps more text than is needed to convey the point, but I do not object to the change itself.
It may be worth explicitly calling out the divergence from RFC 6066, though.
(Also, I note that we require external PSKs to be provisioned with SNI for use with 0-RTT, though there does not seem to be a corresponding change needed to that text.)

ekr:
I'm fine with this as long as it does not create new security problems. My sense is that this is no worse than the current situation, because the client only will do this if the server has shared certs, which is one of the cases where caution is already required based on http://antoine.delignat-lavaud.fr/doc/www15.pdf

@adl, @prosecco, could you weigh in?

ad-l:
@ekr Enabling cross-domain resumption unconditionally is still a dodgy idea for the same reasons that it was back in 2015. The main issue is that it introduces a discrepancy between the application-level security policy (which is still mostly origin based for HTTP) and the transport-level guarantees when applied to the full certificate rather than only the SNI of the connection.

To summarize the WWW'15 paper, the key observation is that it is actually very common in practice for origins with different application-level security policies to appear in the same certificate. The top 2 reasons why this happens is 1. wildcard/SAN certificates 2. CDN and other name-based cloud hosting. The later is particularly bad, because it means an attacker may be able to get his website to appear in the same certificate as an honest website it tries to attack - such an attacker becomes incredibly powerful in particular if a multi-stream application protocol such as HTTP/2, or QUIC is used for stream multiplexing above encryption (I don't believe any of the published theorems on TLS or QUIC security provide any security against this attacker) - for instance, CRIME-like attacks may become possible again.

The main way to exploit such discrepancies that I know of relies on the fact that most HTTP requests fallback to the default virtual host if the "Host" header of the request does not match any configured VH. Since it is commonly the case that separate servers/virtual machines are used for the different names in a certificate (or subdomains with wildcard), redirecting requests from one server to another causes pages to be loaded under the wrong origin. This lets you "port" web vulnerabilities (or lack of strong security, e.g. CSP/HSTS) from one domain to the other. I have tried to convince webserver developers such as apache and nginx that this behavior is too dangerous but since then given up on trying to remove this fallback-to-default mechanism (it simply breaks too much of the Web). I don't know whether current implementations of HTTP/2 still have this behavior - if they do, I think it is too much of a risk to allow cross-domain resumption unless it is explicitly enabled. If they don't, I suggest that this feature is only made available if the negotiated ALPN is NOT http/1.1 or older.

In addition, it is worth noting that cross-domain, same-certificate sharing breaks many proposed or deployed transport-level security mechanisms, including HSTS (see the paper for attack example), but also Token Binding/Channel ID (for instance, this may allow a network+web attacker to force Token-bound cookies on a website for which it shares a certificate).

vasilvv:
@ad-l Don't most of the arguments against cross-domain resumption apply to HTTP2/QUIC connection pooling too?   I agree that pooling has some tricky edge cases, but it's been already deployed as a part of HTTP/2, so I think we're currently headed in the direction of working out those cases rather than backing away from the idea altogether.

Cross-domain resumption is just extension of cross-domain pooling for cases where the server IP does not match (and that should not matter from the cryptographic security perspective, since DNS is not cryptographically verified anyways).  The main source of confusion I am afraid of is when the server starts using SNI from the session cache and ends up using it instead of the new SNI -- which is a problem easily solved by not storing the SNI in it ( @ekr -- should I just ban that explicitly? )

I am somewhat confused by what you mean when you say those things break HSTS and tokbind.  The only attack against HSTS that I see in the paper is the one in section 2 -- and it depends on attacker's ability to send arbitrary headers from the victim domain's origin.  I am not sure how pooling or cross-resumption applies here.  Tokbind also should not be problematic -- as long as the entity with the certificate determines backends correctly and routes request to them, all of the requests can only arrive to the respective backends (also, I should note here that the TLS server in fact *has* to handle tokbind identification here, since the backends do not have access to the TLS exporter of the session).

Client certificates are somewhat tricky -- I can imagine some legitimate cases where you'd want to cross-resume with them, but overall they are more problematic than useful for both pooling and cross-resumption scenarios.  I think I'll just edit the PR and ban cross-resuming if client has presented a cert.

ekr:
@ad-l, any response to @vasilvv ?

grittygrease:
I personally think this is a good idea. It especially makes sense in the HTTP/2 case if the server has advertised explicit support for the new domain with an ORIGIN frame and the previous connection included the domain in its trust set.

davidben:
Per WG meeting, I think we should add the following text:

> On resumption, if reporting an SNI value to the calling application, implementations MUST the value sent in the resumption ClientHello rather than the value sent in the previous session. Note that if a server implementation declines all PSK identities with different SNI values, these two values are always the same.

This matters if the server believes SNI is a per-session property rather than per-connection and does not check equality with the resumption ClientHello value. I know of one popular TLS 1.2 implementation which does this, which is why we can't unilaterally do cross-resumptions in TLS 1.2. today. I.e. the client SHOULD needs to be a client MUST in TLS 1.2.

Otherwise, I think this is a good idea. There are clear benefits to cross-name resumption in some scenarios (same benefits as HTTP/2 cross-name pooling). But we should call this out so that implementations which got it wrong in TLS 1.2 don't continue to get it wrong in TLS 1.3 and mess this up.

ekr:
@vasilvv sorry to bother you but could you please rebase this on top of master. that we we know the merge is the way you want it.

vasilvv:
I've reconciled this with changes made in #1061, so that we're consistent on SNI being per-connection value.  I've added some extra descriptive text explicitly calling this out as a change.

cc @kaduk

kaduk:
(I am leaving for the airport now and may not get to look until I'm back in the US)

kaduk:
I think we probably want to put back some form of:

```
-Clients SHOULD attempt to use each ticket no more than once, with 
-more recent tickets being used first.
```

tomato42:
@kaduk this is already part of the "Client Tracking Prevention" section in "Implementation Notes"

you think it should be restated in the "New Session Ticket Message" section?

kaduk:
Maybe a reference to the "Client Tracking Prevention" section would suffice; I'll put together a PR.

PR #1083, opened 2017-10-04
Fix TLSInnerPlaintext limit

Lekensteyn:
The TLSPlaintext size is at most 2^14, so the TLSInnerPlaintext must be
at least one more to fit the content type. This matches with the end of
the record payload protection section.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2490,7 +2490,7 @@ extensions.
 
 The "post_handshake_auth" extension is used to indicate that a client is willing
 to perform post-handshake authentication {{post-handshake-authentication}}. Servers
-MUST not send a post-handshake CertificateRequest to clients which do not
+MUST NOT send a post-handshake CertificateRequest to clients which do not
 offer this extension. Servers MUST NOT send this extension.
 
 The "extension_data" field of the "post_handshake_auth" extension is zero
@@ -4073,7 +4073,7 @@ a non-zero octet in the cleartext, it MUST terminate the
 connection with an "unexpected_message" alert.
 
 The presence of padding does not change the overall record size limitations
-- the full encoded TLSInnerPlaintext MUST not exceed 2^14 + 1 octets. If the
+- the full encoded TLSInnerPlaintext MUST NOT exceed 2^14 + 1 octets. If the
 maximum fragment length is reduced, as for example by the max_fragment_length
 extension from [RFC6066], then the reduced limit applies to the full plaintext,
 including the content type and padding.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4073,10 +4073,10 @@ a non-zero octet in the cleartext, it MUST terminate the
 connection with an "unexpected_message" alert.
 
 The presence of padding does not change the overall record size limitations
-- the full encoded TLSInnerPlaintext MUST not exceed 2^14 octets. If the maximum
-fragment length is reduced, as for example by the max_fragment_length extension
-from [RFC6066], then the reduced limit applies to the full plaintext,
-including the padding.
+- the full encoded TLSInnerPlaintext MUST not exceed 2^14 + 1 octets. If the
+maximum fragment length is reduced, as for example by the max_fragment_length
+extension from [RFC6066], then the reduced limit applies to the full plaintext,
+including the content type and padding.
 
 Selecting a padding policy that suggests when and how much to pad is a
 complex topic and is beyond the scope of this specification. If the


davidben:
Nice catch! We also misread that and applied the check to the wrong end to begin with, but the +1 seems a good idea anyway. Just seems generally simpler if TLS 1.3 doesn't have a pesky version-specific -1 in figuring out how much data can fit in an outgoing record.

Lekensteyn:
Would be nice to have a new test in bogo that checks for rejection of TLSPlaintext having a size of 2^14+2 (or 2^14+1 in case of TLS 1.2).

davidben:
Yup, will certainly add a test when we fix it. (Though I'll wait for this PR to either go in or be rejected just to have a ruling on whether we should have a +1 in there.)

draft-ietf-tls-tls13.md
```diff
@@ -4073,10 +4073,10 @@ a non-zero octet in the cleartext, it MUST terminate the
...
 
 The presence of padding does not change the overall record size limitations
-- the full encoded TLSInnerPlaintext MUST not exceed 2^14 octets. If the maximum
-fragment length is reduced, as for example by the max_fragment_length extension
-from [RFC6066], then the reduced limit applies to the full plaintext,
-including the padding.
+- the full encoded TLSInnerPlaintext MUST not exceed 2^14 + 1 octets. If the
+maximum fragment length is reduced, as for example by the max_fragment_length
+extension from [RFC6066], then the reduced limit applies to the full plaintext,
+including the content type and padding.
```
davidben: Should max_fragment_length also exclude the content type? It seems sort of weird that, if you previously believed max_send_fragment was 2^14, you don't need version-specific logic, but if you believed it was 2^9, 2^10, 2^11, or 2^12 by way of the extension, you need version-specific logic again.

Lekensteyn: Given the use case (constrained devices with limited memory) I think that it should include content type and padding. The limit seems to exist to reduce the memory requirement for the AEAD plaintext rather than the application data fragment.

The proposed text also matches with @martinthomson's https://tools.ietf.org/html/draft-ietf-tls-record-limit-01#section-4
>   This value is the length of the plaintext of a protected record.  The
>   value includes the content type and padding added in TLS 1.3 (that
>   is, the complete length of TLSInnerPlaintext).

ekr: I see the value of @davidben's point about version-specific logic, but I think the point that @Lekensteyn makes about padding is pretty persuasive. ISTM that there are two sensible resolutions here:

- Move the record limit draft to talk about ciphertext, thus capping the maximum memory consumption under the theory that the worst-case is no padding. And then the plaintext limit would be "whatever fits in here minus 256, which is our worst-case expansion overhead"
- Have this limit talk about the total size of the plaintext, including CT and padding.

-Ekr

martinthomson: I prefer the latter because it is more directly coupled to how records are produced.  

I also think that padding needs to be included (or enforcing the limit risks exposing when padding is in use).  In the simple case, you can simply say that the space available for plaintext is one less than the size limit (the version specific logic is `if (tls13) --limit;`, which is manageable).

Talking about ciphertext means that you have to include knowledge of the ciphersuite in your calculations.  In TLS 1.3, that's relatively simple, but for CBC modes in other versions, it's terrible.

Lekensteyn: I would also prefer the second option. The first option would "waste" some space by enforcing a plaintext limit equal to the worst-case overhead.

draft-ietf-tls-tls13.md
```diff
@@ -4073,10 +4073,10 @@ a non-zero octet in the cleartext, it MUST terminate the
 connection with an "unexpected_message" alert.
 
 The presence of padding does not change the overall record size limitations
-- the full encoded TLSInnerPlaintext MUST not exceed 2^14 octets. If the maximum
-fragment length is reduced, as for example by the max_fragment_length extension
-from [RFC6066], then the reduced limit applies to the full plaintext,
-including the padding.
+- the full encoded TLSInnerPlaintext MUST not exceed 2^14 + 1 octets. If the
```
davegarrett: nit: please capitalize the "not", per RFC 2119

Lekensteyn: done (+ another case)

martinthomson:
So, if we go with 2^14+1 for sizeof(TLSInnerPlaintext), that is fine.  That means that you can (absent padding) expect to be able to send 2^14 octets in every record, across every TLS version.

The issue that this causes is one for the record size limit draft.  The valid maximum for that will be 2^14 in TLS 1.2 and earlier, but 2^14+1 in TLS 1.3.  That's OK, but I'll need to change that draft to make that observation.  We already need to allow for larger values in case an extension expands the record size, so this will be a good test for that code.  It's also going to be lots more annoying to write the cross-checking in code, but that is just how things go.

PR #1086, opened 2017-10-24
Allow the padding extension to change on HRR.

davidben:
The padding extension is typically computed as part of serializing the
ClientHello, in hopes of targetting a particular size. As specified right now,
the second ClientHello must not use this same logic and instead must retain the
previous extension sizethough this would likely not hit the same target size.

The server's going to ignore it anyway, so allow it to be recalcuated. This
avoids unnecessary state in clients and extra logic to serialize the first and
second ClientHellos differently.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1787,6 +1787,9 @@ ClientHello (without modification) except:
   any PSKs which are incompatible with the server's indicated
   cipher suite.
 
+- Optionally adding, removing, or changing the length of the "padding"
+  extension {{RFC7685}}.
+
 Because TLS 1.3 forbids renegotiation, if a server has negotiated TLS
 1.3 and receives a ClientHello at any other time, it MUST terminate
 the connection with an "unexpected_message" alert.


draft-ietf-tls-tls13.md
```diff
@@ -1787,6 +1787,9 @@ ClientHello (without modification) except:
   any PSKs which are incompatible with the server's indicated
   cipher suite.
 
+- Optionally adding, removing, or changing the length of the "padding"
+  extension {{RFC7685}}.
+
```
briansmith: Why not say it MUST be removed?

kaduk: Are the middleboxes that choke on particular sized ClientHellos not going to also choke on ClientHello2s?

Anyway, the whole point of the padding extension is to let you control record sizes on the wire; there seems to be no reason to restrict the available flexibility in one specific case.

davidben: Older TLS versions already don't have two ClientHellos, so if a middlebox is actually getting to interpret the second one as a ClientHello, we probably failed to interop much earlier. (The padding extension was originally for a confused endpoint.)

But yeah, despite being more spec text, it seemed less work all around for implementations to retain the flexibility. One could imagine maybe a datagram transport wanting something funny to apply to both messages. (QUIC does pad ClientHellos, but I don't know the details to know if cookied ones are padded or if the padding extension is suitable. It's a little weird since it's got a discontinuity after N-4 bytes.)

PR #1091, opened 2017-11-06
Middlebox changes2 hrr

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -1914,19 +1916,23 @@ Structure of this message:
...
-           ProtocolVersion version;
+           ProtocolVersion legacy_version = 0x0303;    /* TLS v1.2 */
            Random random;
+           opaque legacy_session_id_echo<0..32>;
            CipherSuite cipher_suite;
+           uint8 legacy_compression_method = 0;
            Extension extensions<6..2^16-1>;
        } ServerHello;
 
 version
```
nharper: nit: legacy_version

draft-ietf-tls-tls13.md
```diff
@@ -3924,7 +3979,7 @@ by an encrypted body, which itself contains a type and optional padding.
 
        struct {
            ContentType opaque_type = 23; /* application_data */
-           ProtocolVersion legacy_record_version = 0x0301; /* TLS v1.x */
+           ProtocolVersion legacy_record_version = 0x0303; /* TLS v1.2 */
```
nharper: I'm assuming this was changed to 0x0303 so that middleboxes which view the ClientHello and ServerHello and assume TLS 1.2 was negotiated can verify that the TLSCiphertext.record_version is 0x0303 as expected.

Is it worth adding language either here or in #middlebox explaining why the 0x0301 (in TLSPlaintext) and 0x0303 values were chosen?

draft-ietf-tls-tls13.md
```diff
@@ -5450,6 +5505,32 @@ ClientHello without 0-RTT data.
...
+
+- If not offering early data, the client sends a dummy
+  change_cipher_spec record immediately before its second flight. This
+  may either be before its second ClientHello or before its encrypted
+  handshake flight. If offering early data, the record is placed
+  immediately after the first ClientHello.
+
+- The server sends a dummy change_cipher_spec record immediately
+  after its first handshake message. This may either be after a
+  ServerHello or a HelloRetryRequest.
```
tmshort: I thought these were being combined?

ekr: I opted to treat it as a separate message for simplicity.

draft-ietf-tls-tls13.md
```diff
@@ -5450,6 +5505,32 @@ ClientHello without 0-RTT data.
...
+## Middlebox Compatibility Mode {#middlebox}
+
+Field measurements have found that a significant number of middleboxes
+misbehave when a TLS client/server pair negotiates TLS 1.3. Implementations
+can increase the chance of making connections through those middleboxes
+by making the TLS 1.3 handshake look more like a TLS 1.2 handshake:
+
+- The client always provides a non-empty session ID in the ClientHello.
+
+- If not offering early data, the client sends a dummy
```
tomato42: what does "dummy" mean here? It's a term not used anywhere else in the draft.

Does the CCS has to be always sent? If not, what the other side is supposed to do when it doesn't receive it? What if it receives it at different time? What if it receives multiple CCS messages one after another?

davidben: The receiver ignores all CCS records during the handshake, without any care as to where they do or don't appear. It's in the record layer section.

This is based on implementor feedback. Our initial prototype had the receiver enforce the CCS positions, but the handshake/CCS synchronization on the read side is a bit of a mess in TLS 1.2. There was a desire from some folks to not complicate their TLS 1.3 state machined with this. Discarding them at the record layer, on the other hand, is straightforward. (Additionally, it allows any implementation to unilaterally disable compatibility mode if it is deployed somewhere these middlebox issues are less common.)

draft-ietf-tls-tls13.md
```diff
@@ -3804,15 +3843,30 @@ then delivered to higher-level clients.
...
+An implementation may receive an unencrypted record of type
+change_cipher_spec consisting of the single byte value 0x01 at any
+time during the handshake and MUST simply drop it without further
+processing.  Note that this record may appear at a point at the
+handshake where the implementation is expecting protected records
+and so it is necessary to detect this
+condition prior to attempting to deprotect the record. An
+implementation which receives any other change_cipher_spec value or
+which receives a protected change_cipher_spec record MUST abort the
+handshake with an "illegal_parameter" alert. After the handshake is
```
mattcaswell: "unexpected_message" seems more appropriate

davidben: Agreed.

draft-ietf-tls-tls13.md
yuhong: Yea, I don't really like it either to have two version fields.

davidben: Me neither, but it's that or an insecure version fallback. Aesthetics and cleanliness are valuable, but not at the cost of security.

yuhong: But the same problem may repeat with future middleboxes.

tmshort: Future middleboxes should understand TLSv1.3... If they don't then it's the manufacturers fault.

davidben: Indeed. For the future, we need to better take into account how to defend our extension points when we design protocols. There's things like GREASE, but GREASE doesn't go far enough here. I have various ideas to try out on this front.

In the meantime, because of how much we've let ossify up to now, we sadly must live with certain constraints if we want TLS 1.3 to simultaneously be deployable and secure.

yuhong: Is this the most common problem though?

ekr: Common enough. I'm surprised you're bothered by this particular piece. We've already moved the ClientHello version out.

yuhong: I am talking a list with only one entry though, and you are talking about middleboxes not servers too.

davidben: We have a few local middleboxes to test with. The setup is nowhere near exhaustive (the variance on these things is *huge*), but it did allow us to iterate on this locally with a number of designs before testing a few out into the wild. This PR came out of what we've learned from all those experiments.

Variants that didn't fix the ServerHello version didn't even survive the local pass. :-(

yuhong: So do you know why these individual middleboxes did it?

yuhong: Also consider that most know or can find out if they are using middleboxes or not, so an "insecure version fallback" is not the only alternative.

ekr: Mostly the only way to know is to measure, and that's just a network attacker.

draft-ietf-tls-tls13.md
riking: Perhaps specify that a "dummy CCS record" is one consisting of a single `0x01` byte.

draft-ietf-tls-tls13.md
```diff
@@ -1653,7 +1653,7 @@ processed and transmitted as specified by the current active connection state.
            hello_verify_request_RESERVED(3),
            new_session_ticket(4),
            end_of_early_data(5),
-           hello_retry_request(6),
+           hello_retry_request_RESERVED(6),
```
Lekensteyn: nit: trailing space

draft-ietf-tls-tls13.md
```diff
@@ -1914,19 +1916,23 @@ Structure of this message:
...
-  For this version of the specification, the version is 0x0304.  (See
-  {{backward-compatibility}} for details about backward compatibility.)
+: In previous versions of TLS, this field was used for version negotiation
+  and represented the selected version number for the connection. Unfortunately,
+  some middleboxes fail when presented with new value. 
+  In TLS 1.3, the TLS server indicates its version using the
+  "supported_versions" extension ({{supported-versions}}),
+  and the legacy_version field MUST
+  be set to 0x0303, which is the version number for TLS 1.2.
+  (See {{backward-compatibility}} for details about backward compatibility.)
```
Lekensteyn: Should anything be said about the order of the SV extension relative to other extensions? The definition of extensions might change between versions.

draft-ietf-tls-tls13.md
yuhong: Is getting into an arms race with middleboxes really a good idea?

draft-ietf-tls-tls13.md
```diff
@@ -1937,12 +1943,24 @@ random
   This structure is generated by the server and MUST be
   generated independently of the ClientHello.random.
 
+legacy_session_id_echo
+: The contents of the client's legacy_session_id field. Note that
+  this field is echoed even if the client's value corresponded to
+  a cached pre-TLS 1.3 session which the server has chosen not
+  to resume. A client which receives a legacy_session_id field
+  which does not match what it sent in the ClientHello
```
kaduk: nit: s/which/that/

draft-ietf-tls-tls13.md
```diff
@@ -1937,12 +1943,24 @@ random
   This structure is generated by the server and MUST be
   generated independently of the ClientHello.random.
 
+legacy_session_id_echo
+: The contents of the client's legacy_session_id field. Note that
+  this field is echoed even if the client's value corresponded to
+  a cached pre-TLS 1.3 session which the server has chosen not
+  to resume. A client which receives a legacy_session_id field
```
kaduk: We should consider saying something about how the TLS 1.3-aware client must do the version check before the resumption check, in case some existing TLS 1.2 code does a resumption/sesion-id check early in processing and would be confused about what protocol is in use.

draft-ietf-tls-tls13.md
```diff
@@ -1914,19 +1916,23 @@ Structure of this message:
...
-  A client that receives a version that was not offered MUST abort the handshake.
-  For this version of the specification, the version is 0x0304.  (See
-  {{backward-compatibility}} for details about backward compatibility.)
+: In previous versions of TLS, this field was used for version negotiation
+  and represented the selected version number for the connection. Unfortunately,
+  some middleboxes fail when presented with new value. 
+  In TLS 1.3, the TLS server indicates its version using the
+  "supported_versions" extension ({{supported-versions}}),
+  and the legacy_version field MUST
+  be set to 0x0303, which is the version number for TLS 1.2.
```
kaduk: Do we want to specify a client alert to send when receiving some other value?

draft-ietf-tls-tls13.md
```diff
@@ -1937,12 +1943,24 @@ random
...
+  alert.
+
 cipher_suite
 : The single cipher suite selected by the server from the list in
   ClientHello.cipher_suites. A client which receives a cipher suite
   that was not offered MUST abort the handshake with an "illegal_parameter"
   alert.
 
+legacy_compression_method
+: A single byte of value 0.
```
kaduk: What should a client do if it receives something else?

draft-ietf-tls-tls13.md
kaduk: Down at line 2050 we (now?) have:

> +Upon receipt of a HelloRetryRequest, the client MUST perform the
 +checks specified in {{server-hello}} and then process the
 +extensions, starting with determining the version using
 +"supported_versions". Clients MUST abort the handshake with

draft-ietf-tls-tls13.md
```diff
@@ -2189,11 +2214,18 @@ be taken into account when designing new extensions:
 %%% Version Extension
 
        struct {
-           ProtocolVersion versions<2..254>;
+           select (Handshake.msg_type) {
+               case client_hello:
+                    ProtocolVersion versions<2..254>;
+                    
+               case server_hello:
+                    ProtocolVersion selected_version;
```
kaduk: There is probably some editorial (?) work to be done where we still talk about HRR all over the place (in the extensions table), etc., but there is no longer a hello_retry_request message type for the actual protocol elements.

draft-ietf-tls-tls13.md
```diff
@@ -2219,9 +2251,16 @@ to support prior versions of TLS SHOULD support TLS 1.2.
 Servers should be prepared to receive ClientHellos that include this
 extension but do not include 0x0304 in the list of versions.
 
-The server MUST NOT send the "supported_versions" extension. The
-server's selected version is contained in the ServerHello.version field as
-in previous versions of TLS.
+A server which negotiates TLS 1.3 MUST respond by sending a
+"supported_versions" extension containing the selected version value
+(0x0304). It MUST set the ServerHello.version field to 0x0303 (TLS
```
kaduk: legacy_version

draft-ietf-tls-tls13.md
```diff
@@ -3804,15 +3843,30 @@ then delivered to higher-level clients.
...
+time during the handshake and MUST simply drop it without further
+processing.  Note that this record may appear at a point at the
+handshake where the implementation is expecting protected records
+and so it is necessary to detect this
+condition prior to attempting to deprotect the record. An
+implementation which receives any other change_cipher_spec value or
+which receives a protected change_cipher_spec record MUST abort the
+handshake with an "illegal_parameter" alert. After the handshake is
+complete, change_cipher_spec MUST be treated as an unexpected record
+type.
```
kaduk: I wonder how much more complicated implementations will have to be to reject CCS post-handshake (as opposed to ignoring it always).

draft-ietf-tls-tls13.md
kaduk: > Is it worth adding language ...

Probably.

davidben: It's not 0x0301 for plaintext and 0x0303 for ciphertext. That wouldn't get through the middleboxes anyway. It's 0x0301 for ClientHello (a long-standing quirk from TLS 1.1) and 0x0303 for everything else. And yeah, you're right that it's to match TLS 1.2.

draft-ietf-tls-tls13.md
tomato42: +1. In case `legacy_version` is not set to 0x0303 and `supported_versions` is present, the connection should fail with `illegal_parameter`

though it probably should be placed in the `legacy_version` description to mirror how Client Hello is described

draft-ietf-tls-tls13.md
```diff
@@ -2001,42 +2032,36 @@ interoperability failure.
...
-contain sufficient information to proceed with the handshake.
-
-Structure of this message:
-
-%%% Key Exchange Messages
-
-       struct {
-           ProtocolVersion server_version;
-           CipherSuite cipher_suite;
-           Extension extensions<2..2^16-1>;
```
davidben: One detail I noticed as I was sorting out some of our tests: at one point we had text which disallowed no-op HelloRetryRequests. When HRR got moved to extensions, this turned into just a rule that the extensions block is non-empty. (That plus no unsolicited extensions did the trick.)

This is gone now, but it actually wouldn't have sufficed anyway because every HRR has a supported_versions extension now. So we probably need to go back to explicit text here somewhere.

draft-ietf-tls-tls13.md
```diff
@@ -5450,6 +5505,32 @@ ClientHello without 0-RTT data.
...
+Field measurements have found that a significant number of middleboxes
+misbehave when a TLS client/server pair negotiates TLS 1.3. Implementations
+can increase the chance of making connections through those middleboxes
+by making the TLS 1.3 handshake look more like a TLS 1.2 handshake:
+
+- The client always provides a non-empty session ID in the ClientHello.
+
+- If not offering early data, the client sends a dummy
+  change_cipher_spec record immediately before its second flight. This
+  may either be before its second ClientHello or before its encrypted
```
davidben: A somewhat irksome consequence of this: if you do completely stateless HRR, you need to skip the CCS before parsing your "first" (actually second) ClientHello and picking the version.

(Then again, we only need this over TCP which is stateful...)

draft-ietf-tls-tls13.md
```diff
@@ -2001,42 +2032,36 @@ interoperability failure.
...
 MUST abort the handshake with an "unexpected_message" alert.
 
-Otherwise, the client MUST process all extensions in the HelloRetryRequest and
-send a second updated ClientHello. The HelloRetryRequest extensions defined in
-this specification are:
+Otherwise, the client MUST process all extensions in the
+HelloRetryRequest and send a second updated ClientHello. The
+HelloRetryRequest extensions defined in this specification are:
+
+- supported_versions (see {{supported_versions}})
```
Lekensteyn: Reference must be `{{supported-versions}}` or the build fails.

draft-ietf-tls-tls13.md
tomato42: I meant that "dummy" is obvious to us, in this discussion, in the context of this PR. Not necessarily to somebody reading the spec in 5 years.

draft-ietf-tls-tls13.md
tomato42: `illegal_parameter` seems like a correct choice

draft-ietf-tls-tls13.md
ekr: Actually, we should ignore it to match how we handle ClientHello.legacy_version

ekr: And that's what the text already says.

draft-ietf-tls-tls13.md
ekr: No, I don't think we need to order it. You just make two passes.

draft-ietf-tls-tls13.md
ekr: Done.

draft-ietf-tls-tls13.md
ekr: Yes, but we don't need to say it here because illegal_parameter's definitions says o.

draft-ietf-tls-tls13.md
ekr: This is on line 2053.

draft-ietf-tls-tls13.md
ekr: Not very it seems.

draft-ietf-tls-tls13.md
ekr: I don't think dummy is confusing here.

PR #1094, opened 2017-11-15
Add text explaining why servers might want to vend multiple tickets

chris-wood:
Currently, the draft says that servers MAY send multiple tickets to clients without further discussion. It might help to be concrete and provide an example for when more than one ticket is useful. This PR provides one example.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3644,7 +3644,7 @@ For instance, the server might send a new ticket after post-handshake
 authentication in order to encapsulate the additional client
 authentication state. Multiple tickets are useful for clients that 
 perform connection racing across interfaces and address families 
-via, for example, Happy Eyeballs {{RFC6555}} or related techniques.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3641,9 +3642,11 @@ single connection, either immediately after each other or
 after specific events.
 For instance, the server might send a new ticket after post-handshake
 authentication in order to encapsulate the additional client
-authentication state. Clients SHOULD attempt to use each
-ticket no more than once, with more recent tickets being used
-first.
+authentication state. Multiple tickets are useful for clients that 
+perform connection racing across interfaces and address families 
+via, e.g., Happy Eyeballs {{RFC6555}} or related techniques.
```
ekr: If were going to add text, I think you should also  mention that you might do multiple HTTP connections

chris-wood: Done.

PR #1098, opened 2017-11-22
Be more explicit on PSS requirements

Lekensteyn:
This PR intends to clarify the requirements for PSS support.

The requirements are intentionally minimal to reduce implementation efforts, but recognizes that some other implementations may be more complete. Notes:

- "Supporting PSS signatures on certificates is a mandatory requirement and I think we should be very clear about the parameters we permit." https://www.ietf.org/mail-archive/web/tls/current/msg23007.html
- Martin Rex wishes to remove TLS requirements on signature algorithms for certificates, hence the "MAY" for other PSS parameters in this PR. https://www.ietf.org/mail-archive/web/tls/current/msg23021.html
- Regardless, rsa_pss_sha256 is currently MTI for CertificateVerify and certificates, hence the strong MUST wording in this PR.
- It does not say anything about non-end-entity certificates, that's up to the PKI verifier. Consider case "CA Key: rsa-pss; EE signature: rsa-pss; EE key: rsa" from https://www.ietf.org/mail-archive/web/tls/current/msg24453.html
- PSS params in certificates are explicitly not restricted, satisfying https://www.ietf.org/mail-archive/web/tls/current/msg24457.html

From what I have heard, boringssl does not (or will not?) implement any PSS support in the certificates (yet?). Don't know if anything should be changed here to reflect that decision, but I thought it is worth mentioning. It is possible that I'll follow boringssl's example in tris.

If a TLS extension is introduced later, hopefully that improves interop with odd keys and signatures that are optional in this PR (PSS pubkey or custom salt lengths).

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2349,8 +2349,12 @@ RSASSA-PSS algorithms
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is new in this document and is also
-  defined for use with TLS 1.2.
+  length of the digest output. An RSA public key (OID rsaEncryption) MUST be
+  supported, an RSASSA-PSS public key (OID id-RSASSA-PSS) MAY be supported.
+  When used as signature algorithm in certificates, the same parameters as used
+  for TLS handshake messages MUST be supported but other parameters MAY be used.
+  This codepoint is new in this document and is also defined for use with TLS
+  1.2.
 
 EdDSA algorithms
 : Indicates a signature algorithm using EdDSA as defined in
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -88,6 +88,7 @@ informative:
   RFC5077:
   RFC5116:
   RFC5246:
+  RFC5756:
   RFC5764:
   RFC5929:
   RFC6176:
@@ -2349,12 +2350,16 @@ RSASSA-PSS algorithms
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output. An RSA public key (OID rsaEncryption) MUST be
-  supported, an RSASSA-PSS public key (OID id-RSASSA-PSS) MAY be supported.
-  When used as signature algorithm in certificates, the same parameters as used
-  for TLS handshake messages MUST be supported but other parameters MAY be used.
+  length of the digest output.
+  When used as signature algorithm in certificates, the length of the salt
+  SHOULD be equal to the length of the digest output, but other salt lengths MAY
+  be used.
   This codepoint is new in this document and is also defined for use with TLS
   1.2.
+: An RSA public key (OID rsaEncryption) MUST be supported, an RSASSA-PSS public
+  key (OID id-RSASSA-PSS {{RFC5756}}) without parameters MUST be supported (such
+  that any hash algorithm and salt length can be used), while an RSASSA-PSS
+  public key (OID id-RSASSA-PSS) with parameters MAY be supported.
 
 EdDSA algorithms
 : Indicates a signature algorithm using EdDSA as defined in


draft-ietf-tls-tls13.md
```diff
@@ -2349,8 +2349,12 @@ RSASSA-PSS algorithms
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is new in this document and is also
-  defined for use with TLS 1.2.
+  length of the digest output. An RSA public key (OID rsaEncryption) MUST be
+  supported, an RSASSA-PSS public key (OID id-RSASSA-PSS) MAY be supported.
```
nmav: what about distinguishing the RSA-PSS keys with and without parameters:
```
an RSASSA-PSS public key (OID id-RSASSA-PSS) without parameters SHOULD be supported, while an RSASSA-PSS public key (OID id-RSASSA-PSS) with parameters MAY be supported`.
```

The reason is that keys explicitly marked as RSA-PSS cannot be used for RSA PKCS#1 1.5 encryption, and thus they provide a way for the server to know that it must protect that key against attacks which utilize RSA ciphersuites under TLS1.2.

tomato42: "(OID id-RSASSA-PSS)" should come with a reference to https://tools.ietf.org/html/rfc5756

Lekensteyn:
Added reference as suggested by @tomato42, marked id-RSASSA-PSS without params as mandatory as suggested by @nmav. Also changed the proposed text to refer explicitly to the salt length (the begin of the paragraph is already explicit about requiring MGF1).

Unchanged:
- No different codepoint for rsaEncryption vs id-RSASSA-PSS
- No requirement to forbid PSS pubkeys with params (a suggestion from @tomato42).

draft-ietf-tls-tls13.md
```diff
@@ -2349,12 +2350,16 @@ RSASSA-PSS algorithms
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output. An RSA public key (OID rsaEncryption) MUST be
-  supported, an RSASSA-PSS public key (OID id-RSASSA-PSS) MAY be supported.
-  When used as signature algorithm in certificates, the same parameters as used
-  for TLS handshake messages MUST be supported but other parameters MAY be used.
+  length of the digest output.
+  When used as signature algorithm in certificates, the length of the salt
+  SHOULD be equal to the length of the digest output, but other salt lengths MAY
```
nmav: > but other salt lengths MAY be used

I find that text a step backwards in terms of interoperability. The previous text of requiring that the salt length MUST be equal to the length of the digest output was specific and left no window of misunderstandings.

Lekensteyn: The intention is that use of a standard length is guaranteed, but if a CA somehow issues certs with odd salt lengths, then TLS would not outright forbid it (and leave it up to the PKI to perform policy decisions). See also https://github.com/tlswg/tls13-spec/pull/1109#discussion_r153673401

It was added based on comments from Martin Rex (I am not strongly in favor in keeping this freedom though).

nmav: It is a slippery slope. TLS currently specifies exactly the combo of parameters to use for RSA-PSS signing. The negotiation includes the hash and the salt size. What the next text makes, is make the TLS negotiation include only the hash algorithm, and leaves the application to deal with the salt size requirements outside TLS. I find the previous text simpler to implement and in alignment with having MUST in certificates with no parameters.

tomato42: how about "but longer salt lengths MAY be used"? AFAIR OpenSSL defaults to maximum allowed size (by public key size) of salt for rsa-pss signatures.

I think it would be silly to reject those certificates only because the salt is too large

Lekensteyn: Note that PKCS#1 v2.2 (https://tools.ietf.org/html/rfc8017#page-42) seems to suggest that zero-length salts are also okay:
>    4.  Typical salt lengths in octets are hLen (the length of the output
>       of the hash function Hash) and 0.  In both cases, the security of
>       RSASSA-PSS can be closely related to the hardness of inverting
>       RSAVP1.

Maybe requiring a single salt length is the best for interop purposes. Btw, I think I'm going to close this in favor of #1109, let's continue the discussion there.

tomato42: I have double-checked now - OpenSSL does default to maximum salt length possible for a given key size when making signatures - so requiring a salt length of the same size as the hash length would make all OpenSSL certificates made with default PSS settings invalid.

while 0 length salts are ok in RFC 8017, using different hash for MGF1 and signature hash is also ok, so I don't think that's the sole data point we will want to use

PR #1102, opened 2017-11-27
Syntax fix

kazu-yamamoto:
This PR fixes obvious syntax errors.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2531,6 +2531,8 @@ to perform post-handshake authentication {{post-handshake-authentication}}. Serv
 MUST NOT send a post-handshake CertificateRequest to clients which do not
 offer this extension. Servers MUST NOT send this extension.
 
+%%% Post Handshake Authentication Extension
+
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2531,6 +2531,8 @@ to perform post-handshake authentication {{post-handshake-authentication}}. Serv
 MUST NOT send a post-handshake CertificateRequest to clients which do not
 offer this extension. Servers MUST NOT send this extension.
 
+%%% Post Handshake Authentication Extension
```
Lekensteyn: I think that `%%% Server Parameters Messages` is more appropriate.

(otherwise it would have to be added to the `Protocol Data Structures and Constant Values` section)

kazu-yamamoto: It's up to @ekr. :-)

ekr: I concur with @Lekensteyn

ekr:
Merged by hand

PR #1105, opened 2017-11-28
add messages that are required to have the supported_versions extension

iluxonchik:
This is mentioned throughout the document, but not in the "Mandatory-to-Implement" section.
closes #1104

PR #1106, opened 2017-11-28
add messages that are required to have the supported_versions extension

### LINKED ISSUE #1104: supported_versions Extension Not Listed In "Mandatory-to-Implement" Section For ServerHello and HelloRetryRequest
In the `Supported Versions` section it states:

> A server which negotiates TLS 1.3 MUST respond by sending a "supported_versions" extension containing the selected version value (0x0304).

In the `Hello Retry Request` section we have:

> The server's extensions MUST contain "supported_versions".

However, this is not mentioned in the "Mandatory-to-Implement" extensions section.

iluxonchik:
This is mentioned throughout the document, but not in the "Mandatory-to-Implement" section.
closes #1104

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5033,7 +5033,7 @@ TLS-compliant application MUST implement the following TLS extensions:
 All implementations MUST send and use these extensions when offering
 applicable features:
 
-  * "supported_versions"   is REQUIRED for all ClientHello messages.
+  * "supported_versions"   is REQUIRED for all ClientHello, ServerHello and HelloRetryRequest messages.
   * "signature_algorithms" is REQUIRED for certificate authentication.
   * "supported_groups"     is REQUIRED for ClientHello messages using
                             DHE or ECDHE key exchange.


iluxonchik:
This is the same as #1105, just without the extra new line additions, etc

draft-ietf-tls-tls13.md
```diff
@@ -5033,7 +5033,7 @@ TLS-compliant application MUST implement the following TLS extensions:
 All implementations MUST send and use these extensions when offering
 applicable features:
 
-  * "supported_versions"   is REQUIRED for all ClientHello messages.
+  * "supported_versions"   is REQUIRED for all ClientHello, ServerHello and HelloRetryRequest messages.
```
Lekensteyn: I was under the impression that this section discusses the requirements for the Client Hello (and how the server uses the CH). SV is mentioned a few lines up, so it should probably not be mentioned again here.

iluxonchik: @Lekensteyn maybe I misinterpreted it, but then only thing mentioned a few lines up is that this extension is required to be supported, not that ServerHello must have it (before, the server chose it's version using the `ServerHello.version` field, just like in TLS 1.2).

I think you're right about that "this section discusses the requirements for the Client Hello (and how the server uses the CH)", since it states below "Servers receiving a ClientHello which does not conform to these requirements MUST abort the handshake with a "missing_extension" alert."

Lekensteyn: "Implementing the following TLS extensions" implies that both sides must follow the requirements of the linked section of the specification. An implementation cannot claim compliance if it fails to handle the (absence of) SV in SH.

iluxonchik: @Lekensteyn okay then, thank you. Should I close this pull request myself then or should it be closed by one of the maintainers?

Lekensteyn: @iluxonchik it's just my view, you could close it or let @ekr decide ;)

ekr: Actually, I think this is an improvement. I just missed it.

PR #1108, opened 2017-11-29
Hrr cleanup

ekr:


ekr:
I tend to narrow things from the right....

Lekensteyn:
> I tend to narrow things from the right....

Programmer habits maybe, for me a "FooTest" is a specialization of "Test", so the current "KeyShareServerHello" looks like a specialization of "ServerHello" to me. It's a minor thing though.

ekr:
Fair enough, but my habit is the opposite. :)

Issue #1089, opened 2017-11-06
post_handshake_auth


kazu-yamamoto:
The structure of `post_handshake_auth` is missing.

kaduk:
> The "extension_data" field of the "post_handshake_auth" extension is zero
length.

kazu-yamamoto:
OK. So, I would like to have either

```
Empty PostHandShakeAuth;
```

or 

```
struct {} PostHandShakeAuth;
```
explicitly.

Lekensteyn:
@ekr not sure if you are getting notifications for the comments to commits, but as a nitpick:

> PostHand**s**hakeAuth

(Handshake is a single word)

PR #1109, opened 2017-11-29
Calrify what the PSS code points mean. This is one end of the spectrum,

ekr:
specifically that it just means you support PSS signatures but with
rsaEncryption as the SPKI type. I think we should either require this
or alternately have the code point mean that I have full PSS support,
but in retrospect, I think that the empty parameters thing isn't that
helpful.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1910,7 +1910,7 @@ Structure of this message:
 version
 : In previous versions of TLS, this field was used for version negotiation
   and represented the selected version number for the connection. Unfortunately,
-  some middleboxes fail when presented with new value. 
+  some middleboxes fail when presented with new value.
   In TLS 1.3, the TLS server indicates its version using the
   "supported_versions" extension ({{supported-versions}}),
   and the legacy_version field MUST
@@ -2376,7 +2376,15 @@ RSASSA-PSS algorithms
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{!SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is new in this document and is also
+  length of the digest output. When used in certificate signatures, the
+  length of the salt MUST be at least as long as the length of the
+  digest output, and the algorithm parameters MUST be DER encoded.
+  This codepoint indicates that the sender
+  will accept certificates in which the public key has OID rsaEncryption,
+  but not necessarily OID id-RSASSA-PSS {{RFC5756}}, whether with
+  or without parameters. Future documents may register codepoints that
+  indicate support for id-RSASSA-PSS.
+  This codepoint is new in this document and is also
   defined for use with TLS 1.2.
 
 EdDSA algorithms


draft-ietf-tls-tls13.md
```diff
@@ -1910,7 +1910,7 @@ Structure of this message:
 version
 : In previous versions of TLS, this field was used for version negotiation
   and represented the selected version number for the connection. Unfortunately,
-  some middleboxes fail when presented with new value. 
+  some middleboxes fail when presented with new value.
```
grittygrease: "new value"->"new values"?

draft-ietf-tls-tls13.md
```diff
@@ -2376,7 +2376,15 @@ RSASSA-PSS algorithms
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{!SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is new in this document and is also
+  length of the digest output. When used in certificate signatures, the
+  length of the salt MUST be at least as long as the length of the
```
martinthomson: the "at least" here is problematic.  I would go with equal.

draft-ietf-tls-tls13.md
```diff
@@ -2376,7 +2376,15 @@ RSASSA-PSS algorithms
...
+  digest output, and the algorithm parameters MUST be DER encoded.
+  This codepoint indicates that the sender
+  will accept certificates in which the public key has OID rsaEncryption,
+  but not necessarily OID id-RSASSA-PSS {{RFC5756}}, whether with
```
martinthomson: the "not necessarily" phrasing here is a little distracting for me.  It implies that maybe you could use a key with a PSS OID.  That's pretty awkward.

draft-ietf-tls-tls13.md
Lekensteyn: Motivation might have come from https://www.ietf.org/mail-archive/web/tls/current/msg24454.html:
> it seems that some existing implementations use weird/unusual salt
> values for RSA-PSS signature with hash algs other than SHA-1.

Don't know about such systems though, does it refer to Microsoft's implementation?

Lekensteyn:
Build is probably failing because "RFC5756" is missing in the (informative) references section.

draft-ietf-tls-tls13.md
nmav: Does it matter whether other non-TLS systems have such constraints? I'd expect any implementation to be updated anyway for TLS1.3, and conform to it.

draft-ietf-tls-tls13.md
```diff
@@ -2376,7 +2376,15 @@ RSASSA-PSS algorithms
   digest used in the mask generation function and the digest being signed are
   both the corresponding hash algorithm as defined in {{!SHS}}. When used in
   signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is new in this document and is also
+  length of the digest output. When used in certificate signatures, the
+  length of the salt MUST be at least as long as the length of the
+  digest output, and the algorithm parameters MUST be DER encoded.
```
nmav: I'm not sure that this is the appropriate document to assert such a MUST. What can a TLS implementation do if a CA document allows including BER parameters?

nmav:
I find that this is actually introducing certificate type negotiation in TLS, and more specifically, negotiation of the container of the RSA-PSS parameters. I think that it is much simpler to require supporting the RSA-PSS certificate container if RSA-PSS keys are supported.

draft-ietf-tls-tls13.md
tomato42: I don't think BER is allowed anywhere in X.509

richsalz: It is.  Look for the word BER in the PKIX RFC 5280.

draft-ietf-tls-tls13.md
tomato42: @Lekensteyn OpenSSL does use maximum salt length allowed for the public key size (since version 1.0.1).
Unfortunately I don't know what's the behaviour of MS implementation.

@nmav because PKI structures that use RSA-PSS signatures are already deployed and implementing any changes in them is non-trivial. If we can be compatible with them by allowing a little wiggle room (flexible salt length), I think we should do that.

draft-ietf-tls-tls13.md
tomato42: I prefer here the #1089 wording:
```
 +: An RSA public key (OID rsaEncryption) MUST be supported, an RSASSA-PSS public
 +  key (OID id-RSASSA-PSS {{RFC5756}}) without parameters MUST be supported (such
 +  that any hash algorithm and salt length can be used), while an RSASSA-PSS
 +  public key (OID id-RSASSA-PSS) with parameters MAY be supported.
```

parameter-less rsa-pss is almost exactly the same as regular rsa, so adding support for it is not big change compared to the effort required to add TLS 1.3

draft-ietf-tls-tls13.md
nmav: In my previous comment I missed that `When used in certificate signatures,` so my it is not  to the point.  Seeing that phrase, I think it makes sense to link to rfc5756 instead of defining formally what to expect on such certificates.

draft-ietf-tls-tls13.md
tomato42: @richsalz the only reference to BER I see in 5280 is:

```
   Where the information is available via HTTP or FTP, accessLocation
   MUST be a uniformResourceIdentifier and the URI MUST point to either
   a single DER encoded certificate as specified in [RFC2585] or a
   collection of certificates in a BER or DER encoded "certs-only" CMS
   message as specified in [RFC2797].
```

so that reiterates it - if you talk about single certificate, it MUST be DER encoded...

tomato42: I don't think that they are allowed, it was that just some CAs were creating malformed (BER encoded) certificates and so support for it was added/ossified.

I have a vague recollection of seeing somewhere among the PKCS#12 related RFCs that any structure that is signed must be DER encoded...

draft-ietf-tls-tls13.md
```diff
@@ -2376,7 +2376,15 @@ RSASSA-PSS algorithms
...
+  This codepoint indicates that the sender
```
kaduk: There is more than one RSASSA-PSS codepoint, so "these" is probably more appropriate.

tomato42: which codepoint you refer to? there is only rsassaPss OID. What's variable is the parameters field in the Algorithm Identifier struct

Lekensteyn: sha256, sha384, etc.

PR #1112, opened 2017-11-30
Clarify certificate extensions

### LINKED ISSUE #1111: Client Certificate extensions: SCT, OCSP Stapling
status_request and signed_certificate_status timestamps are allowed in the CertificateRequest message according to https://tools.ietf.org/html/draft-ietf-tls-tls13-22#page-41

Certificate extensions (https://tools.ietf.org/html/draft-ietf-tls-tls13-22#page-66) however say:

>  An extension MUST only
>  be present in a Certificate message if the corresponding
>  ClientHello extension was presented in the initial handshake.

Is this also the intended behavior for client certificates, or should it instead require extensions to be advertised in CertificateRequest? (This is relevant for future extensions.)

No idea why a client would bother sending SCT or status_request extensions, but that is not forbidden by the TLS 1.3 spec. The status_request (RFC 6066) and [SCT](https://tools.ietf.org/html/rfc6962#section-3.3.1) extensions are written such that it seems unusable in client certs though.

Lekensteyn:
The peer must advertise extensions before they can be sent with a
certificate message.

Fixes #1111

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3325,11 +3325,12 @@ certificate_list
 
 extensions:
 : A set of extension values for the CertificateEntry. The "Extension"
-  format is defined in {{extensions}}. Valid extensions include
-  OCSP Status extension ({{RFC6066}}) and
-  SignedCertificateTimestamps ({{!RFC6962}}).  An extension MUST only be present
-  in a Certificate message if the corresponding
-  ClientHello extension was presented in the initial handshake.
+  format is defined in {{extensions}}. Valid extensions for server certificates
+  include OCSP Status extension ({{RFC6066}}) and
+  SignedCertificateTimestamps ({{!RFC6962}}). Extensions in the Certificate
+  message from the server MUST correspond to one from the ClientHello message,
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3325,11 +3325,12 @@ certificate_list
...
-  OCSP Status extension ({{RFC6066}}) and
-  SignedCertificateTimestamps ({{!RFC6962}}).  An extension MUST only be present
-  in a Certificate message if the corresponding
-  ClientHello extension was presented in the initial handshake.
+  format is defined in {{extensions}}. Valid extensions for server certificates
+  include OCSP Status extension ({{RFC6066}}) and
+  SignedCertificateTimestamps ({{!RFC6962}}). Extensions in the Certificate
+  message from the server MUST correspond to one from the ClientHello message,
+  extensions in the Certificate from the client MUST correspond with an
+  extension in the CertificateRequest message from the server.
```
kaduk: That's a comma splice (switch to semicolon before "extensions in the Certificate from the client MUST [...]" to fix).

ekr:
I fixed it with a period.

PR #1114, opened 2017-12-05
Clarify pss2

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -2298,15 +2298,19 @@ Clients MUST NOT use cookies in their initial ClientHello in subsequent connecti
 TLS 1.3 provides two extensions for indicating which signature
 algorithms may be used in digital signatures. The
 "signature_algorithms_cert" extension applies to signatures in
-certifiates and the "signature_algorithms" extension, which originally
+certificates and to the and the "signature_algorithms" extension, which originally
```
Lekensteyn: remove "and the"

draft-ietf-tls-tls13.md
```diff
@@ -2298,15 +2298,19 @@ Clients MUST NOT use cookies in their initial ClientHello in subsequent connecti
...
+appropriate type for the signature algorithms they are used
+with. This is a particular issue for RSA keys and PSS signatures,
+as described below. If no "signature_algorithms_cert" extension is present,
 then the "signature_algorithms" extension also applies to signatures
 appearing in certificates. Clients which desire the server to authenticate
 itself via a certificate MUST send "signature_algorithms". If a server
 is authenticating via a certificate and the client has not sent a
 "signature_algorithms" extension, then the server MUST abort the
 handshake with a "missing_extension" alert (see {{mti-extensions}}).
+TLS 1.2 implementations SHOULD also process this extension.
```
Lekensteyn: There are two extensions, maybe be explicit about the cert one?

draft-ietf-tls-tls13.md
```diff
@@ -2293,14 +2295,24 @@ Clients MUST NOT use cookies in their initial ClientHello in subsequent connecti
...
+appropriate type for the signature algorithms they are used
+with. This is a particular issue for RSA keys and PSS signatures,
+as described below. If no "signature_algorithms_cert" extension is present,
+then the "signature_algorithms" extension also applies to signatures
+appearing in certificates. Clients which desire the server to authenticate
+itself via a certificate MUST send "signature_algorithms". If a server
+is authenticating via a certificate and the client has not sent a
+"signature_algorithms" extension, then the server MUST abort the
+handshake with a "missing_extension" alert (see {{mti-extensions}}).
+TLS 1.2 implementations SHOULD also process this extension.
```
nmav: That's a large deviation from the -21 draft, for the benefit of specific platforms. I don't believe that text should be merged in TLS1.3 as is. The requirement:
```
If a server
+is authenticating via a certificate and the client has not sent a
+"signature_algorithms" extension, then the server MUST abort the
+handshake with a "missing_extension" alert (see {{mti-extensions}}).
```
is punishing clients which do not need to send this extension, i.e., clients which their validator has no issue handling RSA-PSS. Why this isn't an optional extension?

draft-ietf-tls-tls13.md
```diff
@@ -2364,35 +2380,49 @@ RSASSA-PKCS1-v1_5 algorithms
...
   generation function 1. The
   digest used in the mask generation function and the digest being signed are
-  both the corresponding hash algorithm as defined in {{!SHS}}. When used in
-  signed TLS handshake messages, the length of the salt MUST be equal to the
-  length of the digest output.  This codepoint is new in this document and is also
-  defined for use with TLS 1.2.
+  both the corresponding hash algorithm as defined in {{!SHS}}.
+  The length of the salt MUST be equal to the length of the digest
+  algorithm. If the public key is carried
+  in an X.509 certificate, it MUST use the rsaEncryption OID.
```
nmav: Please spell out the OID if there is no reference given.

draft-ietf-tls-tls13.md
```diff
@@ -2364,35 +2380,49 @@ RSASSA-PKCS1-v1_5 algorithms
...
   "PureEdDSA" algorithms and not the "prehash" variants.
 
+RSASSA-PSS PSS algorithms
+: Indicates a signature algorithm using RSASSA-PSS {{RFC8017}} with mask
+  generation function 1. The
+  digest used in the mask generation function and the digest being signed are
+  both the corresponding hash algorithm as defined in {{!SHS}}.
+  The length of the salt MUST be equal to the length of the digest
+  algorithm. If the public key is carried in an X.509 certificate,
+  it MUST use the RSASSA-PSS OID. When used in certificate signatures,
```
nmav: Same for OID. Please spell it out.

Lekensteyn: RFC5756 would probably be a good reference (https://tools.ietf.org/html/rfc5756)

draft-ietf-tls-tls13.md
ekr: This requirement has been in TLS 1.3 for some time. See:
https://tlswg.github.io/tls13-spec/draft-ietf-tls-tls13.html#rfc.section.4.2.3

In fact, it's required to be there because the default "signature_algorithms" value (from 5246) does not indicate that you support SHA-256.

draft-ietf-tls-tls13.md
```diff
@@ -2293,14 +2295,24 @@ Clients MUST NOT use cookies in their initial ClientHello in subsequent connecti
...
-desire the server to authenticate itself via a certificate MUST send this extension.
-If a server is authenticating via a certificate and the client has not sent a
-"signature_algorithms" extension, then the server MUST
-abort the handshake with a "missing_extension" alert (see {{mti-extensions}}).
-
-The "extension_data" field of this extension in a ClientHello contains a
+TLS 1.3 provides two extensions for indicating which signature
+algorithms may be used in digital signatures. The
+"signature_algorithms_cert" extension applies to signatures in
+certificates and to the and the "signature_algorithms" extension, which originally
```
tomato42: typo: "the and" unnecessary

draft-ietf-tls-tls13.md
```diff
@@ -2293,14 +2295,24 @@ Clients MUST NOT use cookies in their initial ClientHello in subsequent connecti
...
+appeared in TLS 1.2, applies signatures in CertificateVerify
```
tomato42: "applies to signatures"?

draft-ietf-tls-tls13.md
```diff
@@ -2364,35 +2380,49 @@ RSASSA-PKCS1-v1_5 algorithms
...
+RSASSA-PSS PSS algorithms
+: Indicates a signature algorithm using RSASSA-PSS {{RFC8017}} with mask
+  generation function 1. The
+  digest used in the mask generation function and the digest being signed are
+  both the corresponding hash algorithm as defined in {{!SHS}}.
+  The length of the salt MUST be equal to the length of the digest
+  algorithm. If the public key is carried in an X.509 certificate,
+  it MUST use the RSASSA-PSS OID. When used in certificate signatures,
+  the algorithm parameters MUST be DER encoded and identical to those
+  used in the corresponding public key.
```
tomato42: doesn't that prohibit use of parameter-less RSA-PSS SPKI?

draft-ietf-tls-tls13.md
```diff
@@ -2364,35 +2380,49 @@ RSASSA-PKCS1-v1_5 algorithms
...
   {{RFC8032}} or its successors. Note that these correspond to the
   "PureEdDSA" algorithms and not the "prehash" variants.
 
+RSASSA-PSS PSS algorithms
+: Indicates a signature algorithm using RSASSA-PSS {{RFC8017}} with mask
+  generation function 1. The
+  digest used in the mask generation function and the digest being signed are
+  both the corresponding hash algorithm as defined in {{!SHS}}.
+  The length of the salt MUST be equal to the length of the digest
+  algorithm. If the public key is carried in an X.509 certificate,
```
kaduk: So now we are enforcing salt length restrictions on signatures in X509 certificates (both here and above)?

PR #1115, opened 2017-12-06
Tighten language on middlebox

richsalz:
Use "network intermediary" in all but once place.
Add forward/backward links to Appendix D
Add requirement to SHOULD use compatibility.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1747,7 +1747,7 @@ it MUST abort the handshake with either
 a "handshake_failure" or "insufficient_security" fatal alert
 (see {{alert-protocol}}).
 
-###  Client Hello
+###  Client Hello {#client_hello}
 
 When a client first connects to a server, it is REQUIRED to send the
 ClientHello as its first message. The client will also send a
@@ -1821,10 +1821,12 @@ legacy_session_id
   feature which has been merged with Pre-Shared Keys in this version
   (see {{resumption-and-psk}}). A client which has a cached session ID
   set by a pre-TLS 1.3 server SHOULD set this field to that value. In
-  compatibility mode (see {{middlebox}}) this field MUST be non-empty,
+  compatibility mode (see {{intermediary}}), this field MUST be non-empty,
   so a client not offering a pre-TLS 1.3 session MUST generate a
-  new 32-byte value. This value need not be random but SHOULD be
-  unpredictable to avoid ossification.
+  new 32-byte value. This value need not be a cryptographically-secure
+  random value, but SHOULD be
+  unpredictable to avoid intermediaries depending on a specific value
+  (also known as ossification).
   Otherwise, it MUST be set as a zero length vector (i.e., a single
   zero byte length field).
 
@@ -1912,7 +1914,7 @@ Structure of this message:
 version
 : In previous versions of TLS, this field was used for version negotiation
   and represented the selected version number for the connection. Unfortunately,
-  some middleboxes fail when presented with new values. 
+  some network intermediaries fail when presented with new values.
   In TLS 1.3, the TLS server indicates its version using the
   "supported_versions" extension ({{supported-versions}}),
   and the legacy_version field MUST
@@ -1956,8 +1958,8 @@ extensions
   message.
 {:br }
 
-For backward compatibility reasons with middleboxes
-(see {{middlebox}}) the HelloRetryRequest
+For backward compatibility reasons with network intermediaries
+(see {{intermediary}}), the HelloRetryRequest
 message uses the same structure as the ServerHello, but with
 Random set to the special value of the SHA-256 of
 "HelloRetryRequest":
@@ -3885,7 +3887,7 @@ encrypted with the new key. Failure to do so may allow message truncation
 attacks.
 
 
-#  Record Protocol
+#  Record Protocol {#dummy-ccs}
 
 The TLS record protocol takes messages to be transmitted, fragments
 the data into manageable blocks, protects the records, and transmits
@@ -3895,9 +3897,10 @@ then delivered to higher-level clients.
 TLS records are typed, which allows multiple higher-level protocols to
 be multiplexed over the same record layer. This document specifies
 four content types: handshake, application data, alert, and
-change_cipher_spec, with the latter being used only for
-compatibility purposes (see {{middlebox}}).
+change_cipher_spec.
 
+The change_cipher_spec record is used only for compatibility purposes
+(see {{intermediary}}).
 An implementation may receive an unencrypted record of type
 change_cipher_spec consisting of the single byte value 0x01 at any
 time during the handshake and MUST simply drop it without further
@@ -4054,7 +4057,7 @@ zeros
 opaque_type
 : The outer opaque_type field of a TLSCiphertext record is always set to the
   value 23 (application_data) for outward compatibility with
-  middleboxes accustomed to parsing previous versions of TLS.  The
+  network intermediaries accustomed to parsing previous versions of TLS.  The
   actual content type of the record is found in TLSInnerPlaintext.type after
   decryption.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -3896,9 +3899,9 @@ TLS records are typed, which allows multiple higher-level protocols to
 be multiplexed over the same record layer. This document specifies
 four content types: handshake, application data, alert, and
 change_cipher_spec, with the latter being used only for
-compatibility purposes (see {{middlebox}}).
+compatibility purposes (see {{intermediary}}).
 
-An implementation may receive an unencrypted record of type
+{#dummy_css}An implementation may receive an unencrypted record of type
```
kaduk: Are we talking cascading style sheets or change cipher spec?

draft-ietf-tls-tls13.md
```diff
@@ -5560,17 +5563,21 @@ ClientHello without 0-RTT data.
...
 can increase the chance of making connections through those middleboxes
 by making the TLS 1.3 handshake look more like a TLS 1.2 handshake:
 
-- The client always provides a non-empty session ID in the ClientHello.
+- The client always provides a non-empty session ID in the ClientHello as
+  described in {{legacy_session_id}}.
 
 - If not offering early data, the client sends a dummy
-  change_cipher_spec record immediately before its second flight. This
+  change_cipher_spec record (see {{dummy_css}}
```
kaduk: (and here)

draft-ietf-tls-tls13.md
```diff
@@ -5584,7 +5591,10 @@ TLS 1.2 session resumption, which improves the chance of successfully
 connecting through middleboxes. This "compatibility mode" is not explicitly
 negotiated. The client can opt to provide a session ID or not
 and the server has to echo it. Either side can send change_cipher_spec
-at any time during the handshake, as they must be ignored by the peer.
+at any time during the handshake, as it must be ignored by the peer.
+
+In the absence of out of band information, endpoints SHOULD use the
+techniques described here.
```
kaduk: For all time?

richsalz: Yes  Think about it.  If you know it's time to stop doing it, then you have out of band information.

draft-ietf-tls-tls13.md
```diff
@@ -5560,17 +5563,21 @@ ClientHello without 0-RTT data.
...
 can increase the chance of making connections through those middleboxes
 by making the TLS 1.3 handshake look more like a TLS 1.2 handshake:
 
-- The client always provides a non-empty session ID in the ClientHello.
+- The client always provides a non-empty session ID in the ClientHello as
+  described in {{legacy_session_id}}.
 
 - If not offering early data, the client sends a dummy
-  change_cipher_spec record immediately before its second flight. This
+  change_cipher_spec record (see {{dummy_ccs}}
```
martinthomson: unbalanced paren

draft-ietf-tls-tls13.md
```diff
@@ -1817,14 +1817,17 @@ random
   See {{implementation-notes}} for additional information.
 
 legacy_session_id
-: Versions of TLS before TLS 1.3 supported a "session resumption"
+: {#legacy_session_id} Versions
```
Lekensteyn: I don't think that you can use inline anchors like that, the parser does not seem to provide such a feature: https://github.com/cabo/kramdown-rfc2629

draft-ietf-tls-tls13.md
```diff
@@ -5560,17 +5563,21 @@ ClientHello without 0-RTT data.
...
 and stable deployment of TLS 1.3 without 0-RTT prior to enabling 0-RTT.
 
-## Middlebox Compatibility Mode {#middlebox}
+## Network Intermediary Compatibility Mode {#intermediary}
 
-Field measurements have found that a significant number of middleboxes
-misbehave when a TLS client/server pair negotiates TLS 1.3. Implementations
+Field measurements have found that a significant number of network
+intermediaries, or "middleboxes,"
+can cause the TLS endpoints to fail to negotiate TLS 1.3.
```
Lekensteyn: Independent to this PR, but related: is it possible to provide a reference for these "field measurements"? Something like https://www.ietf.org/mail-archive/web/tls/current/msg24517.html (but preferably something more specific). Might be useful for the archeologists in the future.

richsalz: Perhaps bring that up on the list.  A non-normative link to @davidben's IETF-100 slides might be do-able.

richsalz:
Pushed an updated commit that fixes the paren and anchor issue, thanks.

draft-ietf-tls-tls13.md
```diff
@@ -3896,9 +3898,9 @@ TLS records are typed, which allows multiple higher-level protocols to
 be multiplexed over the same record layer. This document specifies
 four content types: handshake, application data, alert, and
 change_cipher_spec, with the latter being used only for
-compatibility purposes (see {{middlebox}}).
+compatibility purposes (see {{intermediary}}).
 
-An implementation may receive an unencrypted record of type
+{#dummy_ccs}An implementation may receive an unencrypted record of type
```
Lekensteyn: There is another unsupported inline anchor here. It seems also unhappy about the use of `client_hello`, maybe try `client-hello`?

richsalz: Thanks, I fixed this. And I moved a sentence from the 2nd to 3rd paragraph.

martinthomson:
I'm having a little trouble with the justification for the use of "network intermediary" here.  The term "middlebox" has a well-established definition (see RFC 3234 for example).  This would seem to be less precise.

The use of "intermediary" implies something more than what we've observed.  I understand "intermediary" to involve protocol participation - TLS termination effectively - and that is not the sort of middlebox that is causing us issues, as David has explained on list.

Also, I realize that you think that the SHOULD is helpful.  I would prefer that our default mode of operation avoid the use of this hacky mode.  Firefox just shipped a pre-release with -22 and the compatibility hacks disabled and we can report an improvement in compatibility in at least one case, even without sending ChangeCipherSpec.  I can see both angles, but we might have to just disagree on this point.

ekr:
On Wed, Dec 6, 2017 at 2:47 PM, Martin Thomson <notifications@github.com>
wrote
>
> Also, I realize that you think that the SHOULD is helpful. I would prefer
> that our default mode of operation avoid the use of this hacky mode.
> Firefox just shipped a pre-release with -22 and the compatibility hacks
> disabled and we can report an improvement in compatibility in at least one
> case, even without sending ChangeCipherSpec. I can see both angles, but we
> might have to just disagree on this point.
>
I go back and forth on the SHOULD, but it's something that would have to go
to the list...

Lekensteyn:
> I would prefer that our default mode of operation avoid the use of this hacky mode. Firefox just shipped a pre-release with -22 and the compatibility hacks disabled and we can report an improvement in compatibility in at least one case.

Do you test both with and without CCS hack? Have you found cases where using CCS breaks while not sending CCS works?

FWIW, in tris I have chosen for unconditionally sending CCS to simplify the implementation (less knobs and code branches that needs to be tested).

ekr:
On Wed, Dec 6, 2017 at 3:38 PM, Peter Wu <notifications@github.com> wrote:

> I would prefer that our default mode of operation avoid the use of this
> hacky mode. Firefox just shipped a pre-release with -22 and the
> compatibility hacks disabled and we can report an improvement in
> compatibility in at least one case.
>
> Do you test both with and without CCS hack?
>
We will be doing so shortly, but it's a 3-armed study (TLS 1.2, TLS 1.3-22,
TLS 1.3-compat), so we don't know how much we get from TLS-22 alone absent
historical controls.


> Have you found cases where using CCS breaks while not sending CCS works?
>
No, we do not have such cases.




FWIW, in tris I have chosen for unconditionally sending CCS to simplify the

richsalz:
If folks don't like the intermediary term, so be it.  But it's important to collect all the compatibility changes into one place with appropriate links.

ekr:
@richsalz: I added the links but omitted the name change and the SHOULD. At this point, a normative change like the SHOULD would need to go through the WG.

@richsalz: I see you aren't in acknowledgements. Please send me a PR to add yourself.

richsalz:
Yes I am, right after Joe :)
The changes you made are okay.

PR #1116, opened 2017-12-06
Clarify TLS protocol invariants.

davidben:
This clarifies the versioning rules used by TLS. They are immediate
corrollaries from how we've always evolved the protocol, but draft-22
demonstrates that there are widely-deployed non-compliant intermediaries
in the network. Document the rules more clearly.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1058,6 +1058,9 @@ An implementation of TLS 1.3 that also supports TLS 1.2 might need to include
 changes to support these changes even when TLS 1.3 is not in use.  See the
 referenced sections for more details.
 
+Additionally, this document clarifies some compliance requirements in TLS 1.2,
+as described in {{protocol-invariants}}.
+
 
 # Protocol Overview
 
@@ -5068,6 +5071,58 @@ Servers requiring this extension SHOULD respond to a ClientHello
 lacking a "server_name" extension by terminating the connection with a
 "missing_extension" alert.
 
+
+## Protocol Invariants
+
+This section describes invariants that TLS endpoints and intermediaries MUST
+follow. It also applies to earlier versions, which assumed these rules but did
+not document them.
+
+TLS is designed to be securely and compatibly extensible. Newer clients or
+servers, when communicating with newer peers, should negotiate the most
+preferred common parameters with downgrade protection.  In particular, it is a
+security requirement (see {{security-handshake}}) that intermediaries passing
+traffic between a newer client and newer server without terminating TLS are
+unable to influence the handshake. At the same time, deployments update at
+different rates, so a newer client or server MAY continue to support older
+parameters, which would allow it to interoperate with older endpoints.
+
+For this to work, implementations MUST correctly handle extensible fields:
+
+- A client sending a ClientHello MUST support all parameters advertised in it.
+  Otherwise, the server may fail to interoperate by selecting one of those
+  parameters.
+
+- A server receiving a ClientHello MUST correctly ignore all unrecognized
+  cipher suites, extensions, and other parameters. Otherwise, it may fail to
+  interoperate with newer clients.
+
+- An intermediary which terminates a TLS connection MUST behave as a compliant
+  TLS server (to the original client) and as a compliant TLS client (to the
+  original server). In particular, it MUST generate its own ClientHello
+  containing only parameters it understands, and it MUST generate a fresh
+  ServerHello random value, rather than forwarding the endpoint's value.
+
+  In this scenario, the intermediary is a pair of TLS endpoints for purposes
+  of protocol requirements and security analysis. In most TLS deployments this
+  would require configuring one or both endpoints to trust the intermediary's
+  certificate authority.
+
+- An intermediary which forwards ClientHello parameters it did not generate MUST
+  NOT process any messages beyond that ClientHello. It MUST forward all
+  subsequent traffic unmodified. Otherwise, it may fail to interoperate with
+  newer clients and servers.
+
+  Forwarded ClientHellos may contain advertisements for features not supported
+  by the intermediary, so the response may include future TLS additions the
+  intermediary does not recognize. These additions MAY change any message beyond
+  the ClientHello arbitrarily. In particular, the values sent in the ServerHello
+  MAY change, the ServerHello format MAY change, and the TLSCiphertext format
+  MAY change.
+
+This specification was constrained by widely-deployed non-compliant TLS
+intermediaries (see {{middlebox}}), however it does not relax the invariants.
+Those intermediaries continue to be non-compliant.
 
 #  Security Considerations
 


draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
 
 
+## Protocol Invariants
+
+This section describes invariants that TLS endpoints and intermediaries MUST
+follow. It also applies to earlier versions, which assumed these rules but did
+not document them.
+
+TLS is designed to be securely and compatibly extensible. Newer clients or
+servers, when communicating with newer peers, should negotiate the most
```
ekr: SHOULD. Also, I note you are being vague about "preferred". TLS has generally assumed that the server picks the *highest* version but otherwise has discretion about how to pick the rest of the params.

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
+preferred common parameters with downgrade protection.  In particular, it is a
```
ekr: I'm not sure what the downgrade protection text means here.

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
+
+For this to work, implementations MUST correctly handle extensible fields:
+
+- A client sending a ClientHello MUST support all parameters advertised in it.
+  Otherwise, the server may fail to interoperate by selecting one of those
+  parameters.
+
+- A server receiving a ClientHello MUST correctly ignore all unrecognized
+  cipher suites, extensions, and other parameters. Otherwise, it may fail to
+  interoperate with newer clients.
```
ekr: In TLS 1.3 you need to add some text about the client handling NST and CR extensions.

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
+
+  In this scenario, the intermediary is a pair of TLS endpoints for purposes
+  of protocol requirements and security analysis. In most TLS deployments this
+  would require configuring one or both endpoints to trust the intermediary's
+  certificate authority.
+
+- An intermediary which forwards ClientHello parameters it did not generate MUST
+  NOT process any messages beyond that ClientHello. It MUST forward all
+  subsequent traffic unmodified. Otherwise, it may fail to interoperate with
+  newer clients and servers.
```
ekr: This seems like it might be a little strong. If it forwards only parameters it understands, then it can presumably could minimally syntax check the response, right?

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
+
+  Forwarded ClientHellos may contain advertisements for features not supported
+  by the intermediary, so the response may include future TLS additions the
+  intermediary does not recognize. These additions MAY change any message beyond
```
ekr: This MAY isn't a normative statement, so probably not capitalized. I think I would do s/MAY/might/ for the rest of this graf.

draft-ietf-tls-tls13.md
```diff
@@ -1058,6 +1058,9 @@ An implementation of TLS 1.3 that also supports TLS 1.2 might need to include
 changes to support these changes even when TLS 1.3 is not in use.  See the
 referenced sections for more details.
 
+Additionally, this document clarifies some compliance requirements in TLS 1.2,
+as described in {{protocol-invariants}}.
```
richsalz: It should probably be " 1.2; see" for style consistency with the rest of the DOC.

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
+- An intermediary which terminates a TLS connection MUST behave as a compliant
+  TLS server (to the original client) and as a compliant TLS client (to the
+  original server). In particular, it MUST generate its own ClientHello
+  containing only parameters it understands, and it MUST generate a fresh
+  ServerHello random value, rather than forwarding the endpoint's value.
+
+  In this scenario, the intermediary is a pair of TLS endpoints for purposes
+  of protocol requirements and security analysis. In most TLS deployments this
+  would require configuring one or both endpoints to trust the intermediary's
+  certificate authority.
```
richsalz: I think this item is a bit misguided in that it lumps CDN's with network devices. I say that not to defend my employer :) but rather because tls-terminating intermediaries has been the problem area.  Maybe that confusion can be cleaned up by a better definition in the first sentence.

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
 "missing_extension" alert.
 
 
+## Protocol Invariants
+
+This section describes invariants that TLS endpoints and intermediaries MUST
+follow. It also applies to earlier versions, which assumed these rules but did
```
richsalz: this implies all earlier versions, the forward-pointer at the top only says 1.2

draft-ietf-tls-tls13.md
davidben: No, TLS-terminating devices have *not* been the problem area for these purposes. It's the non-terminating devices (which forward the ClientHello but then incorrectly try to parse out the response), and the weird "hybrid" cases where a TLS-terminating device has aspects of a non-terminating one. See my slides from IETF 100.

A TLS-terminator is "easy". You take a TLS server and TLS client and connect them back-toback. Your obligations are thus to be a correct server on one end and a correct client on the other. These obligations hold for both CDNs and MITM boxes. Your server needs to ignore unknown values in the ClientHello. Your client needs to not send ClientHello values it doesn't understand.

Thankfully, I haven't heard of any CDNs that broke since you all just do the obvious thing, while some MITM boxes try to be clever about it (like forwarding the original ClientHello along and then sometimes terminate anyway) and make invalid assumptions.

richsalz: Okay, then I still think you made my point.  Should we even talk about TLS-terminating entities, other than to say it's like two TLS connections?  The problems in this protocol were caused by non-terminating, izzat right?

A TLS terminator has no additional requirements beyond any other TLS terminator.  That's a tautology, but it seems to be what you're saying in your comment.

I still think this item needs more clarity.

draft-ietf-tls-tls13.md
kaduk: "is expected to negotiate the most-preferred common parameters, and any attempt to interfere with the negotiation MUST result in either failure of the interference or of the handshake" takes a slightly different approach.
Though that is rather duplicating the next sentence...

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
+traffic between a newer client and newer server without terminating TLS are
+unable to influence the handshake. At the same time, deployments update at
+different rates, so a newer client or server MAY continue to support older
+parameters, which would allow it to interoperate with older endpoints.
+
+For this to work, implementations MUST correctly handle extensible fields:
+
+- A client sending a ClientHello MUST support all parameters advertised in it.
+  Otherwise, the server may fail to interoperate by selecting one of those
+  parameters.
```
kaduk: Do we rather want to say "the client will fail to interoperate if the server selects one of those parameters"?

draft-ietf-tls-tls13.md
kaduk: "the intermediary's certificate authority" is a bit weird, since we don't exactly mean "the certificate authority that issued the intermediary's certificate" and rather "a certificate authority controlled by the intermediate".

draft-ietf-tls-tls13.md
```diff
@@ -5069,6 +5072,58 @@ lacking a "server_name" extension by terminating the connection with a
...
+  newer clients and servers.
+
+  Forwarded ClientHellos may contain advertisements for features not supported
+  by the intermediary, so the response may include future TLS additions the
+  intermediary does not recognize. These additions MAY change any message beyond
+  the ClientHello arbitrarily. In particular, the values sent in the ServerHello
+  MAY change, the ServerHello format MAY change, and the TLSCiphertext format
+  MAY change.
+
+This specification was constrained by widely-deployed non-compliant TLS
```
kaduk: I wonder if it would be more clear to s/This specification/TLS 1.3/ since we've been talking about some other semi-related things for a few paragraphs.

PR #1117, opened 2017-12-07
Allow a future extension to allow external PSKs and certificate-based…

russhousley:
… authentication

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5747,7 +5747,8 @@ the PSK binder. {{PSK-FINISHED}}
 describes a concrete attack on constructions that do not bind to
 the server's certificate (see also {{Kraw16}}). It is unsafe to use certificate-based client
 authentication when the client might potentially share the same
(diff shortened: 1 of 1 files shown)


iluxonchik:
I'm inclined to agree with @kaduk. A `TLS` extension can change anything it wants about the protocol, which means that in the limit, the text you suggested can be added next to every prohibition present in the draft, and while it does make sense to put it some places, this doesn't seem to be such a critical part that needs it mentioned, it just bulks up the spec. 

If this edit is accepted, however, it might be a good idea to be consistent with the already existing wording and use 
> `unless negotiated by some extension`
 
instead, which is also less wordy and cognitively complex than the current one, while conveying the same information.

davidben:
Agreed with others. I won't be unhappy if the text is there, but I think it's redundant and makes an already rather long specification ever so slightly longer. Any extension is a minor (or major, in this document!) protocol revision and is thus allowed to define whatever semantics or protocol tweaks it needs to.

PR #1120, opened 2017-12-10
fix .version mentions with either .legacy_version or supported_versions

iluxonchik:
Some of the places in the draft were still using .version field
referrences in ServerHello and HelloRetryRequest, those were replaced
with either .legacy_version field or "supporeted_versions" extension.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5475,8 +5475,8 @@ TLSPlaintext.legacy_record_version MUST be ignored by all implementations.
 The value of TLSCiphertext.legacy_record_version MAY be ignored,
 or MAY be validated to match the fixed constant value.
 Version negotiation is performed using only the handshake versions
-(ClientHello.legacy_version, as well as the
-ClientHello, HelloRetryRequest and ServerHello "supported_versions" extension).
+(ClientHello.legacy_version, ServerHello.legacy_version, as well as the
+ClientHello, HelloRetryRequest and ServerHello "supported_versions" extensions).
 In order to maximize interoperability with older endpoints, implementations
 that negotiate the use of TLS 1.0-1.2 SHOULD set the record layer
 version number to the negotiated version for the ServerHello and all
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1874,7 +1874,7 @@ ClientHello. Note that this method of detecting optional data differs
 from the normal TLS method of having a variable-length field, but it
 is used for compatibility with TLS before extensions were defined.
 TLS 1.3 servers will need to perform this check first and only
-attempt to negotiate TLS 1.3 if a "supported_version" extension
+attempt to negotiate TLS 1.3 if the "supported_versions" extension
 is present.
 If negotiating a version of TLS prior to 1.3, a server MUST check that
 the message either contains no data after legacy_compression_methods
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2259,7 +2259,7 @@ RFC EDITOR: PLEASE REMOVE THIS SECTION
 While the eventual version indicator for the RFC version of TLS 1.3 will
 be 0x0304, implementations of draft versions of this specification SHOULD
 instead advertise 0x7f00 | draft_version
-in ServerHello and HelloRetryRequest "supported_versions" extension.
+in the ServerHello and HelloRetryRequest "supported_versions" extension.
 For instance, draft-17 would be encoded as the 0x7f11.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -5475,8 +5475,8 @@ TLSPlaintext.legacy_record_version MUST be ignored by all implementations.
 The value of TLSCiphertext.legacy_record_version MAY be ignored,
 or MAY be validated to match the fixed constant value.
 Version negotiation is performed using only the handshake versions
-(ClientHello.legacy_version,
-ClientHello "supported_versions" extension, and ServerHello.version).
+(ClientHello.legacy_version, as well as
+ClientHello, HelloRetryRequest and ServerHello "supported_versions" extension).
```
iluxonchik: This is wrong. HelloRetryRequest should be removed (will do in a rebase).

draft-ietf-tls-tls13.md
```diff
@@ -2259,7 +2259,7 @@ RFC EDITOR: PLEASE REMOVE THIS SECTION
 While the eventual version indicator for the RFC version of TLS 1.3 will
 be 0x0304, implementations of draft versions of this specification SHOULD
 instead advertise 0x7f00 | draft_version
-in ServerHello.version, and HelloRetryRequest.server_version.
+in ServerHello and HelloRetryRequest "supported_versions" extension.
```
kaduk: I think there is a "the" missing here.

draft-ietf-tls-tls13.md
```diff
@@ -5475,16 +5475,16 @@ TLSPlaintext.legacy_record_version MUST be ignored by all implementations.
 The value of TLSCiphertext.legacy_record_version MAY be ignored,
 or MAY be validated to match the fixed constant value.
 Version negotiation is performed using only the handshake versions
-(ClientHello.legacy_version,
-ClientHello "supported_versions" extension, and ServerHello.version).
+(ClientHello.legacy_version, as well as
+ClientHello and ServerHello "supported_versions" extension).
```
kaduk: "as well as the ClientHello and ServerHello "supported_versions" extensions".
It's not clear whether the HelloRetryRequest specialization of ServerHello requires a separate callout here.

draft-ietf-tls-tls13.md
```diff
@@ -5475,16 +5475,16 @@ TLSPlaintext.legacy_record_version MUST be ignored by all implementations.
...
 In order to maximize interoperability with older endpoints, implementations
 that negotiate the use of TLS 1.0-1.2 SHOULD set the record layer
 version number to the negotiated version for the ServerHello and all
 records thereafter.
 
 For maximum compatibility with previously non-standard behavior and misconfigured
 deployments, all implementations SHOULD support validation of certification paths
-based on the expectations in this document, even when handling prior TLS versions'
+based on the expectations in this document, even when handling prior TLS versions
```
kaduk: I don't think this change is correct -- "versions'" with a trailing apostrophe is the possessive form of the plural "versions", i.e., "the handshakes of prior versions of TLS".

iluxonchik: @kaduk you're correct, thank you.

draft-ietf-tls-tls13.md
iluxonchik: @kaduk thank you, I agree, will add it.

draft-ietf-tls-tls13.md
iluxonchik: @kaduk isn't `extension` (singular) the correct form? We're talking about the same extension, not multiple ones. Your observation applies to my other edit as well, I made it explicit in the previous case, so I'll add it here too.

kaduk: It's kind of subtle.  To start off with, we're talking about version negotiation, which means we care about the contents of the extension blocks, not the type off extension.  And there are distinct instances of the same extension type in ClientHello and ServerHello, potentially holding different contents (used for negotiation).  So the presence of different instances justifies the use of the plural, to me.

iluxonchik: @kaduk yeah, I see where you're coming from. Actually since we're on it, shouldn't I add the `ServerHello.legacy_version` to the text too? I missed that one.

iluxonchik: @kaduk I added `ServerHello.legacy_version`, as well as accepted your grammar suggestion in 0ef3784

ekr:
@iluxonchik I merged this by hand. Please send me a PR to add your name to the acknowledgements

iluxonchik:
@ekr okay :smile: I'm in the process of setting up a new domain and a new e-mail server, so I will do a PR as soon as the process is done.

PR #1122, opened 2017-12-10
fix Backward Compatibility section paragraph

### LINKED ISSUE #1121: Confusing Paragraph In Backward Compatibility Section
In the `Appendix D. Backward Compatibility` section, you have the following paragraph:

> TLS 1.x and SSL 3.0 use compatible ClientHello messages.  Servers can
   also handle clients trying to use future versions of TLS as long as
   the ClientHello format remains compatible and the client supports the
   highest protocol version available in the server.

This sounds confusing. Saying that servers can handle clients negotiating future versions, as long as "the client supports the highest protocol version available in the server" doesn't seem to make a lot of sense for that purpose (also, why only the highest protocol version?). In my opinion, something like the following version is a lot clearer:

> TLS 1.x and SSL 3.0 use compatible ClientHello messages.  Servers can
   also handle clients trying to use future versions of TLS as long as
   the ClientHello format remains compatible and there is at lest one protocol version that both the client and the server support.

This doesn't specify if the supported version must be that "future version" or some other version that both support, so this covers both of the cases.

I will submit a pull request shortly.

iluxonchik:
closes #1121

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5465,8 +5465,8 @@ endpoints potentially supporting different versions of TLS.
 
 TLS 1.x and SSL 3.0 use compatible ClientHello messages. Servers can also handle
 clients trying to use future versions of TLS as long as the ClientHello format
-remains compatible and and there is at lest one protocol version that both the
-client and the server support.
+remains compatible and and there is at least one protocol version supported by
+both the client and the server.
 
 Prior versions of TLS used the record layer version number for various
 purposes. (TLSPlaintext.legacy_record_version and TLSCiphertext.legacy_record_version)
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -5465,8 +5465,8 @@ endpoints potentially supporting different versions of TLS.
 
 TLS 1.x and SSL 3.0 use compatible ClientHello messages. Servers can also handle
 clients trying to use future versions of TLS as long as the ClientHello format
-remains compatible and the client supports the highest protocol version available
-in the server.
+remains compatible and and there is at lest one protocol version that both the
+client and the server support.
```
kaduk: I might word it as "there is at least one protocol version supported by both the client and the server", but that's editorial.
Overall this seems an okay direction to go in.

iluxonchik: @kaduk agree, your version is easier to read.

PR #1123, opened 2017-12-11
Added section on cross-protocol attacks

nmav:


draft-ietf-tls-tls13.md
```diff
@@ -5846,6 +5846,33 @@ then erase the \[early_]exporter_master_secret, followed by
 each inner values as soon as it is known that it will not be
 needed again.
 
+### Cross-protocol private key re-use
```
martinthomson: Sentence Case for Titles

draft-ietf-tls-tls13.md
```diff
@@ -5846,6 +5846,33 @@ then erase the \[early_]exporter_master_secret, followed by
...
+attacks which re-use a private key of the same type across protocols,
+when used for digital signatures.
+
+However, in order to provide an easier migration to the new protocol, and
+for compatibility with older systems which do not fully support new types of
+keys and certificates (e.g., RSASSA-PSS certificates), this protocol does
+not enforce separation of the private key and certificate for different
+protocol versions and usage type. That is, it allows re-using keys marked
+for PKCS#1 1.5 encryption and digital signatures under TLS1.2, for
+RSASSA-PSS signatures under TLS1.3.
```
martinthomson: Space between name and number.  "TLS 1.3".

draft-ietf-tls-tls13.md
```diff
@@ -5846,6 +5846,33 @@ then erase the \[early_]exporter_master_secret, followed by
...
+
+That private key sharing between protocols, does not isolate against private
+key attacks which take advantage of vulnerabilities of earlier protocols.
+For example, when re-using the same RSA key for legacy and
+TLS1.3 protocols, a successful Bleichenbacher attack against the TLS1.2 RSA
+ciphersuites will recover the key used for TLS1.3 signatures.
+
+For that reason, it is RECOMMENDED to utilize separate keys between
```
martinthomson: https://grammarpartyblog.com/2012/01/17/use-versus-utilize/

davidben:
Does the versioning advice (as opposed to the advice to separate PKCS1 and PSS keys) actually do anything? We put a context string in the signature, so cross-protocol woes where both protocols perform the same private key operation are fine (e.g. ECDSA keys or RSA-PSS in both 1.2 and 1.3).

As for PSS and PKCS1 separation, I don't think unilateral separation at the server will do anything:

For PSS signing against PKCS1 signing, I find it hard to imagine a scenario where separating the keys at signing would help given that RFC 4055 already does not separate them at verification and the signing inputs include fresh server entropy. And the fact that they *weren't* separated was the whole reason TLS 1.3 thought it could get away with switching to PSS in the same revision that added it. If the WG believes they ought to be separated, it should burn the rsaEncryption provision in RFC 4055 and then allow PKCS1 again. Otherwise, if we're not burning that provision, we should get some analysis on sharing a key here as any text we put in forbidding it is just wishful thinking. (Given that lots of software today accepts RSA-PSS X.509 signatures on rsaEncryption SPKIs per RFC 4055, that analysis is probably needed regardless.)

For PSS signing against PKCS1 decryption, there is indeed the same concern as PKCS1 signing against PKCS1 decryption, that your PKCS1 decryption logic had better not be vulnerable to the Bleichenbacher attack. But note that key separation on the server does not save you from this attack. A successful Bleichenbacher attack on one of the server's keys means I have a raw decryption oracle and can perform any operation I like, at whatever TLS version I like, with the key.

That is, when the attacker's got a Bleichenbacher oracle, what matters isn't what the server *uses* the key for, but what the client *accepts* it for. Though you do not need anything cross-protocol anyway as most clients still accept static RSA. Removing that would be lovely but, sadly, still far from practical. (If the client does *not* accept static RSA and *does* enforce the key usage bit on the certificate *then* you are probably okay.)

PS: Bleichenbacher only gives you an RSA decryption oracle, not the key itself, no? So "will recover the key used for TLS1.3 signatures" is not really right.

draft-ietf-tls-tls13.md
nmav: thanks

draft-ietf-tls-tls13.md
nmav: replaced

draft-ietf-tls-tls13.md
nmav: replaced

nmav:
> I think that this could be said more succinctly. That is, something like:
>>    Use of the same private key for different versions of TLS exposes that key to attacks in all versions of the protocol. Using a different key for each version isolates each key from vulnerabilities in other versions.

Thank you. I like that text and I updated to include it. However I believe there should be more concrete analysis pointing to the actual issues, thus I've also included references.

> Then the specific recommendations:
>>    A deployment of TLS 1.3 that supports earlier versions SHOULD use different keys for each version. For RSA keys, a deployment SHOULD use different keys for RSASS-PSS signatures, PKCS#1v1.5 signatures, and RSA key transport.

You take that proposal one step further. What about making more specific what flags should be there to allow implementation distinguishing these keys?
```
For RSA keys, a deployment SHOULD use different keys for RSASS-PSS signatures, PKCS#1v1.5 signatures, and RSA key transport. The certificates of these keys SHOULD be marked for their usage (see Key Usage of {{RFC5280}}) with the keyEncipherment --for RSA key transport-- and digitalSignature --for RSA signatures-- flags.
```

nmav:
> Does the versioning advice (as opposed to the advice to separate PKCS1 and PSS keys) actually do anything? We put a context string in the signature, so cross-protocol woes where both protocols perform the same private key operation are fine (e.g. ECDSA keys or RSA-PSS in both 1.2 and 1.3).

That versioning only protects when the key is used for signatures only. It wouldn't do anything if you use it for different purpose (decryption and signatures). I made that more apparent in the text.

> For PSS signing against PKCS1 decryption, there is indeed the same concern as PKCS1 signing against PKCS1 decryption, that your PKCS1 decryption logic had better not be vulnerable to the Bleichenbacher attack.

Right. Isn't however the point of the security analysis section to underline these risks? The pace under which we find new vulnerabilities related to that original attack in 90's shows that we should consider a world where it is hard to make a RSA decryption not vulnerable to a variant to the Bleichenbacher attack.

 
> But note that key separation on the server does not save you from this attack. A successful Bleichenbacher attack on one of the server's keys means I have a raw decryption oracle and can perform any operation I like, at whatever TLS version I like, with the key.

Could you elaborate on that? If the key is separated with the right key usage flags, could there be an attack against TLS1.3 with a key used for TLS1.2 key transport marked for keyEncipherment?

> Bleichenbacher only gives you an RSA decryption oracle, not the key itself, no? So "will recover the key used for TLS1.3 signatures" is not really right.

Thanks, that's correct. I've updated the text.

tomato42:
@davidben 
> For PSS signing against PKCS1 decryption, there is indeed the same concern as PKCS1 signing against PKCS1 decryption, that your PKCS1 decryption logic had better not be vulnerable to the Bleichenbacher attack.

yet it is something that significant portion of top internet domains were vulnerable to ([ROBOT](https://robotattack.org/)), despite two freely available and Open Source test suites being able to detect it

> PS: Bleichenbacher only gives you an RSA decryption oracle, not the key itself, no?

no, it also allows you to sign arbitrary messages; so having ability to mark a key as "never use it for RSA key exchange" is valuable

draft-ietf-tls-tls13.md
```diff
@@ -5846,6 +5890,31 @@ then erase the \[early_]exporter_master_secret, followed by
...
+Use of the same private key for different versions of TLS exposes that
+key to attacks in all versions of the protocol, as demonstrated by
+the {{DROWN}} attack. Previous versions of TLS were further susceptible
+to cross-protocol attacks between different ciphersuites {{WAGNER}}.
+This version of TLS takes steps to thwart cross-protocol attacks between
+different ciphersuites, and between different protocol versions sharing
+the same private key (see {{certificate-verify}}) for the same purpose.
+
+However, it does not mandate the separation of the private key and
+certificate for different protocol versions. It is, possible,
```
tomato42: > , possible,

these commas seem excessive to me

draft-ietf-tls-tls13.md
```diff
@@ -5846,6 +5890,31 @@ then erase the \[early_]exporter_master_secret, followed by
...
+to reuse keys for PKCS#1 1.5 RSA decryption under TLS 1.2,
+for RSASSA-PSS signatures under TLS 1.3, enabling attacks
+similar to {{DROWN}} to affect TLS 1.3 deployments.
+For that reason, it is RECOMMENDED to
+
+- Use separate keys between different protocol versions;
+- Use an RSASSA-PSS-only key with its corresponding certificate for TLS 1.3.
```
tomato42: "Use a key with public key ID of RSASSA-PSS for TLS 1.3" ?

tomato42:
> You take that proposal one step further. What about making more specific what flags should be there to allow implementation distinguishing these keys?

while I very like that idea, I'm afraid that that ship has sailed, sunk, was dredged up and now sits in a museum...

davidben:
The new text is better in that it explains the motivations more clearly (that you are concerned about the signing vs. decryption collision, not the signing vs. signing collision), but the recommendations are neither necessary nor sufficient to address those concerns. They're also a bit redundant.

Answering your question above may make the the problem clearer:

>> But note that key separation on the server does not save you from this attack. A successful Bleichenbacher attack on one of the server's keys means I have a raw decryption oracle and can perform any operation I like, at whatever TLS version I like, with the key.
>
> Could you elaborate on that? Would the amendment based on @martinthomson 's suggestion on the comment above address that?

It would not. Remember key separation is not a unilateral thing. Both client and server must know to separate the keys. The proposed split point, TLS versions, is not reflected in the certificate. Suppose we have a server with a TLS 1.2 key and a TLS 1.3 key which implemented static RSA badly and exposed a Bleichenbacher oracle. The attacker can still use the TLS 1.2 key to sign for TLS 1.3, which the client will happily accept. The client doesn't know this is the TLS 1.2 key and that a separate TLS 1.3 key exists.

Moreover, one does not have to attack TLS 1.3. The client probably still supports TLS 1.2, so you can sign a TLS 1.2 ServerKeyExchange. Or one could just attack TLS 1.2 static RSA because the client still supports that one too. (Much as I would love to lose that, usage rates are such that it's not remotely plausible right now.) Put it another way: key separation would not save you from DROWN because the attacker would just use your SSLv2 key to speak SSLv3/TLS.

The advice also does too much in other scenarios. With an ECDSA key, there are no signing vs decryption concerns. The CertificateVerify construction works fine, because *we want and designed TLS 1.3 to be usable with the same keys as TLS 1.2*. We wish for TLS 1.3 to be widely deployable and thus with minimal friction. You should be able to update your system TLS library and have TLS 1.3 available. That implies we need the same keys would work. Moreover, whatever wishful thinking we put in the text, sharing keys is just how people deploy TLS today, and will continue to be how they deploy TLS tomorrow. Key management is hard. It will take a while for all of today's server tooling to be replaced with fancy automated ACME things.

Otherwise we'd just drop RSA altogether. Maybe even ECDSA. Just declare Ed25519 MTI for TLS 1.3.

I believe what you wanted is:

* If you can, turn off static RSA. That thing is terrible. But this is also wishful thinking and won't happen. Too much legacy.

* If you can't do that, separate RSA *signing and decrypt* keys *with separate key usage bits in the certificate*.

The key usage part is important for this to meaningful. That prevents the client from accepting a forged signature off of that key. This is not worthless, but it's only of marginal value:

* Key management is still hard. No one will actually do this.

* In practice, clients don't actually check the cipher-specific RSA key usage bit. My understanding is CAs and antivirus MITMs and the like have massively screwed this up. Chrome checks it for TLS 1.3 because that's new, but I need to get some metrics for what it'd take to clear through this one.

* But it doesn't matter that no one check because clients still support static RSA. The attacker can still attack that one. So we're still stuck needing to implement static RSA without the Bleichenbacher oracle. It's a pity because, as we repeatedly learn, folks keep messing that up. But this PR does not fix that.

Now, what could be interesting is some extension in the certificate that says "I am new enough to do ECDHE. If you are also new enough to do ECDHE, reject this certificate, even for static RSA". That would make key separation actually do something. A new extension would also clear whatever legacy mess we have with key usage bits. But I don't think that's for this specification.

draft-ietf-tls-tls13.md
nmav: thanks, that seems like a typo

draft-ietf-tls-tls13.md
nmav: I do not think we can easily make it more specific; the subject public key identifier which I believe you refer to is only on certificates. PKCS#8 private keys use the algorithmIdentifier field. I think it is better to keep it generic as there are quite few private key types, and what could an RSASSA-PSS-only key may mean on each will vary.

nmav:
@tomato42 
>>    You take that proposal one step further. What about making more specific what flags 
>>should be there to allow implementation distinguishing these keys?
> while I very like that idea, I'm afraid that that ship has sailed, sunk, was dredged up and now sits in a museum...

That may be the case for <TLS1.2, but I do not think that's the case for TLS1.3. At TLS1.3 one can simply define that digitalSignature is the only allowed flag.

nmav:
@davidben 
> Moreover, one does not have to attack TLS 1.3. The client probably still supports TLS 1.2, so you can sign a TLS 1.2 ServerKeyExchange. Or one could just attack TLS 1.2 static RSA because the client still supports that one too. (Much as I would love to lose that, usage rates are such that it's not remotely plausible right now.) Put it another way: key separation would not save you from DROWN because the attacker would just use your SSLv2 key to speak SSLv3/TLS.

If I can read your point right, it narrows down that if that proposal doesn't protect TLS1.2 clients which do RSA, it is not worth it. I understand, though I disagree. That proposal can  keep TLS1.3 communications safe, even if there are issues under TLS1.2, and I've made that explicit now in the text. 

> With an ECDSA key, there are no signing vs decryption concerns.

There is. An ECDSA key is actually in ECDH key with different key usage flags. The same protection should be there, and I've included that in the text.

>    If you can, turn off static RSA. That thing is terrible. But this is also wishful thinking and won't happen. Too much legacy.

Too many attacks also. One was just yesterday. That's why I believe TLS1.3 shouldn't stay silent on that. It should recommend against using it, and that's what this proposal aims for.

> If you can't do that, separate RSA signing and decrypt keys with separate key usage bits in the certificate. he key usage part is important for this to meaningful. That prevents the client from accepting a forged signature off of that key. This is not worthless, but it's only of marginal value

As I mentioned above, I disagree here. The value is being able to ensure that TLS1.3 communications are secure, even if your TLS1.2 implementation is vulnerable. That's not marginal in my opinion.

> In practice, clients don't actually check the cipher-specific RSA key usage bit. My understanding is CAs and antivirus MITMs and the like have massively screwed this up. Chrome checks it for TLS 1.3 because that's new, but I need to get some metrics for what it'd take to clear through this one.

Tell me about this. However, just because TLS1.3 is new, we have the opportunity to clarify these details on the protocol, to allow these or new implementors to get it right. In fact for TLS1.3 the story can be very simple; only the 'digitalSignature' flag is allowed. This proposal may not even be followed by all (it's a recommendation after all), however, it is the point to refer to when one wants to prevent attacks on TLS1.x to propagate to TLS1.3.

Note that this proposal doesn't affect interoperability in any way. RSA key transport is still allowed. It however makes sure that if you use it, you will not be able to attack clients which check the key/certificate flags.

davidben:
> If I can read your point right, it narrows down that if that proposal doesn't protect TLS1.2 clients which do RSA, it is not worth it. I understand, though I disagree. That proposal can keep TLS1.3 communications safe, even if there are issues under TLS1.2, and I've made that explicit now in the text.

 No, you did not read my point right.

Key separations not reflected in the certificate do not do anything. Your text proposes a versions split. Your table proposes a key usages AND versions split. The versions split does not do anything. If your TLS 1.2 key is compromised, the attacker can use that to speak TLS 1.3 because the client has no clue that the key isn't used for TLS 1.3.

From the first message:

> A successful Bleichenbacher attack on one of the server's keys means I have a raw decryption oracle and can perform any operation I like, *at whatever TLS version I like*, with the key.

From the second message:

> Suppose we have a server with a TLS 1.2 key and a TLS 1.3 key which implemented static RSA badly and exposed a Bleichenbacher oracle. *The attacker can still use the TLS 1.2 key to sign for TLS 1.3, which the client will happily accept.* The client doesn't know this is the TLS 1.2 key and that a separate TLS 1.3 key exists.

Making noise about separating "TLS 1.2 RSA signatures" and "TLS 1.3 RSASSA-PSS signatures" is useless because the client will accept the former keys in the latter context.

Separately, yes, none of the key separations you describe will protect clients which accept static RSA. I do indeed think that makes them not worth it, given how hard key management is and how long it will be before such clients actually exist. (I will note that the text doesn't even say what protections its recommendation does and doesn't provide.) If no one will follow the recommendation, we should not clutter what is already an overly long specification with it.

> There is. An ECDSA key is actually in ECDH key with different key usage flags. The same protection should be there, and I've included that in the text.

Sure, recommending that ECDSA keys only be used for ECDSA and marked via key usage is reasonable. Though I expect people are already getting this right.

> Tell me about this. However, just because TLS1.3 is new, we have the opportunity to clarify these details on the protocol, to allow these or new implementors to get it right. In fact for TLS1.3 the story can be very simple; only the 'digitalSignature' flag is allowed.

Indeed. But this PR does not clarify that. It instead goes on about splitting keys by version.

> Note that this proposal doesn't affect interoperability in any way. RSA key transport is still allowed. It however makes sure that if you use it, you will not be able to attack clients which check the key/certificate flags.

The second iteration of the text, which I was commenting on, did *not* do this. Your text said to separate by protocol version, not by key usage. Were on to do only that, the TLS 1.2 key would have both keyEncipherment AND digitalSignature while the TLS 1.3 key have only digitalSignature. You still could use the former for the latter.

Now your text says to separate by versions, but your table by versions AND key usage. I'm going to assume you meant what your table says. That works, but only one of your proposed separations does anything.

ekr:
I also don't think this is the right PR.

From my perspective, the operational constraint is that servers be able to
use their existing RSA certificates which generally need to have both
keyEncipherment (to use static RSA) and digitalSignature (to use (EC)DHE)
set for TLS 1.3. This means that we can't require (or even encourage)
clients to enforce that keyEncipherment is not present.

From my perspective, the threat that needs to be noted is the use of
a Bleichenbacher-type oracle from an RSA-encryption capable TLS 1.2 or
below server to construct digital signatures that can be used in the
context of TLS 1.3 (as described by Jaeger et al). Unfortunately,
I don't believe that that threat can be prevented as long as:

(a) the server is prepared to do TLS 1.2 static RSA (and of course
    is vulnerable)
(b) the client will accept the same key for TLS 1.3 as is used with
    TLS 1.2 (which, as I said, is a requirement)

Having the server just use separate keys doesn't help, as long as the
client will accept the key intended for static RSA.

In other words, I think it would be fine to say something to the effect
of:

    Although TLS 1.3 does not use RSA key transport and so is not
    directly susceptible to Bleichenbacher-type attacks, if TLS 1.3
    servers also support static RSA in the context of previous
    versions of TLS, then it may be possible to impersonate the server
    for TLS 1.3 connections, as shown by Ja\"{e}ger et el [REF]. TLS
    1.3 implementations can prevent this attack by disabling support
    for static RSA across all versions of TLS.

I don't think anything else is useful, for the reasons indicated by
David Benjamin.

ekr:
Note that you *could* address this problem by having two keys with one you just used for static RSA and marked for keyEncipherment and the other used for signature and marked as digitalSignature. If we assume clients enforce then this would help, but it's a big pain for the server because it means two new certs, and so I doubt people will do it.

nmav:
> Note that you could address this problem by having two keys with one you just used for static RSA and marked for keyEncipherment and the other used for signature and marked as digitalSignature. If we assume clients enforce then this would help, but it's a big pain for the server because it means two new certs, and so I doubt people will do it.

Some servers already do that. See for example 'www.google.com'. It uses different certificates with different key usage flags for RSA and *DHE-RSA ciphersuites, so I believe it is feasible and it makes sense to recommend it as good practice which defends against this attack. That's what my last update does.

Security-wise the best defense is to forbid non-RSASSA-PSS certificates for TLS1.3. That would eliminate any clash of algorithms, at the cost of not being able to re-use your old RSA key under TLS1.3. However, given the importance of the latter, that's no longer an option.

nmav:
@davidben Very nice comments. I believe I got your point. Indeed there is no real separation of protocols (since we cannot mandate RSASSA-PSS certs for TLS1.3). The only separation that can be achieved only between RSA encryption under TLS1.3 and RSA signatures under TLS1.x.

davidben:
> See for example 'www.google.com'. It uses different certificates with different key usage flags for RSA and *DHE-RSA ciphersuites, so I believe it is feasible and it makes sense to recommend it as good practice which defends against this attack.

I'm pretty sure our servers do not currently do this. :-)

davidben:
(Is it possible you got switched from an RSA to an ECDSA certificate when you did that test? We do, of course, use different certificates for RSA and ECDHE-ECDSA.)

nmav:
> (Is it possible you got switched from an RSA to an  ECDSA certificate when you did that test? We do, of course, use  different certificates for RSA and ECDHE-ECDSA.)

Actually yes. Then I was wrong, no example of such good practice then :)

nmav:
@ekr why was this closed? Was it for the recommendation present in that section? A section on cross-protocol attacks is missing from the security analysis of the protocol.

ekr:
I added the following text, which I felt captured the relevant points.
https://tlswg.github.io/tls13-spec/draft-ietf-tls-tls13.html#rfc.appendix.E.6

nmav:
Indeed it captures the issue with the RSA ciphersuites. However, we have always seen new attacks taking advantage of obscure ciphersuites of the past, so in a security analysis, I wouldn't rule out cross-protocol attacks in general, even if we don't know any attack now.

ekr:
Nor would I, but I don't think there's anything particularly useful to say in this section about it.

PR #1128, opened 2017-12-27
Renumber key share

ekr:


mimoo:
isn't this change potentially allowing backdoored TLS implementations to work with TLS 1.3? If so I object to it :/ gotta break some bad eggs.

davidben:
TLS 1.3 has hit enough friction as it is. Renumbering allows us to separate the two pressure points, which is very valuable. Remember, the only way "break everything" strategies can ever work is if you do it gradually enough. Bundling them together is *not* desirable.

Moreover, these implementations presumably exist as both clients and servers. If we use 40 for TLS 1.3, trying to catch the clients becomes much more difficult. (Perhaps someone has already written a firewall that blocks extension 40 in ClientHellos!) If we keep them separate, this kind of strategy is still possible.

ekr:
I agree with DavidBen here. We could, if one wanted, have a separate RFC
telling people how to provide a dummy extension 40 for this purpose...

-Ekr

agl:
I support landing this change. We would like to get another draft published with this included so that we can try again at deploying TLS 1.3 in Chrome.

PR #1129, opened 2017-12-29
Clarify when in the handshake we ignore CCS records

mattcaswell:


draft-ietf-tls-tls13.md
```diff
@@ -2344,6 +2344,12 @@ new ClientHello, the client MUST copy the contents of the extension received in
 the HelloRetryRequest into a "cookie" extension in the new ClientHello.
 Clients MUST NOT use cookies in their initial ClientHello in subsequent connections.
 
+When a server is operating statelessly it may receive an unprotected record of
+type change_cipher_spec between the first and second ClientHello (see
+{{Record Protocol}}). Since the server is not storing any state this will appear
```
martinthomson: this link won't work, try `{{record-protocol}}`

mattcaswell: Fixed!

PR #1132, opened 2018-01-03
Tweak protocol invariants text a little.

davidben:
- There are many sorts of TLS middleboxes (CDN, decrypting proxy, local
  antivirus, actual attack, etc), with acceptabilities ranging from
  reasonable to extremely questionable to a terrible idea.
  (Categorization is left as an exercise to the reader.) They also have
  very different security considerations.

  Fundamentally, TLS secures a connection between two endpoints. As
  such, we have something to say about the endpoint-to-endpoint
  connections that make up such a system, but we shouldn't imply this
  section is *sufficient*, merely *necessary*. Adjust the text
  accordingly.

- I wrote "intermediary" in the original PR, predictively based on PR
  #1115. Since the spec seems to have settled on "middlebox", align the
  terminology. (I don't really care what is used, just that it is
  consistent.)

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1874,7 +1874,7 @@ legacy_session_id
   so a client not offering a pre-TLS 1.3 session MUST generate a
   new 32-byte value. This value need not be random but SHOULD be
   unpredictable to avoid but SHOULD be
-  unpredictable to avoid intermediaries fixating on a specific value
+  unpredictable to avoid implementations fixating on a specific value
   (also known as ossification).
   Otherwise, it MUST be set as a zero length vector (i.e., a single
   zero byte length field).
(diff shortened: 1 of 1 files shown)


davidben:
> Should the legacy_session_id description, which also mentions intermediaries, also be tweaked?

Ah yeah. Fixed. I used "implementations" for this one because there is an "(also known as ossification)" parenthetical and, in the general case, endpoints can cause ossification too. Even here, I dunno, maybe a sufficiently confused endpoint will do something silly there. :-)

draft-ietf-tls-tls13.md
```diff
@@ -1874,7 +1874,7 @@ legacy_session_id
   so a client not offering a pre-TLS 1.3 session MUST generate a
   new 32-byte value. This value need not be random but SHOULD be
   unpredictable to avoid but SHOULD be
-  unpredictable to avoid intermediaries fixating on a specific value
+  unpredictable to avoid implementations fixating on a specific value
```
kaduk: If we're touching this line, maybe we should remove the doubled "but SHOULD be unpredictable to avoid"?
(The actual change itself seems fine.)

davidben: Fixed.

PR #1142, opened 2018-01-18
Clarify behavior of signature_algorithms_cert

ekr:


mattcaswell:
In your answer to my question on list you say:

 > Neither side need send it if it has a consistent policy for CertVerify
and chain validation.

I don't see that covered by this PR. Shouldn't this be explicit?

kaduk:
> If no "signature_algorithms_cert" extension is present,
then the "signature_algorithms" extension also applies to signatures
appearing in certificates.

is unaffected by this change

ekr:
As @kaduk says, there is already text that implies that you need not send "signature_algorithms" cert, so I don't think new text is really needed here.

mattcaswell:
> As @kaduk says, there is already text that implies that you need not send "signature_algorithms" cert, so I don't think new text is really needed here.

That sentence says nothing about under what circumstances it is ok to omit "signature_algorithms_cert". It only implies that you may not always get it. I think it would not hurt to be more explicit.

draft-ietf-tls-tls13.md
```diff
@@ -3518,15 +3519,15 @@ The following rules apply to the certificates sent by the server:
 
 - The certificate MUST allow the key to be used for signing (i.e., the
   digitalSignature bit MUST be set if the Key Usage extension is present) with
-  a signature scheme indicated in the client's "signature_algorithms" extension.
+  a signature scheme indicated in the client's "signature_algorithms"/"signature_algorithms_cert"
+  extensions (see {{signature-algorithms}}).
```
kaduk: If signature_algorithms_cert is supposed to apply to signatures relevant to the X.509 library, then it only matters for the signature *on* the leaf certificate, not signatures made *by* the leaf cert (right?).  So only signature_algorithms is relevant for this purpose.

ekr: No, because this applies to chain certs.

ekr: as well

PR #1143, opened 2018-01-18
Clarify rules around "supported_versions" extension in SH and HRR

roelfdutoit:
Issue 1: supported_versions is required in TLS 1.3 SH.
Issue 2: supported_versions.selected_version must be identical in HRR and SH

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2011,10 +2011,10 @@ legacy_compression_method
 
 extensions
 : A list of extensions.  The ServerHello MUST only include extensions
-  which are required to establish the cryptographic context. Currently
-  the only such extensions are "key_share" and "pre_shared_key".
-  All current TLS 1.3 ServerHello messages will contain one of these
-  two extensions, or both when using a PSK with (EC)DHE key establishment.
+  which are required to establish the cryptographic context and negotiate
+  the protocol version. All TLS 1.3 ServerHello messages MUST contain the
+  "supported_versions" extension, and at least one of either the "pre_shared_key"
+  or "key_share" extensions, or both when using a PSK with (EC)DHE key establishment.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -2011,10 +2011,10 @@ legacy_compression_method
...
 extensions
 : A list of extensions.  The ServerHello MUST only include extensions
-  which are required to establish the cryptographic context. Currently
-  the only such extensions are "key_share" and "pre_shared_key".
-  All current TLS 1.3 ServerHello messages will contain one of these
-  two extensions, or both when using a PSK with (EC)DHE key establishment.
+  which are required to establish the cryptographic context and negotiate
+  the protocol version. All TLS 1.3 ServerHello messages MUST contain the
+  "supported_versions" extension, and at least one of either the "pre_shared_key"
+  or "key_share" extensions, or both when using a PSK with (EC)DHE key establishment.
```
kaduk: We are going from "current" to "all MUST" w.r.t. having at least one of pre_shared_key and key_share, which seems to rule out future modes that use neither.  I'm not sure I want to commit us to that indefinitely.

draft-ietf-tls-tls13.md
```diff
@@ -2131,6 +2131,9 @@ supplied in the ServerHello is the same as that in the
 HelloRetryRequest and otherwise abort the handshake with an
 "illegal_parameter" alert.
 
+The value of selected_version in the HelloRetryRequest "supported_versions"
+extension MUST be retained in ServerHello, and a client MUST abort the
```
kaduk: "the ServerHello"

draft-ietf-tls-tls13.md
roelfdutoit: Updated.

ekr:
@roelfdutoit  please submit a PR to add yourself to acknowledgements

Issue #1145, opened 2018-02-10
Require checking CT in encrypted data


ekr:
We don't require checking the TLSCipherText version and we don't say anything about the CT.

kaduk:
I'm not 100% sure I understand the intent, but I do remember some previous discussions where the conclusion seemed to be that
```
ProtocolVersion legacy_record_version = 0x0303; /* TLS v1.2 */
```
implies that it is permitted for a peer receiving the structure to enforce that value, but not required for it to do so.  And that there were valid reasons why different implementors might want to go one way or the other.

davidben:
Enforcing it on the content type might be worth requiring since it's somewhat semantically meaningful, when doing the trial "decryption" step for 0-RTT reject with HRR.

cjpatton:
The need to enforce checking the values of _opaque_type_ and _legacy_record_version_ arises in 
a stronger-than-usual security model [1]. Allow me to explain.

Let's call the _ciphertext stream_ the sequence of bytes read from the socket. (So we're abstracting the transport layer protocol.). TLS specifies a mechanism for parsing the ciphertext stream into a sequence of ciphertexts. Specifically, each ciphertext is _delimited_ by a string (23, 0x0303, _length_), where _length_ is the number of bytes of the next ciphertext. The process by which the stream is parsed into discrete ciphertexts is usually ignored by cryptographers; the proofs, say, for AES-GCM, say nothing about how this process affects security of the protocol that uses it. Recent papers (e.g. [1,2] ) have sought to bridge this gap by accounting for parsing of the ciphertext stream in the security model. The adversary controls delivery of the ciphertext stream; It may deliver the stream in pieces (bit by bit if it likes) or out-of-order, or it may attempt to forge a ciphertext, or just the delimiter bits, inject a ciphertext, etc.

I recently finished an analysis of the TLS 1.3 record layer (draft 23). The paper is in submission, so I'm not sure how much I can say about it without violating anonymity. But one thing we show is that the precise details of how the stream is parsed into discrete ciphertexts is irrelevant for security _as long as the delimiter bits are treated as associated data for AEAD encryption/decryption_. Otherwise, security crucially depends on the details of this mechanism. This leads us to the following recommendation:
- **Option 1:** Revise Section 5.2 so that the _opaque_type_, _legacy_record_version_, and the _length_ fields of _TLSCiphertext_ are used as associated data by the record protection function.

[...]

PR #1158, opened 2018-02-23
Add the record header to additional data for AEAD (de)protection.

cjpatton:
Currently, checking the legacy_record_version fields of the TLSCiphertext is not mandatory, and checking the opaque_type is unspecified. These values need to be enforced for reasons discussed in issue #1145.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -4215,9 +4215,11 @@ data" to be included in the authentication check, as described in Section 2.1
 of {{RFC5116}}. The key is either the client_write_key or the server_write_key,
 the nonce is derived from the sequence number (see {{nonce}}) and the
 client_write_iv or server_write_iv, and the additional data input is the
-plaintext header associated with the record:
+header associated with the record:
 
-       additional_data = opaque_type | legacy_record_version | length
+       additional_data = TLSCiphertext.opaque_type |
+                         TLSCiphertext.legacy_record_version |
+                         TLSCiphertext.length
 
 The plaintext input to the AEAD algorithm is the encoded TLSInnerPlaintext structure.
 Derivation of traffic keys is defined in {{traffic-key-calculation}}.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -521,6 +521,15 @@ informative:
        -
          ins: J. Somorovsky
 
+  Anon18:
+       title: "Secure Channels for Multiplexed Data Streams: Analyzing the TLS 1.3 Record Layer Without Elision"
+       date: 2018
+       seriesinfo: "In sumbission to CRYPTO 2018. RFC EDITOR: PLEASE UPDATE THIS
+       REFERENCE AFTER FINAL NOTIFICATION (2018-4-29)."
+       author:
+       -
+         ins: A. Anonymous
+
 --- abstract
 
 This document specifies version 1.3 of the Transport Layer Security
@@ -4185,8 +4194,6 @@ legacy_record_version
 : The legacy_record_version field is always 0x0303.  TLS 1.3 TLSCiphertexts
   are not generated until after TLS 1.3 has been negotiated, so there are
   no historical compatibility concerns where other values might be received.
-  Implementations MAY verify that the legacy_record_version field is 0x0303
-  and abort the connection if it is not.
   Note that the handshake protocol including the ClientHello and ServerHello
   messages authenticates the protocol version, so this value is redundant.
 
@@ -4207,36 +4214,40 @@ AEAD algorithms take as input a single key, a nonce, a plaintext, and "additiona
 data" to be included in the authentication check, as described in Section 2.1
 of {{RFC5116}}. The key is either the client_write_key or the server_write_key,
 the nonce is derived from the sequence number (see {{nonce}}) and the
-client_write_iv or server_write_iv, and the additional data input is empty
-(zero length).  Derivation of traffic keys is defined in {{traffic-key-calculation}}.
+client_write_iv or server_write_iv, and the additional data input is the
+plaintext header associated with the record:
+
+       additional_data = opaque_type | legacy_record_version | length
 
 The plaintext input to the AEAD algorithm is the encoded TLSInnerPlaintext structure.
+Derivation of traffic keys is defined in {{traffic-key-calculation}}.
 
 The AEAD output consists of the ciphertext output from the AEAD
 encryption operation. The length of the plaintext is greater than the
 corresponding TLSPlaintext.length due to the inclusion of TLSInnerPlaintext.type and
-any padding supplied by the sender.  The length of the
+any padding supplied by the sender. The length of the
 AEAD output will generally be larger than the plaintext, but by an
 amount that varies with the AEAD algorithm. Since the ciphers might
 incorporate padding, the amount of overhead could vary with different
 lengths of plaintext. Symbolically,
 
        AEADEncrypted =
-           AEAD-Encrypt(write_key, nonce, plaintext)
+           AEAD-Encrypt(write_key, nonce, additional_data, plaintext)
 
-In order to decrypt and verify, the cipher takes as input the key,
-nonce, and the AEADEncrypted value. The output is either the plaintext
-or an error indicating that the decryption failed. There is no
-separate integrity check. That is:
+Then the encrypted_record field of TLSCiphertext is set to AEADEncrypted.
+In order to decrypt and verify, the cipher takes as input the key, nonce,
+additional data, and the AEADEncrypted value. The output is either the plaintext
+or an error indicating that the decryption failed. There is no separate
+integrity check. That is:
 
        plaintext of encrypted_record =
-           AEAD-Decrypt(peer_write_key, nonce, AEADEncrypted)
+           AEAD-Decrypt(peer_write_key, nonce, additional_data, AEADEncrypted)
 
 If the decryption fails, the receiver MUST terminate the connection
 with a "bad_record_mac" alert.
 
 An AEAD algorithm used in TLS 1.3 MUST NOT produce an expansion greater than
-255 octets.  An endpoint that receives a record from its peer with
+255 octets. An endpoint that receives a record from its peer with
 TLSCiphertext.length larger than 2^14 + 256 octets MUST terminate
 the connection with a "record_overflow" alert.
 This limit is derived from the maximum TLSPlaintext length of


richsalz:
This would change the wire protocol yet again.  Is it really necessary or just a nice-to-have?

cjpatton:
At a minimum, the values of opaque_type and legacy_record_version need to be enforced.

bifurcation:
@richsalz This would not change the wire protocol, just the logic for encapsulating / decapsulating records.

bifurcation:
I see what you mean.  It wouldn't change the header octets, but it would change the AEAD output (e.g., the authentication tag).

richsalz:
So if we follow practice, this is yet another incompatible change and different draft-id.  Or we wait and roll this up into the post-LC draft.  Or what?  What is bad if we don't do this, and just say it's too late in the process?

cjpatton:
@richsalz, can you explain what you mean by "incompatible change"?

richsalz:
Will a server that implements this be able to properly decode packets sent by a client that doesn't, or vice-versa?

kaduk:
The idea is that it would not, to make enforcing the header contents a mandatory part of protocol implementation.

richsalz:
Which is my point; this is an *incompatible change.*  How important is it?  How we will handle it in terms of marking the draft in an extension, coordinating with browsers, etc?

ekr:
Rich, I worry people are talking past each other here. Let me try to state what I believe to be the case.

Right now, people are deploying TLS 1.3-23. If a client supports TLS 1.3-23 and a server supports TLS 1.3-24, they will not be able to negotiate TLS 1.3, but will successfully be able to negotiate TLS 1.2 if both support that. Similarly, if servers deploy TLS 1.3 (RFC) they will not be able to talk to TLS 1.3-23 clients (and vice versa), but will be able to negotiate TLS 1.2. This is true even if nothing else in the drafts changes but the version numbers (which is nearly the case for 23->24).

The same thing will continue to be true  if we make this change in -25 or the RFC version.

kaduk:
If we make this change we'd need to issue a new draft to do interop testing usefully.

ekr:
On Fri, Feb 23, 2018 at 3:33 PM, kaduk <notifications@github.com> wrote:

> If we make this change we'd need to issue a new draft to do interop
> testing usefully.
>

Yes, we would need to issue a new draft # to do interop.

richsalz:
And now, finally, we're all on the same page and I can repeat my initial question: How important is it that we do this?

ekr:
I don't think it's essential, but as I said above, it's not hard to implement, seems to add some analytical value, and will make reasoning about the various DTLS header variants much easier.

richsalz:
Is there enough time to do a draft re-spin and actually get people to test?  It seems to me that the cycle is four weeks, maybe three if we're lucky.  Is anyone signing up to do that?  Else we have to decide if it goes into the RFC or if it's worth delaying publication by having another cycle.

ekr:
If the consensus is to do this, we will be able to have an implementation in a week or two. Note that the IESG decision date is 2 weeks out and the RFC queue has a certain length to it.

davidben:
> some analytical value

I think it is worth being clear on what that value is. From the earlier bug, the "value" is that it authenticates the (trivially information-less) TLS-level framing layer, which is useful if one's security model is "the (trivially information-less) TLS-level framing layer is authenticated, but it is okay if the transport-level (TCP) framing layer is not", yes? This a little circular.

Incorporating the length is especially fiddly as there are two lengths, the plaintext and the ciphertext length. The names of the variables do imply the latter (just called `length` where), but I am sure every implementer on the planet will first mess up and encrypt with plaintext length and decrypt with ciphertext length, as those as the lengths which are naturally available at those stages. It will be caught with any interop test, but if the gains here are so trivial, why bother?

kaduk:
Yeah, the only real motivation I see to do this is if it makes things easier for DTLS 1.3, which I haven't gotten a chance to think about very hard.  Since we're going to try to shave some extra bytes off the header it might be actually useful, though perhaps it requires there to be multiple accepted variant headers in order to achieve "actually useful" status.

cjpatton:
First a bit of terminology: **authenticating** the record header means to use it as AD for (de)protection (i.e. this PR); **enforcing** the record header means to simply check that the received _opaque_type_ and _legacy_record_version_ are correct.

> From the earlier bug, the "value" is that it authenticates the (trivially information-less) TLS-level framing layer, which is useful if one's security model is "the (trivially information-less) TLS-level framing layer is authenticated, but it is okay if the transport-level (TCP) framing layer is not", yes? This a little circular.

In the intended model (see #1158 for details), the channel is under attack if the sequence of bits consumed differs from the sequence of bits produced. Thus, fiddling with TCP (even just the headers) will at worst amount to an attack. To detect it, we must authenticate (or at least enforce the values of) _all_ of the consumed bits, and not just the ciphertext. All this to say: I don't see how this is circular, since authenticating (or enforcing) the TLS header suffices to detect any attack on TCP that is relevant to the security goals of the channel.

**Is the model too strong?** I don't think so. For one, it's security conservative. Not enforcing the record header values permits compliant implementations that use these values in unintended ways. For example, an implementation might receive _opaque_type_ = 21 (alert) and immediately close the channel. This may seem farfetched, particularly to folks who are actually working on deploying TLS 1.3. But I think it's important to be precise about this kind of stuff, and currently, the draft is not precise about the record header semantics.

**Why authenticate?** Enforcing the record header values is sufficient for security. One reason I favor authentication is that it reduces the complexity of a formal analysis: _If all of the bits produced by the channel sender are authenticated, then one can prove that the details of the record header's format are irrelevant to security_. This also permits a certain amount of flexibility for the spec. For example, if in a later version of TLS, the value of _opaque_type_ is changed to 89, then this does not break interoperability with TLS 1.3. To take a more useful example, if the maximum record length is changed to 2^17 bytes, then we will need more than 16 bits to encode the _length_ field; this can be changed in the spec without impacting security.

I appreciate the fact that making this change will require effort. From where I stand, authentication is the better way forward, and it doesn't much matter to me if this change gets adopted now or later. That said,  if the consensus is to enforce rather than authenticate, I will open a new PR and close this one.

tomato42:
>  but I am sure every implementer on the planet will first mess up and encrypt with plaintext length and decrypt with ciphertext length, as those as the lengths which are naturally available at those stages. It will be caught with any interop test, but if the gains here are so trivial, why bother?

exactly this, and all AEADs we have in TLS 1.3 (AES-GCM, AES-CCM, Chacha20/Poly1305) feed ciphertext length (sans tag) to the authentication tag internally, so adding it to Authenticated Data is redundant

so the only thing that the attacker can change are the version and type, which either _do not affect the state of the system under attack_ or will cause connection abort.

the change should be limited to just **enforcing** the record header for protected records as that will not cause new interoperability pains and can be verified with no more complex testing than the **authenticate** option

cjpatton:
> Incorporating the length is especially fiddly as there are two lengths, the plaintext and the ciphertext length. The names of the variables do imply the latter (just called length here), but I am sure every implementer on the planet will first mess up and encrypt with plaintext length and decrypt with ciphertext length, as those as the lengths which are naturally available at those stages.

Indeed, there are actually three lengths the implementor needs to worry about: the unpadded plaintext, the plaintext, and the ciphertext. The spec needs to clear about what length is intended but I this seems easy to do.

cjpatton:
>  all AEADs we have in TLS 1.3 (AES-GCM, AES-CCM, Chacha20/Poly1305) feed ciphertext length (sans tag) to the authentication tag internally, so adding it to Authenticated Data is redundant

This point was brought up before, and it's well taken. I would only emphasize that not every AEAD scheme incorporates the length into authentication; sometimes it's needed for the proof, but not always. From an analytical perspective, it's useful to not have to make assumptions about the structure of your primitives: in this case, your AEAD scheme.

I think the more important point (and one that @ekr has made) is that authenticating _length_ is simply not needed. If this value is mangled in transit, then it necessarily changes the inputs to the deprotection function, so the attack will be detected. 

The reason to authenticate the _length_ field along with the others is that authenticating the whole header reduces the complexity of the analysis.

tomato42:
>  doing so means we needn't worry about these weird edge-case implementations.

we don't have to worry about them at all, they aren't implementing _TLS_ if they ignore length tag

> However, this logic is somewhat brittle, since for it to be true, the receiver needs to precisely interpret length as the length (in bytes) of the next ciphertext.

and if it doesn't do that, it'll interpret ciphertext as authentication tag which will almost certainly cause the tag verification to fail, and even if attacker picks that ciphertext block carefully, how is that different from an attack that just modifies the authentication tag directly?

ekr:
In terms of what was ADed I would proposes that the rule would simply be "everything up to the plaintexT" as it appears on the wire. That is straightforward to understand and implement, especially on the receiver. 

The one complexity from the implementation perspective is on the protecting side, you need to know the AEAD output length in advance. However, with all our existing AEAD algorithms, the expansion is constant and so this is straightforward, and of course it was required to be able to do this for TLS 1.2 and below anyway.

cjpatton:
In fact, determining the ciphertext length is straight-forward for any RFC5116-compliant AEAD scheme, right? From my read of the RFC, it seems to require that the length of the ciphertext be easily computable from the length of the plaintext. This is fairly mild condition.

ekr:
On Mon, Feb 26, 2018 at 6:21 AM, Christopher Patton <
notifications@github.com> wrote:

> In fact, determining the ciphertext length is straight-forward for any
> RFC5116-compliant AEAD scheme, right? From my read of the RFC, it seems to
> require that the length of the ciphertext be easily computable from the
> length of the plaintext.
>

I assume you are referring to:

   Each AEAD algorithm MUST provide a description relating the length of
   the plaintext to that of the ciphertext.  This relation MUST NOT
   depend on external parameters, such as an authentication strength
   parameter (e.g., authentication tag length).  That sort of dependence
   would complicate the use of the algorithm by creating a situation in
   which the information from the AEAD registry was not sufficient to
   ensure interoperability.

That does seem dispositive.

-Ekr

kaduk:
Actually getting around to looking at the dtls spec, the "interesting" part comes in from having both DTLSCiphertext and DTLSShortCiphertext as possibilities (mostly discounting DTLSPlaintext for now).  De-muxing is done based on the first byte ("ContentType", nominally, in the historical order).  The non-short version also has the opaque content type and length fields that we've been discussing already, and both have a subset of the full epoch and sequence number used for protection.  Now, the epoch+sequence are already used as input to the AEAD, either for key selection or as part of the IV, but the bits on the wire are not the full epoch+sequence in either case.  It would be hard to not enforce the headers for the short DTLS ciphertext and still interoperate, but I guess it would be pretty easy to ignore the content type and length from the long DTLS header.  So, I guess there is some marginal benefit from AAD-ing the DTLS header(s), which puts me barely over the edge of thinking we should do this, but it remains a pretty close call to me.

(I think the DTLS draft should also ask IANA to add a note to the ContentType 0 (reserved) entry that the 0 value is used for de-muxing the DTLS short ciphertext, but that's a matter for elsewhere.)

martinthomson:
Does [draft-ietf-tls-dtls-connection-id](https://tools.ietf.org/html/draft-ietf-tls-dtls-connection-id) change this any?

kaduk:
> Does draft-ietf-tls-dtls-connection-id change this any?

I suppose it does, yes.  (It makes the case for authenticating the headers stronger.)

ekr:
Obsoleted by #1162

PR #1159, opened 2018-02-28
Addressing IANA questions

seanturner:
During the IANA review some questions were raised and this addresses them.

One remaining question needs to be resolved:

rsa_pss_sha256, rsa_pss_sha384, and rsa_pss_sha512 are listed in the SignatureScheme registry text in the IANA consideration, but we changed that extension so now the values need to be chosen from:
rsa_pss_rsae_sha256
rsa_pss_rsae_sha384
rsa_pss_rsae_sha512
rsa_pss_pss_sha256
rsa_pss_pss_sha384
rsa_pss_pss_sha512

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5289,12 +5289,14 @@ The registries and their allocation policies are below:
   are marked as "Yes" for each of these additions.
 
 This document also uses the TLS ExtensionType Registry originally created in
-{{RFC4366}}. IANA has updated it to reference this document. The registry and
-its allocation policy is listed below:
+{{RFC4366}}. IANA has updated it to reference this document.  Changes to the
+registry follow:
 
- - IANA \[SHALL update/has updated] Values with the first byte in the range
-   0-254 (decimal) are assigned via Specification Required [RFC8126].  Values
-   with the first byte 255 (decimal) are reserved for Private Use [RFC8126].
+ - IANA \[SHALL update/has updated] this registration policy as follows:
+
+   Values with the first byte in the range 0-254 (decimal) are assigned
+   via Specification Required [RFC8126].  Values with the first byte
+   255 (decimal) are reserved for Private Use [RFC8126].
 
 - IANA \[SHALL update/has updated] this registry to include the
   "key_share", "pre_shared_key", "psk_key_exchange_modes",
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -5324,7 +5324,7 @@ by IANA:
   The registry \[shall be/ has been] initially populated with the values described in
   {{signature-algorithms}}. The following values SHALL be marked as
   "Recommended": ecdsa_secp256r1_sha256, ecdsa_secp384r1_sha384,
-  rsa_pss_sha256, rsa_pss_sha384, rsa_pss_sha512, ed25519.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -5278,18 +5278,24 @@ The registries and their allocation policies are below:
...
+  to include values for "missing_extension" and "certificate_required".  The
+  "DTLS-OK" column is marked as "Yes" for each new alert.
 
 - TLS HandshakeType Registry: Future values are allocated via
   Standards Action {{RFC8126}}. IANA \[SHALL update/has updated] this registry
   to rename item 4 from "NewSessionTicket" to "new_session_ticket"
   and to add the "hello_retry_request_RESERVED", "encrypted_extensions",
-  "end_of_early_data", "key_update", and "message_hash" values.
+  "end_of_early_data", "key_update", and "message_hash" values.  The "DTLS-OK"
+  are marked as "Yes" for each of these additions.
```
kaduk: I guess we were maybe talking about removing key_update for DTLS in favor of just using the epoch, but presumably DLTS 1.3 could update the registry as needed.
Though, hmm, since DTLS 1.3 doesn't actually exist yet and these are 1.3-only messages, maybe it makes sense to not mark them okay until DTLS 1.3 exists, i.e., in the DTLS 1.3 IANA considerations.

draft-ietf-tls-tls13.md
```diff
@@ -5278,18 +5278,24 @@ The registries and their allocation policies are below:
...
 
 This document also uses the TLS ExtensionType Registry originally created in
 {{RFC4366}}. IANA has updated it to reference this document. The registry and
 its allocation policy is listed below:
 
+ - IANA \[SHALL update/has updated] Values with the first byte in the range
```
kaduk: Is this shall/has spurious so that we should just start with "Values"?

draft-ietf-tls-tls13.md
martinthomson: The current plan is to use key_update in DTLS.  There are a few reasons why maintaining some degree of parity with TLS is nice here.  Asymmetric updates are one.

kaduk:
Do we also want to add text about the location of the SignatureScheme registry?

draft-ietf-tls-tls13.md
seanturner: No I think it's just a bad cut/paste job.  I'll fix shortly.

seanturner:
I didn't include text about where to put SignatureScheme because I'm not quite sure how to say put it here.  I'm also less worried about having a trail of crumbs for instructions on where to put it; I also think IANA will place it based on our email.

draft-ietf-tls-tls13.md
```diff
@@ -5278,17 +5278,25 @@ The registries and their allocation policies are below:
...
-{{RFC4366}}. IANA has updated it to reference this document. The registry and
-its allocation policy is listed below:
+{{RFC4366}}. IANA has updated it to reference this document.  Changes to the
+registry follow:
+
+ - IANA \[SHALL update/has updated] the registration policy as follows:
+
+   Values with the first byte in the range 0-254 (decimal) are assigned
+   via Specification Required [RFC8126].  Values with the first byte
+   255 (decimal) are reserved for Private Use [RFC8126].
```
ekr: Doesn't this belong in the IANA draft?

ekr:
Merged by hand

PR #1162, opened 2018-03-02
Additional data

ekr:
Previous PR plus some cleanup.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -643,15 +643,18 @@ RFC EDITOR PLEASE DELETE THIS SECTION.
 (*) indicates changes to the wire protocol which may require implementations
     to update.
 
+draft-25
+- Add the header to additional data (*)
+
+- Minor clarifications.
+
+- IANA cleanup.
+
 draft-24
 
 - Require that CH2 have version 0303 (*)
 
 - Some clarifications
-
-
-draft-24
-- 
 
 draft-23
 - Renumber key_share (*)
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -4212,36 +4225,42 @@ AEAD algorithms take as input a single key, a nonce, a plaintext, and "additiona
...
        AEADEncrypted =
-           AEAD-Encrypt(write_key, nonce, plaintext)
+           AEAD-Encrypt(write_key, nonce, additional_data, plaintext)
 
-In order to decrypt and verify, the cipher takes as input the key,
-nonce, and the AEADEncrypted value. The output is either the plaintext
-or an error indicating that the decryption failed. There is no
-separate integrity check. That is:
+Then the encrypted_record field of TLSCiphertext is set to AEADEncrypted.
+In order to decrypt and verify, the cipher takes as input the key, nonce,
```
kaduk: Probably want a paragraph break here.

cjpatton:
Looks good to me, for what it's worth :)

seanturner:
ekr: please merge this so we can get a new version out for the IESG.

ekr:
@kaduk: can you be more precise about the appendix D text? is the issue that we say "ignored" but then use it as additional_data?

kaduk:
```
The value of
   TLSCiphertext.legacy_record_version MAY be ignored, or MAY be
   validated to match the fixed constant value.
```

draft-ietf-tls-tls13.md
```diff
@@ -5677,7 +5678,8 @@ Prior versions of TLS used the record layer version number for various
 purposes. (TLSPlaintext.legacy_record_version and TLSCiphertext.legacy_record_version)
 As of TLS 1.3, this field is deprecated. The value of
 TLSPlaintext.legacy_record_version MUST be ignored by all implementations.
-The value of TLSCiphertext.legacy_record_version MAY be ignored,
+The value of TLSCiphertext.legacy_record_version is included in the
+additional data for deprotection but MAY otherwise be ignored
 or MAY be validated to match the fixed constant value.
```
kaduk: I guess this works, since just being included in the AAD does not necessarily imply that the sender used the right values.

ekr: That was my reasoning.

PR #1166, opened 2018-03-06
Notes from kaduk's IESG review

kaduk:
A number of things I noticed while doing my "IESG" review.

Some of these may be a little controversial; I'll try to note them in inline comments and why I wanted to make them.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1426,8 +1426,8 @@ Note:
   low-entropy sources is not secure. A low-entropy secret, or password, is
   subject to dictionary attacks based on the PSK binder.  The specified PSK
   authentication is not a strong password-based authenticated key exchange even
-  when used with Diffie-Hellman key establishment.  That is, it does not provide
-  resistance againt an attacker that can observe the handshake performing
+  when used with Diffie-Hellman key establishment.  That is, it does not
+  prevent an attacker that can observe the handshake from performing
   a brute-force attack on the password/pre-shared key.
 
 
@@ -2336,7 +2336,7 @@ this specification MUST negotiate TLS 1.2 or prior as specified in
 Servers MAY abort the handshake upon receiving a ClientHello with
 legacy_version 0x0304 or later.
 
-If this extension is present in the ClientHello, servers MUST NOT the
+If this extension is present in the ClientHello, servers MUST NOT use the
 ClientHello.legacy_version value for version negotiation and MUST use only the
 "supported_versions" extension to determine client
 preferences. Servers MUST only select a version of TLS present in that
@@ -4577,7 +4577,7 @@ bad_record_mac
 record_overflow
 : A TLSCiphertext record was received that had a length more than
   2^14 + 256 bytes, or a record decrypted to a TLSPlaintext record
-  with more than 2^14 bytes (or some other configured limit).
+  with more than 2^14 bytes (or some other negotiated limit).
   This alert should never be observed in communication between
   proper implementations, except when messages were corrupted
   in the network.
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -1659,8 +1659,7 @@ like that of C.
        } T;
 
 Fixed- and variable-length vector fields are allowed using the standard vector
-syntax (but arrays of either sort must themselves only contain fixed-length
-elements). Structures V1 and V2 in the variants example below demonstrate this.
+syntax. Structures V1 and V2 in the variants example below demonstrate this.
 
 The fields within a structure may be qualified using the type's name, with a
 syntax much like that available for enumerateds. For example, T.f2 refers to
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -2649,9 +2649,7 @@ filters
   some of the required certificate extension OIDs and supplied a certificate
   that does not satisfy the request, the server MAY at its discretion either
   continue the connection without client authentication, or abort the handshake
-  with an "unsupported_certificate" alert.  The client MAY abort the handshake
-  if a given certificate_extension_oid value appears multiple times in the
-  filters list.
+  with an "unsupported_certificate" alert.
 
 PKIX RFCs define a variety of certificate extension OIDs and their corresponding
 value types. Depending on the type, matching certificate extension values are
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -604,7 +604,7 @@ mechanism defined in {{RFC5077}} and replaces it with the mechanism
 defined in {{resumption-and-psk}}. {{negotiated-groups}} updates
 {{RFC4492}} by modifying the protocol attributes used to negotiate
 Elliptic Curves.  Because TLS 1.3 changes the way keys are derived it
-updates {{RFC5705}} as described in {{exporters}} it also changes
+updates {{RFC5705}} as described in {{exporters}}.  It also changes
 how OCSP messages are carried and therefore updates {{RFC6066}}
 and obsoletes {{RFC6961}} as described in section {{ocsp-and-sct}}.
 
@@ -624,7 +624,7 @@ connection: A transport-layer connection between two endpoints.
 
 endpoint: Either the client or server of the connection.
 
-handshake: An initial negotiation between client and server that establishes the parameters of their subsequent interactions.
+handshake: An initial negotiation between client and server that establishes the parameters of their subsequent interactions within TLS.
 
 peer: An endpoint. When discussing a particular endpoint, "peer" refers to the endpoint that is not the primary subject of discussion.
 
@@ -644,9 +644,11 @@ RFC EDITOR PLEASE DELETE THIS SECTION.
     to update.
 
 draft-26
+
 - Clarify that you can't negotiate pre-TLS 1.3 with supported_versions.
 
 draft-25
+
 - Add the header to additional data (*)
 
 - Minor clarifications.
@@ -660,6 +662,7 @@ draft-24
 - Some clarifications
 
 draft-23
+
 - Renumber key_share (*)
 
 - Add a new extension and new code points to allow negotiating PSS
@@ -673,6 +676,7 @@ draft-23
 - Add some text on the security of static RSA.
 
 draft-22
+
 - Implement changes for improved middlebox penetration (*)
 
 - Move server_certificate_type to encrypted extensions (*)
@@ -1102,19 +1106,19 @@ are many minor differences.
 
 - The handshake state machine has been significantly restructured to
   be more consistent and to remove superfluous messages such as
-  ChangeCipherSpec.
+  ChangeCipherSpec (except when needed for middlebox compatibility).
 
-- Elliptic curve algorithms are now in the base spec and includes new signature
-  algorithms, such as ed25519 and ed448. TLS 1.3 removed point format
+- Elliptic curve algorithms are now in the base specification and new signature
+  algorithms are included, such as ed25519 and ed448. TLS 1.3 removed point format
   negotiation in favor of a single point format for each curve.
 
 - Other cryptographic improvements including the removal of compression and
-  custom DHE groups, changing the RSA padding to use PSS, and the removal of
+  custom DHE groups, changing the RSA padding to use RSASSA-PSS, and the removal of
   DSA.
 
 - The TLS 1.2 version negotiation mechanism has been deprecated in favor
   of a version list in an extension. This increases compatibility with
-  servers which incorrectly implemented version negotiation.
+  existing servers that incorrectly implemented version negotiation.
 
 - Session resumption with and without server-side state as well as the
   PSK-based ciphersuites of earlier TLS versions have been replaced by a
@@ -1136,6 +1140,9 @@ TLS 1.2:
 * The "supported_versions" ClientHello extension can be used to negotiate
   the version of TLS to use, in preference to the legacy_version field of
   the ClientHello.
+
+* The "signature_algorithms_cert" extension allows a client to indicate
+  which signature algorithms it can validate in X.509 certificates
 
 An implementation of TLS 1.3 that also supports TLS 1.2 might need to include
 changes to support these changes even when TLS 1.3 is not in use.  See the
@@ -1225,7 +1232,8 @@ In the Key Exchange phase, the client sends the ClientHello
 symmetric cipher/HKDF hash pairs; either a set of Diffie-Hellman key shares (in the
 "key_share" extension {{key-share}}), a set of pre-shared key labels (in the
 "pre_shared_key" extension {{pre-shared-key-extension}}) or both; and
-potentially additional extensions.
+potentially additional extensions.  Additional fields and/or messages
+may also be present for middlebox compatibility.
 
 The server processes the ClientHello and determines the appropriate
 cryptographic parameters for the connection. It then responds with its
@@ -1254,7 +1262,9 @@ CertificateRequest:
   omitted if client authentication is not desired. \[{{certificate-request}}]
 
 Finally, the client and server exchange Authentication messages. TLS
-uses the same set of messages every time that authentication is needed.
+uses the same set of messages every time that certificate-based
+authentication is needed.  (PSK-based authentication happens as a side
+effect of key exchange.)
 Specifically:
 
 Certificate:
@@ -1288,7 +1298,8 @@ At this point, the handshake is complete, and the client and server
 derive the keying material required by the record layer to exchange
 application-layer data protected through authenticated encryption.
 Application data MUST NOT be sent prior to sending the Finished message and
-until the record layer starts using encryption keys.
+until the record layer starts using encryption keys, except as specified
+in \[{{zero-rtt-data}}].
 Note that while the server may send application data prior to receiving
 the client's Authentication messages, any data sent at that point is,
 of course, being sent to an unauthenticated peer.
@@ -1329,7 +1340,7 @@ the server MUST abort the handshake with an appropriate alert.
 ~~~
 {: #tls-restart title="Message flow for a full handshake with mismatched parameters"}
 
-Note: The handshake transcript includes the initial
+Note: The handshake transcript incorporates the initial
 ClientHello/HelloRetryRequest exchange; it is not reset with the new
 ClientHello.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -1288,7 +1298,8 @@ At this point, the handshake is complete, and the client and server
 derive the keying material required by the record layer to exchange
 application-layer data protected through authenticated encryption.
 Application data MUST NOT be sent prior to sending the Finished message and
-until the record layer starts using encryption keys.
+until the record layer starts using encryption keys, except as specified
+in \[{{zero-rtt-data}}].
```
kaduk: I think we've talked about this statement before, so feel free to tell me that it's already decided to keep it this way -- I don't have any new arguments to add.

draft-ietf-tls-tls13.md
```diff
@@ -1646,7 +1659,8 @@ like that of C.
        } T;
 
 Fixed- and variable-length vector fields are allowed using the standard vector
-syntax. Structures V1 and V2 in the variants example below demonstrate this.
+syntax (but arrays of either sort must themselves only contain fixed-length
+elements). Structures V1 and V2 in the variants example below demonstrate this.
```
kaduk: Not entirely clear that we actually need to say this, since we do call it an informal presentation language.  But I believe we need a fixed per-element size in order to do arrays, and don't think we say that explicitly anywhere.

draft-ietf-tls-tls13.md
```diff
@@ -1860,6 +1874,9 @@ ClientHello (without modification) except:
 - Optionally adding, removing, or changing the length of the "padding"
   extension {{RFC7685}}.
 
+- Other modifications that may be allowed by an extension defined in the
+  future and present in the HelloRetryRequest
```
kaduk: This is a potentially big change, letting future extensions modify ClientHello2 without a formal Update to the core spec.

draft-ietf-tls-tls13.md
```diff
@@ -2528,8 +2545,8 @@ Legacy algorithms
   in this context with either with RSA using RSASSA-PKCS1-v1_5 or ECDSA.  These values
   refer solely to signatures which appear in certificates (see
   {{server-certificate-selection}}) and are not defined for use in
-  signed TLS handshake messages, even if they appear in the "signature_algorithms"
-  list (this is necessary for backward compatibility with TLS 1.2).
+  signed TLS handshake messages, although they MAY appear in "signature_algorithms"
+  and "signature_algorithms_cert" for backward compatibility with TLS 1.2,
```
kaduk: This is just normalizing to match the language used for PKCS#1 and is not intended to be a behavior change

draft-ietf-tls-tls13.md
```diff
@@ -2633,7 +2650,9 @@ filters
   some of the required certificate extension OIDs and supplied a certificate
   that does not satisfy the request, the server MAY at its discretion either
   continue the connection without client authentication, or abort the handshake
-  with an "unsupported_certificate" alert.
+  with an "unsupported_certificate" alert.  The client MAY abort the handshake
+  if a given certificate_extension_oid value appears multiple times in the
+  filters list.
```
kaduk: This is a new restriction, but I think it matches the intent.

draft-ietf-tls-tls13.md
```diff
@@ -2941,9 +2962,14 @@ the client and servers MUST supply "key_share" values as described
...
 
+Any future values that are allocated must ensure that the transmitted
+protocol messages unambiguously identify which mode was selected by
+the server; at present, this is indicated by the presence of the "key_share"
+in the ServerHello.
+
 ### Early Data Indication
 
-When a PSK is used, the client can send application data
+When a PSK is used and early data is allowed for that PSK, the client can send application data
```
kaduk: This is somewhat awkward language, but I do think we want to note that not all PSKs allow this.

draft-ietf-tls-tls13.md
```diff
@@ -2976,7 +3002,7 @@ MUST be the first PSK listed in the client's "pre_shared_key" extension.
 For PSKs provisioned via NewSessionTicket, a server MUST validate that
 the ticket age for the selected PSK identity (computed by subtracting
 ticket_age_add from PskIdentity.obfuscated_ticket_age modulo 2^32)
-is within a small tolerance of the
+is within a small tolerance (a small number of seconds) of the
```
kaduk: Is it better for us to make up numbers or implementers to make up numbers?

draft-ietf-tls-tls13.md
```diff
@@ -3003,7 +3029,7 @@ MUST behave in one of three ways:
   HelloRetryRequest.  A client MUST NOT include the "early_data" extension in
   its followup ClientHello.  The server then ignores early data by skipping
   all records with external content type of "application_data" (indicating
-  that they are encrypted).
+  that they are encrypted), up to the configured max_early_data_size.
```
kaduk: (Parallelism with the previous item)

draft-ietf-tls-tls13.md
```diff
@@ -3356,7 +3384,7 @@ Based on these inputs, the messages then contain:
 Certificate
 : The certificate to be used for authentication, and any
   supporting certificates in the chain. Note that certificate-based
-  client authentication is not available in 0-RTT mode.
+  client authentication is not available in resumption (or 0-RTT) flows.
```
kaduk: Please double-check this, but the pictures don't show a server Certificate for resumption.

draft-ietf-tls-tls13.md
```diff
@@ -4643,11 +4672,6 @@ bad_certificate_status_response
   provided by the server via the "status_request" extension
   (see {{RFC6066}}).
 
-bad_certificate_hash_value
-: Sent by servers when a retrieved object does not have the correct hash
-  provided by the client via the "client_certificate_url" extension
-  (see {{RFC6066}}).
```
kaduk: These alerts are only usable for client_certificate_url, which we deemed to be silly and removed a while ago.

draft-ietf-tls-tls13.md
```diff
@@ -4936,9 +4960,7 @@ Where Secret is either the early_exporter_master_secret or the
 exporter_master_secret.  Implementations MUST use the exporter_master_secret unless
 explicitly specified by the application. The early_exporter_master_secret is
 defined for use in settings where an exporter is needed for 0-RTT data.
-A separate interface for the early exporter is RECOMMENDED, especially
-on a server where a single interface can make the early exporter
-inaccessible.
+A separate interface for the early exporter is RECOMMENDED.
```
kaduk: I don't understand what the removed clause was trying to say.

draft-ietf-tls-tls13.md
```diff
@@ -5110,7 +5132,7 @@ ClientHello, otherwise falling back to a 1-RTT handshake.
 This is necessary for the ClientHello storage mechanism
 described in {{client-hello-recording}} because otherwise the server
 needs to store an unlimited number of ClientHellos and is a useful optimization for
-single-use tickets because it allows efficient rejection of ClientHellos
+self-contained single-use tickets because it allows efficient rejection of ClientHellos
```
kaduk: tickets that are lookup keys to a database will need to be looked up before they could be used, and if you can't find it, you know it's not usable at all, let alone for 0-RTT.

draft-ietf-tls-tls13.md
```diff
@@ -5204,7 +5226,7 @@ applicable features:
 
 A client is considered to be attempting to negotiate using this
 specification if the ClientHello contains a "supported_versions"
-extension with 0x0304 as the highest version number contained in its body.
+extension with 0x0304 contained in its body.
```
kaduk: This one I could go either way on, but it seems that if you're going to allow 1.3 to be negotiated, you still need to provide what 1.3 will need.

draft-ietf-tls-tls13.md
martinthomson: This is not the case.  You can have an array of variable-length elements.  Those elements necessarily include length prefixes.  What you can't have is arrays of things that do not have a means by which the length is known.

See for example from ClientHello:
```
           Extension extensions<8..2^16-1>;
```

draft-ietf-tls-tls13.md
martinthomson: I think that this is reasonable.  We do this for key_share.

Nit: period.

draft-ietf-tls-tls13.md
martinthomson: Seems weak.  Is there a prohibition on repeated values? Should there be?  If so, then this is an alertable offense.  If there are reasons to allow duplicate values, then we don't need this.

draft-ietf-tls-tls13.md
martinthomson: +1

draft-ietf-tls-tls13.md
martinthomson: Implementers will manage.  FWIW, NSS kicks the can even further down the road.

draft-ietf-tls-tls13.md
martinthomson: Yes.  This is correct.  "PSK" rather than resumption perhaps?

draft-ietf-tls-tls13.md
martinthomson: Oh, if a server accepts a ClientHello and sends its flight as part of processing that, then it has the full exporter immediately available.  A single exporter interface would return the full exporter.

draft-ietf-tls-tls13.md
kaduk: Wow, massive braino, sorry about that.  Presumably I was thinking about C arrays or something

draft-ietf-tls-tls13.md
kaduk: ack.

draft-ietf-tls-tls13.md
kaduk: I don't have a great understanding of how the filters actually work, myself, so I can only say that I *think* there is not a reason to allow duplicate values.  I also had in my head that this was going to be MUST when I first noticed it, so I'll go ahead and make the MAY-->MUST change for now -- thanks for pointing it out.

draft-ietf-tls-tls13.md
kaduk: Okay.

draft-ietf-tls-tls13.md
kaduk: Ah. I will propose a rewording.

draft-ietf-tls-tls13.md
```diff
@@ -1415,7 +1426,9 @@ Note:
   low-entropy sources is not secure. A low-entropy secret, or password, is
   subject to dictionary attacks based on the PSK binder.  The specified PSK
   authentication is not a strong password-based authenticated key exchange even
-  when used with Diffie-Hellman key establishment.
+  when used with Diffie-Hellman key establishment.  That is, it does not provide
+  resistance againt an attacker that can observe the handshake performing
```
tomato42: s/handshake performing/handshake and perform/ ?

draft-ietf-tls-tls13.md
```diff
@@ -2320,8 +2336,8 @@ this specification MUST negotiate TLS 1.2 or prior as specified in
 Servers MAY abort the handshake upon receiving a ClientHello with
 legacy_version 0x0304 or later.
 
-If this extension is present, servers MUST ignore the
-ClientHello.legacy_version value and MUST use only the
+If this extension is present in the ClientHello, servers MUST NOT the
```
tomato42: > servers MUST NOT the

??

draft-ietf-tls-tls13.md
```diff
@@ -4544,7 +4577,7 @@ bad_record_mac
 record_overflow
 : A TLSCiphertext record was received that had a length more than
   2^14 + 256 bytes, or a record decrypted to a TLSPlaintext record
-  with more than 2^14 bytes.
+  with more than 2^14 bytes (or some other configured limit).
```
tomato42: "negotiated limit" feels more appropriate here

draft-ietf-tls-tls13.md
kaduk: I'm going with a slightly different disambiguation.

draft-ietf-tls-tls13.md
ekr: So, it does seem like there might be a valid reason to provide multiple copies, or at least a coherent interpretation.

Specifically, because the semantics are all values must be present.

```
X = [A, B]
```
and
```
X=[A]
X=[B]
```

Seem equivalent.

I agree we have to resolve this, but I think MAY is the worst way, because then it creates interop problems. Either duplicates may appear or they may not, so we should say *that*

draft-ietf-tls-tls13.md
```diff
@@ -2925,6 +2944,7 @@ The server MUST NOT send a "psk_key_exchange_modes" extension.
 
 %%% Key Exchange Messages
 
+       /* Managed by IANA */
```
ekr: I'm inclined not to have this. Why do people need to know this at this point?

draft-ietf-tls-tls13.md
```diff
@@ -3155,7 +3180,9 @@ corresponding binder value (see {{psk-binder}} below). If this value is
 not present or does not validate, the server MUST abort the handshake.
 Servers SHOULD NOT attempt to validate multiple binders; rather they
 SHOULD select a single PSK and validate solely the binder that
-corresponds to that PSK. In order to accept PSK key establishment, the
+corresponds to that PSK.  There are security implications of
+validating the selected binder in this way, see \[{{client-hello-recording}}]
```
ekr: I would say instead "see [] for the security rationale for this requirement"

draft-ietf-tls-tls13.md
```diff
@@ -3240,8 +3267,8 @@ receive the server's Finished, only then sending the EndOfEarlyData
 message, followed by the rest of the handshake.
 In order to avoid deadlocks, when accepting "early_data",
 servers MUST process the client's ClientHello and then immediately
-send the ServerHello, rather than waiting for the client's
-EndOfEarlyData message.
+send their ServerHello, rather than waiting for the client's
+EndOfEarlyData message before sending their ServerHello.
```
ekr: I tend to use "its" here.

draft-ietf-tls-tls13.md
```diff
@@ -3470,11 +3499,12 @@ certificate_list
...
 : A set of extension values for the CertificateEntry. The "Extension"
   format is defined in {{extensions}}. Valid extensions for server certificates
-  include OCSP Status extension ({{RFC6066}}) and
-  SignedCertificateTimestamps ({{!RFC6962}}). Extensions in the Certificate
-  message from the server MUST correspond to one from the ClientHello message.
-  Extensions in the Certificate from the client MUST correspond with an
-  extension in the CertificateRequest message from the server.
+  at present include OCSP Status extension ({{RFC6066}}) and
+  SignedCertificateTimestamps ({{!RFC6962}}) and future extensions may
+  be defined for this message as well. Extensions in the Certificate
```
ekr: I'm not sure this is needed, but it's ungrammatical in any case.

I think you instead need a semicolon "; future extensions may"

draft-ietf-tls-tls13.md
```diff
@@ -3552,7 +3582,8 @@ The following rules apply to the certificates sent by the server:
 
 - The server's end-entity certificate's public key (and associated
   restrictions) MUST be compatible with the selected authentication
-  algorithm (currently RSA, ECDSA, or EdDSA).
+  algorithm from the client's "signature_algorithms" extension
```
ekr: This seems unnecessary.

draft-ietf-tls-tls13.md
ekr: It's not negotiated, it's just imposed. The record limit extension is an announcement

draft-ietf-tls-tls13.md
```diff
@@ -4936,9 +4959,11 @@ Where Secret is either the early_exporter_master_secret or the
...
 explicitly specified by the application. The early_exporter_master_secret is
 defined for use in settings where an exporter is needed for 0-RTT data.
-A separate interface for the early exporter is RECOMMENDED, especially
-on a server where a single interface can make the early exporter
-inaccessible.
+A separate interface for the early exporter is RECOMMENDED.  This is especially
+relevant on the server, where an implementation that processes the ClientHello
+and generates its first flight immediately would not be able to produce
+early exporter values because the key schedule has already computed the
+handshake secret and discarded the early secret.
```
ekr: The actual rationale was for security, so I would prefer not to add this text. There are lots of ways to screw up your database.

draft-ietf-tls-tls13.md
ekr: Concur.

draft-ietf-tls-tls13.md
tomato42: I read "configured" in this context as something externally configured (as in application/framework configuration), not something configured through Handshake

though I don't agree that it's imposed - if the server doesn't reply with an extension to notify the client that it understood the extension, the limit is not in force. Just the fact that the server needs to answer to extension that client send makes it a "negotiated limit", not "configured limit" or "imposed limit"

draft-ietf-tls-tls13.md
kaduk: Yes.  (I had fixed some things and then failed to actually push it up.)

draft-ietf-tls-tls13.md
kaduk: negotiated still seems better than configured, though neither is great.

draft-ietf-tls-tls13.md
kaduk: This one I agree is not needed.  I was first looking at the CertificateType enum, which we are not adding any values to, so is not otherwise mentioned in the IANA considerations and it seemed worth mentioning that IANA is involved in those values.  Here, we are creating a new registry, and that's enough mention of IANA.

draft-ietf-tls-tls13.md
kaduk: "servers" is plural, though.  But that's easy to fix

draft-ietf-tls-tls13.md
kaduk: The idea was to clarify that signature_algorithms_cert is not relevant for this choice, as opposed to in the next bullet point.

draft-ietf-tls-tls13.md
kaduk: So you want to go back to "especially on a server where a single interface can make the early exporter inaccessible", or just stop after "is RECOMMENDED."?

draft-ietf-tls-tls13.md
ekr: OK, I can live with it.

draft-ietf-tls-tls13.md
ekr: I actually am concerned about the opposite mistake :)

ekr:
Merged by hand.

PR #1167, opened 2018-03-07
prevent trivial enumeration of externally set PSKs on server

tomato42:
as I mentioned on mailing list previously (and @ekr expressed [tepid approval](https://www.ietf.org/mail-archive/web/tls/current/msg25429.html) of) current recommendation allows for enumeration of externally set identities (in case they are simple user names or other human-readable identifiers) as previously bugs like this were considered to be severe vulnerabilities

this PR aims to close off that attack

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -3114,7 +3114,9 @@ identities
 binders
 : A series of HMAC values, one for
   each PSK offered in the "pre_shared_keys" extension and in the same
-  order, computed as described below.
+  order, computed as described below. If the number of binders does not match
+  number of identities the server MUST abort the connection with
+  "illegal_parameter" alert.
 
 selected_identity
 : The server's chosen identity expressed as a (0-based) index into
@@ -3151,10 +3153,11 @@ If backwards compatibility is important, client provided, externally
 established PSKs SHOULD influence cipher suite selection.
 
 Prior to accepting PSK key establishment, the server MUST validate the
-corresponding binder value (see {{psk-binder}} below). If this value is
-not present or does not validate, the server MUST abort the handshake.
-Servers SHOULD NOT attempt to validate multiple binders; rather they
-SHOULD select a single PSK and validate solely the binder that
+corresponding binder value (see {{psk-binder}} below). If this value does not
+validate, the server MUST ignore the associated identity and continue
+connection as if PSK key establishment wasn't attempted. Servers SHOULD NOT
+attempt to validate multiple binders; rather they
+SHOULD select a single PSK and try to validate solely the binder that
 corresponds to that PSK. In order to accept PSK key establishment, the
 server sends a "pre_shared_key" extension indicating the selected
 identity.


tomato42:
while there was also a comment expressed saying that the session tickets should be self-validating, I think the existing text is sufficient:
```
ticket
: The value of the ticket to be used as the PSK identity.
  lookup key or a self-encrypted and self-authenticated value. Section
  4 of {{RFC5077}} describes a recommended ticket construction mechanism.
```

draft-ietf-tls-tls13.md
```diff
@@ -3152,10 +3152,9 @@ established PSKs SHOULD influence cipher suite selection.
 
 Prior to accepting PSK key establishment, the server MUST validate the
 corresponding binder value (see {{psk-binder}} below). If this value is
-not present or does not validate, the server MUST abort the handshake.
-Servers SHOULD NOT attempt to validate multiple binders; rather they
-SHOULD select a single PSK and validate solely the binder that
```
kaduk: There are security considerations in not attempting to validate multiple binders, as mentioned in #1166 .  Do we really want to remove this text?

ekr: Uh no.

The current algorithm is:

```
1. Select a PSK X
2. Validate the binder corresponding to X.
3. If the binder does not validate, abort the connection
4. If the binder validates, proceed with X.
````

I would be willing to have us consider replacing step 3 with

```
3. If the binder does not validate, continue the connection as if no PSK had been presented.
```

But I agree that we can't just remove this text.

**Comment by davidben** [comment]
FWIW, TLS 1.2 PSK has the same issue, though it does say:

> If the server does not recognize the PSK identity, it MAY respond
>  with an "unknown_psk_identity" alert message.  Alternatively, if the
>  server wishes to hide the fact that the PSK identity was not known,
>   it MAY continue the protocol as if the PSK identity existed but the
>  key was incorrect: that is, respond with a "decrypt_error" alert.

which is basically the (rather silly) RSA premaster mitigation and does work in TLS 1.3.

**Comment by tomato42** [comment]
> FWIW, TLS 1.2 PSK has the same issue

in TLS 1.2 the client can present just one identity so attacker can't leverage captured client hello to try to guess other identities.

That's also the reason why I added the mandatory checking of length of binders, so that you can't put a known-good at the end and see an abort if none of the previous identities match by not sending the binder for it.

**Comment by davidben** [comment]
I don't follow why a captured ClientHello is important for 1.2. The attacker can just connect, present a different identity and see if it worked. You don't need a known good value.

A known good one isn't of any use in 1.3 either because the binder incorporates the other identities too (just not the binders). Once you change the identity list, the known good binder is no longer known good.

(Ignore my claim about the RSA premaster trick being a suitable mitigation in 1.3. Not sure what I was thinking there. That doesn't work.)

**Comment by tomato42** [comment]
> I don't follow why a captured ClientHello is important for 1.2. The attacker can just connect, present a different identity and see if it worked. You don't need a known good value.

that was in relation to the changes I made to the patch, not your comment

> A known good one isn't of any use in 1.3 either because the binder incorporates the other identities too (just not the binders)

yes, you're right, for some reason I though the whole extension is truncated off for the binder calculation. That does indeed make captured client helllo's with known-good identities of less use.

it does not help us though when the attacker has one valid PSK (both identity and secret) and wants to guess other identities.

File: draft-ietf-tls-tls13.md (hunk 1 above)
Comment (tomato42) [reply draft-ietf-tls-tls13.md]: @kaduk thing is that if I have one valid PSK, I can then leverage it to enumerate other identities on the server, can't I?

Comment (kaduk) [reply draft-ietf-tls-tls13.md]: That's what we are trying to prevent by forcing the server to pick one binder to validate and only considering that one binder.  Having a known-good PSK doesn't help, as either you have the server try that one first and it works, or the server tries a different one first and it (presumably) doesn't work.  But the known-good PSK doesn't come into play at all in the second case -- you could do just as well guessing things without putting a known-good value in.

**Comment by davidben** [comment]
> it does not help us though when the attacker has one valid PSK (both identity and secret) and wants to guess other identities.

That's true, although I don't think the current state of the PR helps that case. Suppose I send:

identities = [tomato42, davidben]
binders = [garbage, binder_for_davidben]

then the server will chose no PSK identity if tomato42 is a user and davidben otherwise.

**Comment by tomato42** [comment]
exactly, though @kaduk mentioned in https://github.com/tlswg/tls13-spec/pull/1167#discussion_r172954034 that skipping PSKs in case the binder does not validate is a good idea: "There are security considerations in not attempting to validate multiple binders, as mentioned in #1166 "

Though I can't find the specific part that discusses that...

**Comment by kaduk** [comment]
> Though I can't find the specific part that discusses that...

https://github.com/tlswg/tls13-spec/pull/1166/files#diff-9d84740dcc569a0a5a359d0fba461a05R3183

**Comment by tomato42** [comment]
@kaduk yes I've noticed that part, but it just redirects to "Client Hello Recording"

and there, the most severe consequence I found is a _possibility_ of cache pollution:
```
This will cause the ClientHello to be accepted, and may cause side
effects such as replay cache pollution, although any 0-RTT data will
not be decryptable because it will use different keys.
```

isn't almost certain identity enumeration a more severe issue? and wouldn't be able to reconcile that cache pollution with continuing verification of binders?

ekr:
This is far too big a change to make in a PR after IESG approval.

I think the proposal here is OK and unless I am misunderstanding @davidben's comment, will work, but you need to get it signed off on by the chairs (presumably on-list)

seanturner:
This is way late in the process to just approve/merge.  You MUST take this to the list.

PR #1169, opened 2018-03-07
Addressing Adam's IESG comments

seanturner:


draft-ietf-tls-tls13.md
```diff
@@ -1104,8 +1104,8 @@ are many minor differences.
   be more consistent and to remove superfluous messages such as
   ChangeCipherSpec.
 
-- Elliptic curve algorithms are now in the base spec and includes new signature
-  algorithms, such as ed25519 and ed448. TLS 1.3 removed point format
+- Elliptic curve algorithms are now in the base spec and new signature
+  algorithms, such as ed25519 and ed448, are included. TLS 1.3 removed point format
```
kaduk: I think this will conflict with #1166 (but don't care which version is used).

draft-ietf-tls-tls13.md
```diff
@@ -3682,9 +3682,14 @@ attackers could obtain a signature of a message with a chosen 32-byte
...
-"TLS 1.3, server CertificateVerify" and for a client signature is
-"TLS 1.3, client CertificateVerify".
+The context string for a server signature is:
+~~~~
+"TLS 1.3, server CertificateVerify"
+~~~~
+The context string for a client signature is:
+~~~
+"TLS 1.3, client CertificateVerify"
+~~~~
```
ekr: If we are going to indent like this, then you should remove the quotes. I tend to think we should just tell RFC-ed to break the line in the convenient place.

seanturner: Can do.

PR #1171, opened 2018-03-08
Addressing Ben's IESG Comment

seanturner:


draft-ietf-tls-tls13.md
```diff
@@ -1836,7 +1836,7 @@ a "handshake_failure" or "insufficient_security" fatal alert
 ###  Client Hello
 
 When a client first connects to a server, it is REQUIRED to send the
-ClientHello as its first message. The client will also send a
+ClientHello as its first TLS Handhsake message. The client will also send a
```
kaduk: Or just its first TLS message?

seanturner: Yeah I could see that.

PR #1172, opened 2018-03-08
Addressing Spencer's IESG comments

seanturner:


draft-ietf-tls-tls13.md
```diff
@@ -5232,7 +5232,7 @@ follow. It also applies to earlier versions, which assumed these rules but did
 not document them.
 
 TLS is designed to be securely and compatibly extensible. Newer clients or
-servers, when communicating with newer peers, SHOULD negotiate the
+servers, when communicating with newer peers, MUST negotiate the
```
martinthomson: I think that the intent was to remove normative language here.

"when communicating with newer peers, negotiate common parameters based on the server's preferences"  or similar might be ideal.

Note that preference is still correct, but I'm not aware of any way in which the client's preference matters, except for it's ability to withhold options.

seanturner: If that's the case then, yeah we can drop the 2119-language because there is only one way to do it.

PR #1173, opened 2018-03-08
Forced additions to the abstract

seanturner:
The I-D checklist requires that updates and obsoletes be included in the abstract; thought RFC.  I personally feel this is [redacted] stupid, but I do not think that we should risk a process appeal to fight this battle.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -536,6 +536,9 @@ This document specifies version 1.3 of the Transport Layer Security
 (TLS) protocol.  TLS allows client/server applications to
 communicate over the Internet in a way that is designed to prevent eavesdropping,
 tampering, and message forgery.
+
+This document updates RFCs 4492, 5705, and 6066 and it obsoletes
+RFCs 5077, 5246, and 6961.
 --- middle
 
 


draft-ietf-tls-tls13.md
```diff
@@ -536,6 +536,9 @@ This document specifies version 1.3 of the Transport Layer Security
 (TLS) protocol.  TLS allows client/server applications to
 communicate over the Internet in a way that is designed to prevent eavesdropping,
 tampering, and message forgery.
+
+This document updates RFCs 4492, 5705, and 6066 and it obsoletes
+RFCs 5077, 5246, and 6961.
```
ekr: Also obsoletes 7627, no?

kaduk: We don't seem to mention that anywhere at the moment.

ekr: No, but 7627 applies only to TLS 1.2 and below, so I think transitively.

seanturner: D.1 contains:

~~~~
TLS 1.2 and prior supported an "Extended Master Secret" [RFC7627]
extension which digested large parts of the handshake transcript into
the master secret.  Because TLS 1.3 always hashes in the transcript
up to the server CertificateVerify, implementations which support
both TLS 1.3 and earlier versions SHOULD indicate the use of the
Extended Master Secret extension in their APIs whenever TLS 1.3 is
used.
~~~~

I'm worried that following this logic to the extreme we'll need to obsolete all of the RFCs listed in the 1.2 header:

From RFC 5246:
~~~~
Updated by RFC 6176, RFC 7905, RFC 5878, RFC 5746, RFC 7465, RFC 7568, RFC 7919, RFC 7507, RFC 7685, RFC 7627
Obsoletes RFC 3268, RFC 4366, RFC 4346
~~~~

But maybe we don't have to ... when you refer to a draft that's been updated you don't necessarily have to include the document that did the updating it's just assumed to be part of it (this is part of chasing references); you can include the reference if you really want to.  I think the same should hold true here - if we're obsoleting the base document (RFC5246) then everything that updates it really ought to just be obsoleting automagically.  I'm not sure it does actually work that way but maybe it ought to.

ekr: I have comet o a similar conclusion. Merging.

PR #1184, opened 2018-03-09
I wasn't happy withthe EC cleanup, so I tried Dale's approach

ekr:


draft-ietf-tls-tls13.md
```diff
@@ -2908,22 +2908,24 @@ value of the following struct:
...
+peers MUST validate each other's public value Q by ensuring
 that the point is a valid point on the elliptic curve.
 The appropriate validation procedures are defined in Section 4.3.7 of {{X962}}
 and alternatively in Section 5.6.2.3 of {{?KEYAGREEMENT=DOI.10.6028/NIST.SP.800-56Ar2}}.
-This process consists of three steps: (1) verify that the point (X, Y) is not the point at
-infinity (O), (2) verify that both X and Y are integers in the correct
-interval, (3) ensure that (X, Y) is a correct solution to the elliptic curve equation.
-For these curves, implementors do not need to verify membership in the correct subgroup.
+This process consists of three
+steps: (1) verify that Q is not the point at infinity (O), (2) verify
```
martinthomson: Make this a real list.

ekr: I can do this, but it's the same non-list as before :)

PR #1188, opened 2018-03-18
Explain identity exposure attack

ekr:
My sense of the discussion on the list (and talking to the chairs) is that there is not consensus to make a change. This just attempts to document the situation.

### DIFF
--- a/draft-ietf-tls-tls13.md
+++ b/draft-ietf-tls-tls13.md
@@ -6370,6 +6370,20 @@ server-to-client encryption keys because that would entail
 the reuse of those keys. This parallels the use of the early
 application traffic keys only in the client-to-server direction.
 
+## PSK Identity Exposure
+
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-tls13.md
```diff
@@ -6370,6 +6370,20 @@ server-to-client encryption keys because that would entail
 the reuse of those keys. This parallels the use of the early
 application traffic keys only in the client-to-server direction.
 
+## PSK Identity Exposure
+
+Because implementations respond to an invalid PSK binder by aborting
+the handshake, it may be possible for an attacker to verify whether
+a given PSK identity is valid. Specifically, if a server accepts
+both external PSK and certificate-based handshakes, a valid identity
```
kaduk: I might say "valid PSK identity" but it's probably okay as-is.

PR #1194, opened 2018-07-17
Fix minor mistakes, extension name

tniessen:
I am not sure whether the specification still permits changes, feel free to close this if not.


kaduk:
These all look like things best left to the RFC Editor staff (and this repo should effectively be considered closed at this point).

PR #1210, opened 2021-02-11
Security Property - Protection of endpoint identities

emanjon:
As discussed on the TLS list. It seems like cipher suites without confidentiality is already registered. Feel free to reformulate in any way. Refering to the cipher suites in the document is not very useful for the reader, and it does not seem to be a rule that a NULL encryption could not be recommended even if that seems very unlikely.

I also suggest adding "(certificate)" after client identity. The security property does obviously not hold for PSK authentication.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5436,7 +5436,8 @@ Protection of endpoint identities:
 : The server's identity (certificate) should be protected against passive
   attackers. The client's identity (certificate) should be protected against
   both passive and active attackers. This property does not hold for cipher
-  suites without confidentiality.
+  suites without confidentiality; while this specification does not define any such cipher suites,
+  other documents may do so.
 {:br}
 
 Informally, the signature-based modes of TLS 1.3 provide for the
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5434,8 +5434,9 @@ Key Compromise Impersonation (KCI) resistance:
 
 Protection of endpoint identities:
 : The server's identity (certificate) should be protected against passive
-  attackers. The client's identity should be protected against both passive
-  and active attackers.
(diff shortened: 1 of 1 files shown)


martinthomson:
Why not simply must be protected (or MUST)?

Being consistent about TLS 1.3 depending on the use of confidentiality protection would be better.

emanjon:
I agree that enforcing MUST would be better. I am quite bored of writing TLS profiling documents for various SDOs and products forbidding weak options. Reading RFC 8446 Appedix E, I understood the security properties as something that would be enforced in the future. I would support that interpretation of the security properties.

However:
- Now is seems like registration of cipher suites breaking one of the security properties might already have been done.
- The TLS WG has adopted draft-ietf-tls-external-psk-guidance, which breaks some of the other security properties.

richsalz:
There were "always" registrations of null ciphersuites going back to SSLv3 (and I think SSLv2).  The comments imply they were newly added, which isn't true; they were just carried forward.

In the earlier days of TLS evolution it was "BYOC" bring your own cipher.  Not true any more.

martinthomson:
TLS 1.3 deliberately broke with this history by making no pre-existing cipher suite work with it.  I think that it might be entirely reasonable to say that it assumes encryption.  Though I guess what needs to happen now is that we need to decide whether to state that assumption or to say something else.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5434,8 +5434,9 @@ Key Compromise Impersonation (KCI) resistance:
 
 Protection of endpoint identities:
 : The server's identity (certificate) should be protected against passive
-  attackers. The client's identity should be protected against both passive
-  and active attackers.
+  attackers. The client's identity (certificate) should be protected against
+  both passive and active attackers. This property does not hold for cipher
+  suites without confidentiality.
```
ekr, line 5439:
```suggestion
  suites without confidentiality; while this specification does not define any such cipher suites,
  other documents may do so.
```

ekr (approved):

PR #1217, opened 2021-02-21
Random Number Generation and Seeding

emanjon:
Does not "require" a PRNG, a TRNG works as well.

Since TLS 1.3 was published, CFRG has published RFC 8937. I think RFC 8937 is a great idea. I think it would be good if TLS 1.3 pointed implementors to that. Given the history of Dual EC, tls extended random, and that several different signal intelligence agencies have been controlling hardware security companies like Crypto AG in secret I think it is essential to not trust a single of randomness.

https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwjN39aenfvuAhWvtIsKHQ-wArgQFjABegQIAxAD&url=https%3A%2F%2Fwww.washingtonpost.com%2Fgraphics%2F2020%2Fworld%2Fnational-security%2Fcia-crypto-encryption-machines-espionage%2F&usg=AOvVaw0gGtw_W1_z-DVSlMSqi9Zr

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -93,6 +93,7 @@ informative:
   RFC7465:
   RFC7568:
   RFC7685:
+  RFC8937:
 
   SSL2:
        title: "The SSL Protocol"
@@ -5060,7 +5061,8 @@ provides several recommendations to assist implementors.
 
 ## Random Number Generation and Seeding
 
-TLS requires a cryptographically secure pseudorandom number generator (CSPRNG).
+TLS requires a cryptographically secure pseudorandom number generator (CSPRNG)
+or a true random number generator (TRNG).
 In most cases, the operating system provides an appropriate facility such
 as /dev/urandom, which should be used absent other (e.g., performance) concerns.
 It is RECOMMENDED to use an existing CSPRNG implementation in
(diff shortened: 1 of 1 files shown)


tomato42:
-1
we should not recommend use of TRNGs, they commonly have biased output and as such require whitening. If you whiten the output you get a CSPRNGs

emanjon:
The PR did not "recommend" TRNGs. I don't have any interest in using TRNGs and I don't really care. I just saw it as a correction. With your terminlogy, the current text is correct, but I am not sure TRNGs with withening is typically described as CSPRNG even if I agrre with you that they could be seen as such.

The important part of the PR is to include RFC 8937.

tomato42:
I'm ok with adding reference to RFC8937

davidben:
I don't think this is generally applicable enough to go as strong as RECOMMENDED here. For instance, it assumes you have a deterministic signature scheme available with your key. And while faulty entropy source are a bit more general of a concern, the worries about bad CSPRNG algorithms are a bit more specific. Applications that get to pick the CSPRNG algorithm can just pick one they believe in. (After all, what's the HKDF machinery in RFC8937 itself but way to consume entropy? Or really most of symmetric cryptography.)

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5060,7 +5061,8 @@ provides several recommendations to assist implementors.
 
 ## Random Number Generation and Seeding
 
-TLS requires a cryptographically secure pseudorandom number generator (CSPRNG).
+TLS requires a cryptographically secure pseudorandom number generator (CSPRNG)
+or a true random number generator (TRNG).
```
ekr, line 5065:
```suggestion
or a true random number generator (TRNG).
```

I suggest we remove this piece. The rest makes sense.

ekr (approved):

PR #1218, opened 2021-02-21
Recommending deterministic ECDSA for IoT devices without any security…

emanjon:
… considerations is not good.

Deterministic ECDSA is good in many cases but may not be the best case in accesable IoT devices.

In the last 5 years, there has been a large amount of academic papers showing that purely deterministic ECC algorithms in accesable IoT devices suffers from side-channel and fault injection attacks. For a list of papers see e.g. Section 1 of https://tools.ietf.org/html/draft-mattsson-cfrg-det-sigs-with-noise-02

Recommending deterministic ECDSA for IoT devices without any security considerations is not good...

Would be good with some further guidance. Section 1 of https://tools.ietf.org/html/draft-mattsson-cfrg-det-sigs-with-noise-02 is probably the best overview but it is a draft. Any of the paper references there would do. No standardized solutions yet.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5159,8 +5159,10 @@ Cryptographic details:
   generator (see {{random-number-generation-and-seeding}}) when generating Diffie-Hellman
   private values, the ECDSA "k" parameter, and other security-critical values?
   It is RECOMMENDED that implementations implement "deterministic ECDSA"
-  as specified in {{!RFC6979}}.
-
+  as specified in {{!RFC6979}}. Note that purely deterministic ECC signatures such as
+  deterministic ECDSA and EdDSA may be vulnerable to certain side-channel and fault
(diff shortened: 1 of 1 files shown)


tomato42:
how is this relevant to TLS? client random and server random ensure that the peers will not make a signature over the same handshake hash...

emanjon:
I think the requirement to have signatures over the the same message was specific to the Minerva attack. Side-channel attacks seems to be a general problem when the whole or parts of the signed message is known. TLS stops attacks where the attacker choses the message but does to my understanding not stop all of the published side-channel attacks on deterministic ECDSA. By using randomness, TLS protects against these kind of side-channel attacks much better than e.g. COSE. I do not know it the reamining attacks are mostly theoretical or if they are practical.

tomato42:
why you're mentioning Minerva? it's about leaking the bit size of k, it doesn't matter if the k was calculated from the deterministic algorithm or if it was selected randomly, the leak is equally destructive in either case

PR #1230, opened 2021-06-05
RFC 8996 "Deprecating TLS 1.0 and TLS 1.1"

emanjon:
RFC 8996 should be referenced like RFCs 6176 and 7568


ekr:
I provided an alternate PR in https://github.com/tlswg/tls13-spec/pull/1252

PR #1234, opened 2021-07-20
Fix a couple references to sig_algs/sig_algs_cert.

davidben:
The discussion about root certificates should mention signature_algorithms_cert. The discussion about what algorithms the key must support signing with should not, as that's the TLS signature.

(CC @sleevi)

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -1989,7 +1989,7 @@ The signatures on certificates that are self-signed or certificates that are
 trust anchors are not validated, since they begin a certification path (see
 {{RFC5280}}, Section 3.2).  A certificate that begins a certification
 path MAY use a signature algorithm that is not advertised as being supported
-in the "signature_algorithms" extension.
+in the "signature_algorithms" and "signature_algorithms_cert" extensions.
 
 Note that TLS 1.2 defines this extension differently. TLS 1.3 implementations
 willing to negotiate TLS 1.2 MUST behave in accordance with the requirements of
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc8446bis.md
```diff
@@ -3018,8 +3018,8 @@ The following rules apply to the certificates sent by the server:
 
 - The certificate MUST allow the key to be used for signing (i.e., the
   digitalSignature bit MUST be set if the Key Usage extension is present) with
-  a signature scheme indicated in the client's "signature_algorithms"/"signature_algorithms_cert"
-  extensions (see {{signature-algorithms}}).
+  a signature scheme indicated in the client's "signature_algorithms"
+  extension (see {{signature-algorithms}}).
```
kaduk, line 3022: The lead-in here is "certificates" plural sent by the server, and I think this change only makes sense if the rules only apply to the EE cert.

sleevi: The existing language also only makes sense if it applies to the EE cert (`digitalSignature` is for EEs, we'd expect `keyCertSign` for the CA certificates).

I believe "the certificate" referenced in the lead-in here is in reference to the previous bullet point ("the server's end entity certificate"), but perhaps that same language could/should just be repeated here for this bullet?

davidben: Ah yeah there are a few odd things here. I've added the word "end-entity". I've also merged in the bullet point above, since it's really talking about the same thing. It's also weird that the second bullet point believes we've already selected a signature algorithm, while the third believes we haven't yet. I didn't carry over "server's" from the second bullet point because that's already implicit from the whole section.

(I suspect that second bullet point was originally talking about the cipher suite, but we've taken the auth portions out of the cipher suites.)

Issue #1235, opened 2021-08-02
Not clear why 0-RTT doesn't provide perfect forward secrecy


cbartle891:
Section 2.3 states that "This [0-RTT] data is not forward secret, as it is encrypted solely under keys derived using the offered PSK."

However, Section 8.1 states "If the tickets are not self-contained but rather are database keys, and the corresponding PSKs are deleted upon use, then connections established using PSKs enjoy forward secrecy. This improves security for all 0-RTT data."

What does "improves security" mean here? Does it mean that the 0-RTT data is forward secret in this case? If not, why not?

kaduk:
Forward secrecy is something of a relative property; a given connection X gains forward secrecy with respect to future compromise of entity Y when secret Z is discarded.  There are further subtleties, such as if X is still active when Y is compromised then X might still actually be compromised even if Z has already been discarded, but considering all three of X, Y, and Z are necessary to make statements about forward secrecy with any precision.
The thing that we typically mean these days when we talk about "forward secrecy" without further precision is that an ephemeral (e.g., (EC)DH) key exchange mechanism is used, and we're implicitly picking one endpoint X as the subject of discussion, so that Y is the communications peer and Z is the peer's ephemeral DH value.  This requires some level of trust in the peer, since a malicious peer could fail to destroy Z or just retain the data in question in some other form.

0-RTT data is protected by a PSK associated with a session ticket, and when the server is using stateless session tickets, the PSK itself is encrypted to the server using the STEK.  So the only forward secrecy available to such 0-RTT data is when the STEK is discarded, which we typically recommend to be on an hours to days timescale (but could in principle be much longer, and there are no doubt many sites that essentially never rotate their STEK); given the lack of knowledge of the peer/server behavior, clients should not assume that they have any forward secrecy guarantee for 0-RTT data they send in this case.
When tickets are not stateless but rather are database keys, the server knowledge of the PSK used to protect the 0-RTT data is tied to the database entry for the ticket, and so the PSK can be discarded sooner than in the stateless-ticket case; in the single-use ticket case the server's copy of the PSK is expected to be destroyed when the resumed session is established, which would provide a similar timescale of forward secrecy for the client that a typical ECDH exchange would provide.  But the client still has little visibility into what strategy the server is using, and might be ill-advised to place much confidence in the forward secrecy of its offered 0-RTT data, absent out-of-band knowledge of the server's behavior.

cbartle891:
Thanks! It's much clearer to me now.

PR #1236, opened 2021-08-08
Clarifying the security properties of 0-RTT data.

nimia:
This is following issue #1235.
0-RTT data may sometimes be forward-secret, and sometimes not.
This commit attempts to clarify that in Sections 2.3 and 8.1.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -914,8 +914,8 @@ as for a 1-RTT handshake with PSK resumption.
 IMPORTANT NOTE: The security properties for 0-RTT data are weaker than
 those for other kinds of TLS data.  Specifically:
 
-1. This data is not forward secret, as it is encrypted solely under
-keys derived using the offered PSK.
+1. This data is not necessarily forward secret, as it is encrypted solely under
+keys derived using the offered PSK; see {{single-use-tickets}}.
 
 2. There are no guarantees of non-replay between connections.
 Protection against replay for ordinary TLS 1.3 1-RTT data is
@@ -4479,7 +4479,7 @@ clients implement any kind of retry behavior already needs to
 implement some sort of anti-replay defense.
 
 
-## Single-Use Tickets
+## Single-Use Tickets {#single-use-tickets}
 
 The simplest form of anti-replay defense is for the server to only
 allow each session ticket to be used once. For instance, the server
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc8446bis.md
```diff
@@ -914,8 +914,8 @@ as for a 1-RTT handshake with PSK resumption.
 IMPORTANT NOTE: The security properties for 0-RTT data are weaker than
 those for other kinds of TLS data.  Specifically:
 
-1. This data is not forward secret, as it is encrypted solely under
-keys derived using the offered PSK.
+1. This data is not necessarily forward secret, as it is encrypted solely under
+keys derived using the offered PSK; see {{single-use-tickets}}.
```
kaduk, line 918: In light of my comment on the linked issue, I'd suggest a phrasing more akin to:

The client in general must assume that this data is not forward secret: it is encrypted solely under keys derived using the offered PSK and the presented ticket inherently must allow for recovery of that PSK.  In specific circumstances, with a cooperating server, some level of forward secrecy may be provided for this data; see {{single-use-tickets}}.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -4489,8 +4489,9 @@ provided, the server would then fall back to a full handshake.
 
 If the tickets are not self-contained but rather are database keys,
 and the corresponding PSKs are deleted upon use, then connections established
-using PSKs enjoy forward secrecy. This improves security for
-all 0-RTT data and PSK usage when PSK is used without (EC)DHE.
+using PSKs enjoy not only anti-replay protection, but also forward secrecy.
```
kaduk, line 4492:
```suggestion
using PSKs enjoy not only anti-replay protection, but also forward secrecy once all copies of the PSK from the database entry have been deleted.
```

draft-ietf-tls-rfc8446bis.md
nimia: I pushed a new phrasing, hope this is better?
(The new phrasing tries to give a high-level answer as to "forward secret, yes or no", and avoids mentioning the PSK for conciseness.)

cbartle891:
This is great. Thanks so much for writing this up!

cbartle891 (approved):

kaduk (commented):
This seems okay.
I don't get excited by having the initial statement be "not necessarily forward secret" but can't put my finger on exactly why, and I don't have an alternative suggestion, so I won't object to this.

richsalz:
"forward secret" as an adjective is wierd.

cbartle891:
> I don't get excited by having the initial statement be "not necessarily forward secret" but can't put my finger on exactly why, and I don't have an alternative suggestion, so I won't object to this.

Presumably because it doesn't indicate in what circumstances it's forward secret (i.e. it could be completely random). Maybe it's more accurate to say "forward secret only under certain circumstances on the server side."

nimia:
How about this:
The protocol does not provide any forward secrecy guarantees for this data. The server's behavior determines what forward secrecy guarantees, if any, are applicable (see {{single-use-tickets}}). This behavior is not communicated to the client as part of the protocol. Therefore, absent out-of-band knowledge of the server's behavior, the client should assume that this data is not forward secret.

cbartle891:
@nimia Sounds good to me.

Issue #1245, opened 2021-11-14
Forward secrecy, long connections, and key exfiltration


emanjon:
Discussions regarding the security properties of long connections (as well as associations consisting of several connections) have come up in TSVWG when discussing RFC6083bis (DTLS/SCTP), in UTA when discussing the (D)TLS IoT profile, and in SAAG when discussing the terms forward secrecy vs. PFS.

https://mailarchive.ietf.org/arch/msg/saag/6ImeENhteXGdLsnaJHRoN6LW1zk/
https://datatracker.ietf.org/doc/draft-ietf-tsvwg-dtls-over-sctp-bis/
https://datatracker.ietf.org/doc/draft-ietf-uta-tls13-iot-profile/

RFC 8446 discusses forward secrecy from EC(DHE), KeyUpdate, and tickets/resumption and how this limits the effect of key leakage in one direction (compromise at time T2 does not compromise keys at time T1 < T2). Protection in the other direction (compromise at time T1 does not compromise keys at time T2) can be achieved by rerunning EC(DHE). If an authentication key has been compromised, rerunning EC(DHE) gives protection against passive attackers. If a traffic key has been compromised, rerunning EC(DHE) gives protection against active attackers.

Using the terms in RFC 7624, forward secrecy without rerunning EC(DHE) still allows an attacker to do static key exfiltration. Rerunning EC(DHE) forces and attacker to dynamic key exfiltration (or content exfiltration). I think dynamic key exfiltration is important to mitigate pervasive monitoring.

Frequently rerunning EC(DHE) is often recommended and used in long-term associations to limit the effects of key leakage. For IPsec, NIST SP 800-77r1 (2020) and ANSSI DAT-NT-003-EN (2015) recommends rerunning EC(DHE) every hour (if possible) and ANSSI also recommends every 100 GB of data.

I made PR #1246 with considerations and recommendations. I think RFC8446 need some considerations on this. I think RFC8446 should have recommendations, but that could maybe be more controversial and maybe even out of scope for the current RFC8446bis update. Note that while some security properties in RFC 8446 are strictly linked to a single connection, RFC 8446 Appendix E already discusses forward secrecy and key leakage between connections.

When it comes to long lived connections, considerations regarding exiery of authentication credentials and exporter_secret might also be good. Some application (at least RFC6083 DTLS/SCTP) relies on frequently exporting new secrets using the TLS-Exporter. I madea added a separate issue on the expiry of authentication credentials #1247

PR #1246, opened 2021-11-14
Forward secrecy, long connections, and key exfiltration

emanjon:
#1245


draft-ietf-tls-rfc8446bis.md
```diff
@@ -5545,6 +5546,8 @@ connections established with PSKs associated with other tickets.
...
+
```
ekr, line 5550: I think this is good except for the final recommendation which I'm wary of. I'm going to merge it without that and file an issue.

Issue #1247, opened 2021-11-14
Lifetime, expiry, and validity of certificates and PSKs


emanjon:
3GPP has a lot of use cases with (D)TLS and they keep increasing. Most of them use X509 but there are a few uses of external PSKs. The external PSKs in 3GPP have a lifetime and after that they expire, exactly like a certificate. This is common also in other uses of TLS with external PSKs. When working on this we noted that RFC8446 does not say anything about what to do when the external PSK expire. There is also no way to signal the expiry with an alert similar to certificate_expired.

When working on DTLS/SCRP for 3GPP we noted that TLS 1.3 does not have post-handshake server authentication. Mutually authenticated TLS is not used much on the Web but is very commonly used in a lot of other industries. Best practice for certificate lifetimes keeps changing towards shorter and shorter lifetimes. When working on this we noted that RFC8446 does not give any normative statement on what to do when a certificate expires. (Note that the currently suggestion in RFC6083bis is to set up new connections frequently instead so this is not a problem there).

How should the expiry of an external PSK be signaled with an alert? Would it be a new alert

`external_psk_expired: The external PSK has expired or is not currently valid.`

Or should certificate_expired be redefined?

`certificate_expired: A credential used for authentication has expired or is not `

I think TLS 1.3 also need to say something on what to do when the authentication credential expire. As far as I can see, the connection would have to be terminated unless there has been some post-handshake authentication. I would suggest something like this:

`If the certificate or external PSK last used by a party  has expired, the other party MUST terminate the connection with a "certificate_expired" alert.`

tomato42:
This area is definitely underspecified, I'd be for double duty of the `certificate_expired` alert as both the cause and the expected course of action is fundamentally the same.

But for X.509 there is another expiry date for the used credential: freshness of the revocation information (be it CRL or OCSP). I wonder if we couldn't add a new message to transmit up to date OCSP response for the certificate? Then reuse it for proof of possession of the updated PSK?

emanjon:
There were some good comments from Ilari Liusvaara and Viktor Dukhovni on the UTA list regarding responsibilities between TLS and the application as well as resumption PSKs. They point out that Certificate and PSK lifetimes is mostly the application's responsibility. Ilari points out that tickets also have lifetimes. TLS should probably have an alert indicating that the connection was closed because the ticket lifetime expired.

[...]

PR #1251, opened 2021-11-20
Expand the duty of the certificate alerts to PSKs (external and resum…

emanjon:
Expand the duty of the certificate alerts to PSKs (External and resumption)

#1247


tomato42:
+1 on the changes, but -1 overall: I think we should change those definitions together with adding information about handling expiration of certificates and PSKs while the connection is running

ekr:
I think more would be needed here because in many cases PSKs are optional (e.g., resumption). I would suggest we just have an "invalid_psk" alert but with more text.

ekr:
Closing per discussion in Vienna.

PR #1252, opened 2022-03-07
This attempts to split the difference on the 1.1 and 1.0 deprecation.

ekr:


ekr:
I think this is what we want here.

PR #1254, opened 2022-03-07
No rekey in early data. Fixes #1249

### LINKED ISSUE #1249: AEAD integrity limits for 0-RTT data
DTLS 1.3 introduces integrity limits for the AEAD algorithms, which is good. TLS 1.3 does not have any AEAD integrity limits, probably because the assumption that only a single AEAD forgery can be attempted. My understanding is that this is not true for 0-RTT data which works more like DTLS and where an attacker can attempt any number of forgery attempts on a single AEAD key.

RFC 8446 has relatively strong anti-replay protection:
   "The server MUST ensure that any instance of it (be it a machine, a thread, or any other
    entity within the relevant serving infrastructure) would accept 0-RTT for the same 0-RTT
    handshake at most once."
but this does not stop an attacker from attempting an unlimited number of forgeries.

I think consideration / mitigation of this is missing from RFC8446. Some alternative suggestions:

1. Should the 0-RTT requirement above be strengthen to:
   "The server MUST ensure that any instance of it (be it a machine, a thread, or any other
    entity within the relevant serving infrastructure) would process a 0-RTT
    handshake with the same PSK, ClientHello.random pair at most once."

2. should RFC8446bis continue to ignore AEAD limits for 0-RTT? 0-RTT data already has lower security (weaker replay protection)? In this case I think some considerations should be added describing that the integrity protection of is weaker than for normal application data.

3. should RFC8446bis introduce DTLS 1.3 type integrity limits for client_early_traffic_secret? This aligns the integrity limits of 0-RTT data in TLS with the normal application data in DTLS.

ekr:


draft-ietf-tls-rfc8446bis.md
```diff
@@ -3855,6 +3855,9 @@ safely encrypted under a given set of keys.  {{AEAD-LIMITS}} provides
 an analysis of these limits under the assumption that the underlying
 primitive (AES or ChaCha20) has no weaknesses. Implementations SHOULD
 do a key update as described in {{key-update}} prior to reaching these limits.
+Note that it is not possible to perform a KeyUpdate for early data
+and therefore implementations SHOULD not exceed the limits
+when sending early data.
```
martinthomson, line 3860: I might prefer:

> therefore implementations MUST stop sending early data before exceeding these limits.

No point in being all "SHOULD"y when it comes to this.

FWIW, I worked out that you can exceed the key usage limits with AES and early data.  You need an awful lot of it, but it is at least theoretically possible with one record per byte.  2^32 bytes > 2^24.5 records.

ekr:
The reason for the SHOULD is that the current text says: 
> primitive (AES or ChaCha20) has no weaknesses. Implementations SHOULD
do a key update as described in {{key-update}} prior to reaching these limits.
Note that it is not possible to perform a KeyUpdate for early data
and therefore implementations SHOULD not exceed the limits
when sending early data.

We could of course change both.

PR #1269, opened 2022-10-21
Errors for bogus tickets. Fixes #1247

### LINKED ISSUE #1247: Lifetime, expiry, and validity of certificates and PSKs

ekr:


ekr:
As noted in the original issue, we he have unknown_psk_identity, so I think this is unnecessary after all.

@emanjon @chris-wood ?

draft-ietf-tls-rfc8446bis.md
```diff
@@ -4102,6 +4103,13 @@ decrypt_error:
   to correctly verify a signature or validate a Finished message
   or a PSK binder.
 
+ticket_invalid:
```
martinthomson, line 4106: "ticket" has a very specific connotation that I think you might want to avoid here.  How about `psk_rejected`?  My only reservation there is that it looks like your goal is to signal that the psk *identifier* is not acceptable to the server.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -4102,6 +4103,13 @@ decrypt_error:
...
+: The server received a ticket that is invalid and is necessary
+to complete the handshake. For instance, the server might
+be configured to require PSK mode. Note that if a ticket
+is being used for session resumption, the server MAY
+ignore the ticket and perform a full handshake instead.
```
martinthomson, line 4111: Do you really want 2119 language here?  This is just how session resumption is supposed to work.

```suggestion
is being used for session resumption, the server is expected
to ignore the ticket and perform a full handshake instead.
```

draft-ietf-tls-rfc8446bis.md
```diff
@@ -4836,7 +4846,7 @@ The registries and their allocation policies are below:
   with the values from {{alert-messages-appendix}}. The
   "DTLS-OK" column is marked as "Y" for all such values.
   Values marked as "_RESERVED" have comments
-   describing their previous usage.
+   describing their previous usage
```
martinthomson, line 4849:
```suggestion
   describing their previous usage.
```

Other list items have a period.

tomato42:
Shouldn't we add a security section to that? It feels to me like it may be very easy to mess up the identity checks in a way that provides information to the attacker (Bleichenbacher style).

ekr:
Do you have some specific text to propose?

tomato42:
It's very implementation dependent, how the ticket is constructed. A ticket that's RSA encrypted, AES encrypted, or just a database key will have completely different behaviours and possible attacks, so unfortunately at best I think we can provide only something vague and generic.

Rather bad, bad something like this:
```
Note that the decryption and processing of the ticket may provide a side channel about information contained in the ticket or internal state of the server.
Implements should analyse if the particular ticket construction doesn't leak sensitive information when processed from malformed messages or unexpected sources. Using authenticated encryption over all fields of the ticket is one of recommended mitigation strategies. 
```

tomato42:
I wasn't thinking about the presence or even value of the ticket being the side channel, I was thinking about timing side-channels.

ekr:
This really seems out of scope then.

PR #1270, opened 2022-10-21
Update KeyUpdate limits with text from RFC 9147. Fixed #1257

### LINKED ISSUE #1257: Too many KeyUpdates risk key collision within a connection
In https://github.com/tlswg/dtls13-spec/pull/257 we added some text to DTLS about how rekeying too many times (2^64 is allowed with the 64-bit DTLS 1.3 epoch) risks key reuse within a connection, and accompanied it with a MUST NOT guidance to rekey more than 2^48 times (but receivers MUST NOT enforce the rule); this overrides the usual MUST-level requirement to match an "update_requested" KeyUpdate with another KeyUpdate before sending more data.

It seems that some similar considerations apply for non-D TLS as well.

ekr:


draft-ietf-tls-rfc8446bis.md
```diff
@@ -3863,14 +3873,15 @@ primitive (AES or ChaCha20) has no weaknesses. Implementations SHOULD
 do a key update as described in {{key-update}} prior to reaching these limits.
 Note that it is not possible to perform a KeyUpdate for early data
 and therefore implementations SHOULD not exceed the limits
-when sending early data.
+when sending early data.
```
martinthomson, line 3876:
```suggestion
when sending early data.
```

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3517,6 +3517,16 @@ a KeyUpdate with the old key is received before accepting any messages
 encrypted with the new key. Failure to do so may allow message truncation
 attacks.
 
+With a 128-bit key as in AES-128, rekeying 2^64 times has a high
+probability of key reuse within a given connection.  Note that even
+if the key repeats, the IV is also independently generated.  In order
```
martinthomson, line 3522: So you start by saying that rekeying has a high collision chance, but then you say that it doesn't because the nonce also has a random starting point and therefore expands the space.  This is confusing.

There are two analyses you might use here:

1. The key analysis (the first sentence), which is pretty simple.  Simplistic even.
2. The joint analysis of key + nonce, which is 224 bits wide, which the note implies.


This latter one needs to consider the number of records in each epoch as well.  If the per-epoch record limit is 2^25, then assuming a purely random key + a sequential nonce (which is not true but it should work for the purposes of rough analysis), then you can treat the space as drawing from a 199 bit space at random, so you really only need to worry about 2^99 key updates.  Rekeying only 2^48 times retains an advantage in excess of 2^50 there, so maybe that's a good approach.  And should that ever happen, nonces will be used far less (2^48 records is an unlikely number of records for a single connection, let alone 2^72.5), so the odds of collision are probably much lower still.

So I would probably drop this note, except to *maybe* say - at the end - while the risk of collision for a key is low, the use of random base nonces further reduces the risk of collision for record protection, even where keys do happen to be reused.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3517,6 +3517,16 @@ a KeyUpdate with the old key is received before accepting any messages
 encrypted with the new key. Failure to do so may allow message truncation
 attacks.
 
+With a 128-bit key as in AES-128, rekeying 2^64 times has a high
```
martinthomson, line 3520:
```suggestion
With a 128-bit key as in AES-128, rekeying 2<sup>64</sup> times has a high
```

You might like to do this throughout.  It does read better (in HTML) and the text output remains the same.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3517,6 +3517,16 @@ a KeyUpdate with the old key is received before accepting any messages
...
+probability of key reuse within a given connection.  Note that even
+if the key repeats, the IV is also independently generated.  In order
+to provide an extra margin of security, sending implementations MUST
+NOT allow the epoch to exceed 2^48-1.  In order to allow this value
+to be changed later, receiving implementations MUST NOT enforce this
+rule.  If a sending implementation receives a KeyUpdate with
```
martinthomson, line 3526: Well, this is all based on a predicate (128-bit keys) that might not hold true for all cipher suites.  I would instead say that this version of the specification applies this limit for all AEADs, no matter what their key size, but implementations MUST NOT enforce this limit as it could be extended in future versions of the specification.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3517,6 +3517,16 @@ a KeyUpdate with the old key is received before accepting any messages
...
+request_update set to "update_requested", it MUST NOT send its own
+KeyUpdate if that would cause it to exceed these limits and SHOULD
+instead ignore the "update_requested" flag.
```
martinthomson, line 3529: Do you need to advise that the connection be abandoned at this point, or maybe when this last set of keys expires?

ekr:
Note that this is precisely the language that appears in 9147.

draft-ietf-tls-rfc8446bis.md
ekr: Well, I don't think this text is incorrect. It says that there is a high chance of key collision which there is, so this is correct.

draft-ietf-tls-rfc8446bis.md
ekr: I don't think we need to. 9147 doesn't say anything about this.

ekr:
I made some updates, and am just merging in anticipation of the draft deadline. Feel free to file a PR if you think changes are indicated. @martinthomson

PR #1276, opened 2022-10-21
Clarify 0-RTT cache further. Fixes #1225

### LINKED ISSUE #1225: Cache timing warning
Ben Schwartz writes:

> RFC8446 (and bis) currently describe a "cache timing" attack on use of Early Data:
>
>   *  Exploiting cache timing behavior to discover the content of 0-RTT
>     messages by replaying a 0-RTT message to a different cache node
>      and then using a separate connection to measure request latency,
>      to see if the two requests address the same resource.
>
> In fact, many users of TLS (e.g. HTTPS, DNS-over-TLS) are even more vulnerable to cache-based attacks:
>* The attacker can probe the cache without triggering a cache fill (Cache-Control: only-if-cached, RD=0)
>* The attacker can observe the remaining cache lifetime, which indicates when cache fill occurred (coincident with the > replay or not), and when a resource will expire out of the cache (best time to try the attack).
>
> I think we should probably broaden the attack description. e.g.
>
>      *  Exploiting cache behavior to discover the content of 0-RTT messages
>         by locating a cache node that does not have a resource of interest
>        cached, replaying a 0-RTT message to it, and then using a separate
>         connection to check if the resource was added to the cache.
>
> I also wonder if we should strengthen the advice about replay defenses.

ekr:


ekr:
As provided by @davidben

Issue #1281, opened 2022-12-31
Add text that keys must be deleted when the connection closes


emanjon:
The document has text about deleting client_/server_application_traffic_secret_N and deleting tickets but says nothing about all the other keys and secrets as far as I can see (but I might have missed something). If the implementation store secrets and keys long term, the security properties promised by TLS 1.3 is broken. I'll make a PR with suggested text.

PR #1282, opened 2022-12-31
#1281 describe effects of creating new long-term keys

emanjon:
I think it would make sense to give more guidance on what to do when the connection closes.

#1281

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5544,7 +5544,7 @@ Forward secret with respect to long-term keys:
   the handshake is complete, this does not compromise the security of the
   session key (see {{?DOW92=DOI.10.1007/BF00124891}}), as long as the session key itself (and all material that could be used to recreate the session key) has
   been erased. In particular, the key shares, shared secrets, and keys derived in the TLS Key Schedule other than `binder_key`, `resumption_secret`, and PSKs derived from the `resumption_Secret` also need to be erased.  The forward secrecy property is not satisfied
-  when PSK is used in the "psk_ke" PskKeyExchangeMode.  Failing to erase keys or secrets intended to be ephemeral or connection-specific in effect creates additional long-term keys that must be protected, and compromise of those long-term keys (even after the handshake is complete) can result in loss of secrecy of the connection's traffic.
+  when PSK is used in the "psk_ke" PskKeyExchangeMode.  Failing to erase keys or secrets intended to be ephemeral or connection-specific in effect creates additional long-term keys that must be protected. Compromise of those long-term keys (even after the handshake is complete) can result in loss of secrecy of the connection's traffic.
 
 Key Compromise Impersonation (KCI) resistance:
 : In a mutually authenticated connection with certificates, compromising the long-term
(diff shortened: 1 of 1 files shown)


kaduk (commented):
I don't think we have strong grounds for making this a MUST-level protocol requirement.  (How does it aid interoperability?)
Can we just describe in non-normative prose the risks/hazards of failing to do so?

emanjon:
>Can we just describe in non-normative prose the risks/hazards of failing to do so?

Yes, that seems like an acceptable outcome that it likely is easy to reach consensus about. I'll change the PR to do that.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5544,7 +5544,11 @@ Forward secret with respect to long-term keys:
   the handshake is complete, this does not compromise the security of the
   session key (see {{?DOW92=DOI.10.1007/BF00124891}}), as long as the session key itself has
   been erased. The forward secrecy property is not satisfied
```
kaduk, line 5546:
```suggestion
  session key (see {{?DOW92=DOI.10.1007/BF00124891}}), as long as the session key itself (and all material that could be used to recreate the session key) has
  been erased. In particular, the key shares, shared secrets, and keys derived in the TLS Key Schedule other than `binder_key`, `resumption_secret`, and PSKs derived from the `resumption_Secret` also need to be erased.  The forward secrecy property is not satisfied
```

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5544,7 +5544,11 @@ Forward secret with respect to long-term keys:
...
-  when PSK is used in the "psk_ke" PskKeyExchangeMode.
+  when PSK is used in the "psk_ke" PskKeyExchangeMode. If key shares, shared secrets,
+  or any keys derived in the TLS Key Schedule except binder_key, resumption_secret,
+  and PSKs derived from the resumption_secret are not erased short-term after
+  closure of a connection this creates additional long-term keys, and the property
+  is no longer satisfied.
```
kaduk, line 5551:
```suggestion
  when PSK is used in the "psk_ke" PskKeyExchangeMode.  Failing to erase keys or secrets intended to be ephemeral or connection-specific in effect creates additional long-term keys that must be protected, and compromise of those long-term keys (even after the handshake is complete) can result in loss of secrecy of the connection's traffic.
```

kaduk (commented):
I'm not a huge fan of the way this reads; an alternate approach is outlined in the inline comments/suggestions.

draft-ietf-tls-rfc8446bis.md
martinthomson: You don't need to erase the shares, you need to erase the secrets from which those shares were generated (I don't know a name that would allow you to concisely identify those).

draft-ietf-tls-rfc8446bis.md
martinthomson: Avoid the comma splice on "protected, and compromise" -> "protected. Compromise"

draft-ietf-tls-rfc8446bis.md
emanjon: Changed to "private keys related to key shares". I think "private keys related to key shares, shared secrets" cover everything.

For DH you need to delete both private keys and the resulting shared secret. In a KEM you need to delete one private key and the shared secret.

draft-ietf-tls-rfc8446bis.md
emanjon: I fixed the comma splice

emanjon:
Ben's rewriting has been merged and the two comments from Martin has been addressed.

Issue #1287, opened 2023-01-12
Reusing psk identities enables tracking


emanjon:
I cannot find anything in the document regarding this which is surprising as reusing psk identities makes tracking trivial. The document has text for tickets but the considerations are the same for all psk identities, not just tickets.

I suggest modifying the tracking text to apply to all psk identities.

Motivation for changes is given by the charter.
```
Security and privacy goals will place emphasis on the following:

- Identify and mitigate other (long-term) user tracking or fingerprinting
vectors enabled by TLS deployments and implementations.
```

PR #1288, opened 2023-01-12
 Reusing psk identities enables client and server tracking

emanjon:
#1287

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5247,8 +5247,8 @@ Cryptographic details:
 
 ## Client Tracking Prevention {#client-tracking}
 
-Clients SHOULD NOT reuse a ticket for multiple connections. Reuse
-of a ticket allows passive observers to correlate different connections.
+Clients SHOULD NOT reuse a psk identity such as a ticket for multiple connections. Reuse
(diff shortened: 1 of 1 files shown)


emanjon:
#1289
Added server tracking by changing the heading. The rest of the text talks about connections and therefore apply to both client and server tracking.

kaduk (commented):
If we're going to strongly disrecommend (SHOULD NOT is essentially "MUST NOT unless you have a good excuse), we should probably have some plausible, generic, strategy for an alternative.  For connections that use actual pre-shared keys, how might you propose that they produce non-reused psk identifiers when the actual shared key does not change?

martinthomson:
A PSK identifier could be a nonce (random), plus an encryption of the true identifier, using that nonce as input to the key derivation process.  There are options for similar protections that use something like FPE (such as [this](https://quicwg.org/load-balancers/draft-ietf-quic-load-balancers.html#section-4)), but they might not be strong enough to pass serious scrutiny as a generic mechanism.

ekr:
I agree that this is a real issue but I also think that the SHOULD NOT is too strong given the state of the art. I don't really think that a SHOULD NOT is that great. PR #1294 takes a more descriptive approach.

ekr:
I am closing based on PR #1294.

PR #1292, opened 2023-01-27
same certificate as both server and client cert

emanjon:
Suggestion to add explicit text that this is allowed. Otherwise people might wonder if it is secure to do so.

https://mailarchive.ietf.org/arch/msg/tls/5MlYCijn65C4yZ6SiFI5xLOlX8Q/

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5164,8 +5164,8 @@ always be verified to ensure proper signing by a trusted certificate authority
 Users should be able to view information about the certificate and trust anchor.
 Applications SHOULD also enforce minimum and maximum key sizes. For example,
 certification paths containing keys or signatures weaker than 2048-bit RSA or
-224-bit ECDSA are not appropriate for secure applications.
-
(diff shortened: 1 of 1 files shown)


kaduk:
Surely we should mention Drucker, N. and S. Gueron, "Selfie: reflections on TLS 1.3 with PSK", DOI 10.1007/s00145-021-09387-y, May 2021, <https://eprint.iacr.org/2019/347.pdf>. if we are to admit the possibility of using the same certificate for client and server connections

emanjon:
@kaduk I would also be fine to specify that is shall not be done, but I think TLS 1.3 should say something. Seems to be done already and seems to be provable secure if used with only TLS 1.3. Could limited the MAY to TLS 1.3 or higher.

What would the mention of "Selfie: reflections on TLS 1.3 with PSK" say more concretely? My understanding is that the paper is talking about PSK and while certificates in TLS have public signature keys.

ekr:
I don't think the MAY helps here. I think instead we should be clear that we don't have a lot of formal analysis on this topic.

@karthikbhargavan is anything known about this?

ekr:
Closing based on alternate PR#1300

PR #1317, opened 2023-06-15
Make replay attack wording consistent.

sayrer:
Consistent phrasing in the Replay Attack section.


ekr:
I don't actually think that this is an improvement. In the second case, attackers actually *can* mount this attack, but in the first two cases, they are potential attacks. I don't think the parallel structure is necessary.

PR #1318, opened 2023-06-15
Remove em dashes.

sayrer:
Rewritten, as Ekr indicated this might be worth looking at. Not attached to my solution.


ekr:
Upon reflection I feel that the current text is fine.

PR #1319, opened 2023-06-15
Wire format compatibility.

sayrer:
Fixes "negotiation" and comes with a slight rewrite. Not attached to my solution.

PR #1321, opened 2023-07-07
MT's proposed change. Fixes #1310. Fixes #1319

### LINKED ISSUE #1310: Simplify TLS 1.1 fallback
> If negotiating TLS 1.1 or below, TLS 1.3 servers MUST, and TLS 1.2 servers SHOULD, set the last 8 bytes of their ServerHello.Random value to the bytes: 
>>  44 4F 57 4E 47 52 44 00
> Note that [RFC8996] and Appendix E.5 forbid the negotation of TLS versions below 1.2; implementations which do not follow that guidance MUST behave as described above.

I think that this would be better phrased as:

> RFC 8996 [RFC8996] and Appendix E.5 forbid the negotiation of TLS versions below 1.2.  However, server implementations which do not follow that guidance MUST set the last 8 bytes of their ServerHello.random value to the bytes:
>> 44 ...

Notes:
1. I think that it is `ServerHello.Random` and not `ServerHello.random`.
2. I don't think that we should loosen the mandate for TLS 1.2 servers.  Servers that implement TLS 1.2 can simply choose to ignore this specification (as they already did prior to us starting work on TLS 1.3.

ekr:


martinthomson (approved):

sayrer:
It's "negotiation" ...as raised in #1319. Maybe this typo is just being missed?

chris-wood (approved):

draft-ietf-tls-rfc8446bis.md
```diff
@@ -1563,16 +1563,15 @@ Random value to the bytes:
 
       44 4F 57 4E 47 52 44 01
 
-If negotiating TLS 1.1 or below, TLS 1.3 servers MUST, and TLS 1.2
-servers SHOULD, set the last 8 bytes of their ServerHello.Random value to the
+{{RFC8996}} and {{backward-compatibility-security}} forbid
+the negotation of TLS versions below 1.2. However, server
```
ekr, line 1567:
```suggestion
the negotiation of TLS versions below 1.2. However, server
```

PR #1325, opened 2023-07-10
Recommend not using legible identities. Fixes #1308

### LINKED ISSUE #1308: Requirements or guidance to not use external psk identities like john.mattsson@ericsson.com
How the external PSK identity is chosen matters a lot.

- An identifiers such as john.mattsson@ericsson.com can be used to trivially identify someone.
- A random identifier such as dxv1hwqvf6 can typically only be used for tracking/fingerprinting. But if the person behind one of the connections is identified than the attacker knows the identity behind all the connections.
- A random identifier such as dxv1hwqvf6 that is only used once cannot be used for tracking/fingerprinting.

As far as I can see the TLS specification does not say anything regarding psk identities like john.mattsson@ericsson.com. Modern EAP methods such as EAP-TLS 1.3 and EAP-AKA FS requires a "privacy-friendly" identifier. TLS 1.3 should say something about identifiers like john.mattsson@ericsson.com that can be used to identity someone. SHOULD NOT or MUST NOT seems appropriate.

ekr:


ekr:
Fixed

ekr:
@chris-wood @martinthomson @emanjon

sayrer:
Doesn't look like the "Hello Encrypted Client Hello" thing is fixed. Maybe just not pushed?

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5327,6 +5327,15 @@ Clients and Servers SHOULD NOT reuse a key share for multiple connections. Reuse
...
 of a client key share to the same server additionally allows the server to correlate different connections.
 
+It is RECOMMENDED that the labels for external identities be selected so that they
+do not provide additional information about the identity of the
+user. For instance, if the label includes an e-mail address, then
+this trivially identifies the user to a passive attacker
+(unlike the client's Certificate, which is encrypted). There are a number of potential
+ways to avoid this risk, including (1) using random identity labels
+(2) pre-encrypting the identity under a key known to the server or (3)
+using the Hello Encrypted Client Hello {{?I-D.ietf-tls-esni}} extension.
```
chris-wood, line 5337:
```suggestion
using the Encrypted Client Hello {{?I-D.ietf-tls-esni}} extension.
```

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5327,6 +5327,15 @@ Clients and Servers SHOULD NOT reuse a key share for multiple connections. Reuse
 of a key share allows passive observers to correlate different connections. Reuse
 of a client key share to the same server additionally allows the server to correlate different connections.
 
+It is RECOMMENDED that the labels for external identities be selected so that they
+do not provide additional information about the identity of the
+user. For instance, if the label includes an e-mail address, then
+this trivially identifies the user to a passive attacker
+(unlike the client's Certificate, which is encrypted). There are a number of potential
```
richsalz, line 5334: i would remove the parens here as this is not really an "aside."

martinthomson (approved):

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5327,6 +5327,15 @@ Clients and Servers SHOULD NOT reuse a key share for multiple connections. Reuse
 of a key share allows passive observers to correlate different connections. Reuse
 of a client key share to the same server additionally allows the server to correlate different connections.
 
+It is RECOMMENDED that the labels for external identities be selected so that they
```
kaduk, line 5330: FWIW, the actual protocol element is the "identity", which we describe as "a label for a key.  For instance...a label for a pre-shared key established externally".  We do use the phrasing "PSK identity" in the subsequent paragraph, so maybe "the identities used to identify (label) externally shared keys" if we want to increase consistency?

kaduk (commented):
I might have swapped the order of the two paragraphs talking about external PSK identities, but it doesn't really matter

PR #1333, opened 2023-12-18
Privacy and PSK identifiers

emanjon:
Based on Christian Huitema discussion of potential solutions.
https://mailarchive.ietf.org/arch/msg/tls/QuKsIu1gZFDfLn1x-ZnOE_LQxyc/

Encrypting the PSK identity using mechanisms external to TLS is missing in the current text. I added as little as possible. I did not go into details of the external encryption (could be asymmetric, symmetric group key, or pairwise symmetric with trial decryption).

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5355,7 +5355,7 @@ will generally be possible for an external observer to track
 clients and/or servers across connections. Use of the
 Encrypted Client Hello {{?I-D.ietf-tls-esni}} extension can
 mitigate this risk, as can mechanisms external to TLS that
-rotate the PSK identity.
+rotate or encrypt the PSK identity.
 
 
 ## Unauthenticated Operation


ekr (approved):
This LGTM. Will give it a few days to see if anyone objects.

emanjon:
Christian Huitema pointed out in a private discussion that alternatively the PSK can be encrypted instead of the PSK ID. This is already used in real-world systems, e.g., IETF ACE. I think it would be good to mention that as well. This is very similar to the TLS internal session tickets but external to TLS.

martinthomson:
Does this need more words to describe what sort of encryption would be acceptable?

Something that included a nonce would be OK, just as something rerandomizable would.  But something that is effectively static would not achieve the stated goal.

seanturner:
@emanjon thoughts?

ekr:
I am just going to merge this minimal change. We're giving people a hint, not telling them how to design it.

PR #1334, opened 2023-12-28
Tracking by replaying a flight of 0-RTT data

emanjon:
Information that the replay attack described in Section 8 can be used for server tracking is missing.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5337,6 +5337,12 @@ applications SHOULD NOT offer tickets across connections that are meant to be
 uncorrelated. For example, {{FETCH}} defines network partition keys to separate
 cache lookups in web browsers.
 
+The attack described in {{anti-replay}} where a network attacker mounts an attack by
+replaying a flight of 0-RTT data, can also be used to determine that a server,
+located in different locations or utilizing different network addresses,
+is in fact, one and the same. The mitigations outlined in {{anti-replay}} also prevent
+this tracking attack.
+
 Clients and Servers SHOULD NOT reuse a key share for multiple connections. Reuse
 of a key share allows passive observers to correlate different connections. Reuse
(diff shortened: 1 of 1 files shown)


ekr:
@emanjon I'm not sure if this is correct.  Suppose instead that we set of servers with some distributed state system so that they behave as one even though they are physically distinct

martinthomson:
Yeah, I'm similarly unsure.

The basic idea of the attack relies on the server recognizing 0-RTT in a way that is observable, but not having anti-replay protections in place such that replay is possible.  Generally speaking, those two things are aligned.

A server can recognize that this is a 0-RTT attempt - like NSS could recognize 0-RTT attempts from other NSS server implementations - but if it recognizes the attempt to the point that it is observable beyond that, then it should also share the anti-replay state.  Having decryption keys, but not anti-replay state is a pretty big error.

emanjon:
I was mostly thinking of a IoT device that is moving around together with a person. IoT devices using TLS are often TLS servers.

ekr:
I feel like I'm still missing something, because this device will have the same certificate (or I guess maybe PSK?), right? So when you connect to it, won't you be able to tell that it's the same just from that?

I had assumed that the idea here was that you wanted to distinguish two devices which had the same identity.

ekr:
@emanjon: following up here. I'm planning to submit a clean version for pubreq by end of month and will close this then unless I hear from you.

emanjon:
Sorry that I forgot to follow up on this even even Sean poked me.
 
>feel like I'm still missing something, because this device will have the same certificate (or I guess maybe PSK?), right? So >when you connect to it, won't you be able to tell that it's the same just from that?
>
>I had assumed that the idea here was that you wanted to distinguish two devices which had the same identity.

Yes, that is correct. The attack already exists. An active attacker can identify a sever by sending a new CH (in the cert case) or by replaying a CH in the PSK case. These seems like simpler attacks achieving the same thing. I think we can close this and not do any changes.

PR #1336, opened 2024-01-17
not-same might be greater, this is less

martinthomson:
So say that, alternative to #1335.

Closes #1335.


draft-ietf-tls-rfc8446bis.md
```diff
@@ -2881,7 +2881,7 @@ servers which are authenticating with an external PSK
 MUST NOT send the CertificateRequest message either in the main handshake
 or request post-handshake authentication.
 {{RFC8773}} provides an extension to permit this,
-but has not received the level of analysis as this specification.
+but has received less analysis then this specification.
```
dconnolly, line 2884:
```suggestion
but has received less analysis than this specification.
```

seanturner (approved):

PR #1337, opened 2024-01-23
AEAD-LIMITS ref

seanturner:
I see a dead link for the ref to the AEAD-LIMITS ref. I believe this is the same paper just without the date on the 1st page.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -153,7 +153,7 @@ informative:
          ins: A. Luykx
        -
          ins: K. Paterson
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc8446bis.md
```diff
@@ -153,7 +153,7 @@ informative:
          ins: A. Luykx
        -
          ins: K. Paterson
-       target: http://www.isg.rhul.ac.uk/~kp/TLS-AEbounds.pdf
+       target: https://eprint.iacr.org/2024/051.pdf
        date: August 2017
```
ekr, line 157:
```suggestion
       date: January 2024
```

PR #1340, opened 2024-03-04
Mention hybrid key exchange for split TLS ClientHello

loganaden:
Would it be worth mentioning this specific case https://tldr.fail/, where TLS ClientHello processing is not implemented correctly ?

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -5255,7 +5255,7 @@ TLS protocol issues:
 
 - Do you correctly handle handshake messages that are fragmented to
   multiple TLS records (see {{record-layer}})? Do you correctly handle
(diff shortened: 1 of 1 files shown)


davidben:
The case in that website is not what the bullet point is talking about. The bullet point is talking about a single TLS ClientHello spread over multiple TLS _records_. That's about issues like CVE-2014-3511.

The common problem with hybrids is when a single-record ClientHello does not fit in a TCP packet.

loganaden:
Thanks for point this out. I will change the PR to have a dedicate line for the tldr issue. Is that ok @davidben ?

ekr:
This just seems like a basic failure of how the TLS stack interacts with the TCP layer. I'm not sure we need to mention it.

loganaden:
Understood @ekr . I'm closing it.

PR #1353, opened 2024-04-04
Address erratum 6820

### LINKED ISSUE #1352: Address erratum 6820
[Erratum 6820](https://www.rfc-editor.org/errata/eid6820) was motivated by this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1751157).  The fix likely requires only a modest change but it might be worthwhile as the existing text has tripped up at least two implementers.

seanturner:
Closes #1352.

Text is as suggested in [erratum](https://www.rfc-editor.org/errata/eid6820).

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -4217,8 +4217,7 @@ missing_extension:
 
 unsupported_extension:
 : Sent by endpoints receiving any handshake message containing an extension
-  known to be prohibited for inclusion in the given handshake message, or including
-  any extensions in a ServerHello or Certificate not first offered in the
+  in a ServerHello or Certificate not first offered in the
   corresponding ClientHello or CertificateRequest.
 
 unrecognized_name:


tomato42:
That makes it less clear wrt EncryptedExtensions message, I think that message should be listed explicitly if we remove the general handshake messages part.

tomato42:
Actually, I think HelloRetryRequest is in the same boat, it can't have arbitrary extensions.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -4217,8 +4217,7 @@ missing_extension:
 
 unsupported_extension:
 : Sent by endpoints receiving any handshake message containing an extension
-  known to be prohibited for inclusion in the given handshake message, or including
-  any extensions in a ServerHello or Certificate not first offered in the
+  in a ServerHello or Certificate not first offered in the
```
seanturner, line 4220:
```suggestion
  in a ServerHello, Certificate, or EncryptedExtension not first offered in the
```
@tomato42 Like so?

martinthomson: or...

```suggestion
  in a ServerHello, HelloRetryRequest, EncryptedExtensions, or Certificate not first offered in the
```

tomato42: @martinthomson yes, I think we need `ServerHello, HelloRetryRequest, EncryptedExtensions, or Certificate` to actually be exhaustive

tomato42 (commented):
:+1:

PR #1354, opened 2024-04-11
Forbid the sender from sending duplicate supported groups entries.

bob-beck:
Supported Groups is intimately tied to Key Share, where Key Share must be sent in the preference order specified by Supported Groups. 

Duplicate Key Shares for a group are already forbidden from being sent, but allowing duplicates in Supported Groups makes this a bit muddled if supported groups is permitted to be, for example,  ABA and the client sends key share B then A.  

Now, nothing sane should actually be sending duplicate supported groups in a preference order, but it's still not forbidden to do so today. I'm suggesting we just not allow this so server side implementations can reject attempts to do so.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -2245,7 +2245,8 @@ Finite Field Groups (DHE):
 {:br }
 
 Items in "named_group_list" are ordered according to the sender's
-preferences (most preferred choice first).
+preferences (most preferred choice first). The "named_group_list"
+MUST NOT contain any duplicate entries.
 
 As of TLS 1.3, servers are permitted to send the "supported_groups"
 extension to the client. Clients MUST NOT act upon any information
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -2246,7 +2246,7 @@ Finite Field Groups (DHE):
 
 Items in "named_group_list" are ordered according to the sender's
 preferences (most preferred choice first). The "named_group_list"
-MUST NOT contain any duplicate entries.  A sender MAY abort a connection
+MUST NOT contain any duplicate entries.  A recipient MAY abort a connection
 with a fatal illegal_parameter alert if it detects a duplicate entry.
 
 As of TLS 1.3, servers are permitted to send the "supported_groups"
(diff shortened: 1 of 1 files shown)


ekr:
This seems reasonable, but it's a normative change, so the chairs need to tell us how to proceed.

@seanturner

draft-ietf-tls-rfc8446bis.md
```diff
@@ -2245,7 +2245,8 @@ Finite Field Groups (DHE):
 {:br }
 
 Items in "named_group_list" are ordered according to the sender's
-preferences (most preferred choice first).
+preferences (most preferred choice first). The "named_group_list"
+MUST NOT contain any duplicate entries.
```
martinthomson, line 2249: When you specify a requirement on the sender, it always helps to specify how the receiver might react if the sender misbehaves.  In this case, I think it best to only *allow* the receiver to reject the handshake, rather than require it.  After all, I believe that some implementations are tolerant of that.

```suggestion
MUST NOT contain any duplicate entries.  A sender MAY abort a connection
with a fatal illegal_parameter alert if it detect a duplicate entry.
```

(I'm guessing the alert type here.)

More generally, do we want to do this sort of thing with all such lists, not just groups?  key exchange mode? certificate type? etc...

martinthomson (changes requested):

draft-ietf-tls-rfc8446bis.md
tomato42: yes, probably "MAY abort a connection with an `illegal_parameter` Alert"

ekr:
@seanturner  do the chairs want to entertain this?

draft-ietf-tls-rfc8446bis.md
bob-beck: Belatedly, yes, I'd be completely fine with martin's suggestion.

seanturner:
@ekr this PR is ready to merge; see [mail](https://mailarchive.ietf.org/arch/msg/tls/oI1A3tBmpyck26aV3Ysy_yfgCTw/).

draft-ietf-tls-rfc8446bis.md
bob-beck: Martin I beleatedly noticed you change says that the sender may abort the connection, it's the recipient that may abort the connection. Please ensure you are ok with this.

martinthomson: Yes, of course you are right :)

tomato42 (approved):

PR #1361, opened 2024-06-13
Unify client and server certificate selection text

davidben:
I noticed this as I was looking for where we officially wrote down the implications of the X.509 Key Usage extension. We wrote it down for the server, but forgot to for the client.

The root issue is that we unified certificate negotiation in TLS 1.3 (ClientHello/Certificate and CertificateRequest/Certificate are now nice and symmetric), but forgot to correspondingly unify some of this prose. As a result, we said the same thing in different ways, and forgot different things in either place.

This change merges the two.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -2010,7 +2010,7 @@ RSASSA-PKCS1-v1_5 algorithms:
 : Indicates a signature algorithm using RSASSA-PKCS1-v1_5 {{RFC8017}}
   with the corresponding hash algorithm as defined in {{!SHS=DOI.10.6028/NIST.FIPS.180-4}}. These values
   refer solely to signatures which appear in certificates (see
-  {{server-certificate-selection}}) and are not defined for use in signed
+  {{certificate-selection}}) and are not defined for use in signed
   TLS handshake messages, although they MAY appear in "signature_algorithms"
   and "signature_algorithms_cert" for backward compatibility with TLS 1.2.
 
@@ -2050,7 +2050,7 @@ Legacy algorithms:
   algorithms with known weaknesses, specifically SHA-1 which is used
   in this context with either (1) RSA using RSASSA-PKCS1-v1_5 or (2) ECDSA.  These values
   refer solely to signatures which appear in certificates (see
-  {{server-certificate-selection}}) and are not defined for use in
+  {{certificate-selection}}) and are not defined for use in
   signed TLS handshake messages, although they MAY appear in "signature_algorithms"
   and "signature_algorithms_cert" for backward compatibility with TLS 1.2.
   Endpoints SHOULD NOT negotiate these algorithms
@@ -2059,7 +2059,7 @@ Legacy algorithms:
   them as the lowest priority (listed after all other algorithms in
   SignatureSchemeList). TLS 1.3 servers MUST NOT offer a SHA-1 signed
   certificate unless no valid certificate chain can be produced
-  without it (see {{server-certificate-selection}}).
+  without it (see {{certificate-selection}}).
 {:br }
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc8446bis.md
```diff
@@ -3100,66 +3100,61 @@ in TLS 1.2 and below.
...
 If the server cannot produce a certificate chain that is signed only via the
 indicated supported algorithms, then it SHOULD continue the handshake by sending
-the client a certificate chain of its choice that may include algorithms
-that are not known to be supported by the client.
+a certificate chain of its choice that may include algorithms that are not known
+to be supported by the peer.
 This fallback chain SHOULD NOT use the deprecated SHA-1 hash algorithm in general,
-but MAY do so if the client's advertisement permits it,
-and MUST NOT do so otherwise.
+but MAY do so if the peers's advertisement permits it,
```
bob-beck, line 3145: Nit: it is a singular peer, so I believe this should be peer's

davidben: Ah yeah that was a typo.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3100,66 +3100,61 @@ in TLS 1.2 and below.
...
+  clients SHOULD send this extension when the server is identified by name.
+
+All certificates provided by the sender MUST be signed by a
+signature algorithm advertised by the peer, if it is able to provide such
 a chain (see {{signature-algorithms}}).
 Certificates that are self-signed
 or certificates that are expected to be trust anchors are not validated as
 part of the chain and therefore MAY be signed with any algorithm.
 
 If the server cannot produce a certificate chain that is signed only via the
```
Andrei-Popov, line 3140: "Server" or "sender"?
If the **server** cannot produce... then it SHOULD continue the handshake... not known to be supported by the **peer**.
Perhaps it should be either "server and client" or "sender and receiver" or "peer and peer".

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3100,66 +3100,61 @@ in TLS 1.2 and below.
...
-
-- The certificates MUST be signed using an acceptable signature
-  algorithm, as described in {{certificate-request}}.  Note that this
-  relaxes the constraints on certificate-signing algorithms found in
-  prior versions of TLS.
-
-- If the CertificateRequest message contained a non-empty "oid_filters"
-  extension, the end-entity certificate MUST match the extension OIDs
-  that are recognized by the client, as described in {{oid-filters}}.
+Future TLS extensions may extend this criteria to further guide this selection
```
Andrei-Popov, line 3156: This is always the case; I see no reason to spell it out.

davidben: Removed.

draft-ietf-tls-rfc8446bis.md
davidben: Ah yeah, this part is weirdly written. It's because of the client can decline to send a cert but the server can't. Rephrased to hopefully be clearer.

MikeBishop (approved):

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3100,67 +3100,63 @@ in TLS 1.2 and below.
...
-- The "server_name" {{RFC6066}} and "certificate_authorities" extensions are used to
-  guide certificate selection. As servers MAY require the presence of the "server_name"
-  extension, clients SHOULD send this extension when the server is identified by name.
+- If the peer sent a "certificate_authorities" extension, at least one of the
+  certificates in the certificate chain SHOULD be issued by one of the listed
+  CAs.
 
-All certificates provided by the server MUST be signed by a
-signature algorithm advertised by the client, if it is able to provide such
+The following rules additionally apply to certificates sent by the client:
```
richsalz, line 3121: s/rules/rule/ ?

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3100,67 +3100,63 @@ in TLS 1.2 and below.
...
+
+- If the CertificateRequest message contained a non-empty "oid_filters"
+  extension, the end-entity certificate MUST match the extension OIDs
+  that are recognized by the client, as described in {{oid-filters}}.
+
+The following rules additionally apply to certificates sent by the server:
```
richsalz, line 3127: Likewise, s/rules/rule/ ?

ilucas39 (approved):
◢💙◣            ◢💖◣
💙💙💙◣ ◢💖💖💖
◥💙💙💙💖💖💖◤
    ◥💙💙💖💖◤
         ◥💙💖◤
              ◥ ◤

draft-ietf-tls-rfc8446bis.md
davidben: Done.

draft-ietf-tls-rfc8446bis.md
davidben: Done.

PR #1366, opened 2024-11-18
Appendix on RPK misbinding attacks

ms-s:


draft-ietf-tls-rfc8446bis.md
```diff
@@ -6125,6 +6137,11 @@ analysis has been done.  Implementations can ensure safety from
 cross-protocol related output by not reusing PSKs between TLS 1.3 and
 TLS 1.2.
 
+## Misbinding when using RPKs
```
martinthomson, line 6140:
```suggestion
## Misbinding Using Raw Public Keys
```

draft-ietf-tls-rfc8446bis.md
```diff
@@ -6125,6 +6137,11 @@ analysis has been done.  Implementations can ensure safety from
...
+
+When TLS 1.3 is used with raw public keys {{RFC7250}} for peer authentication, it may be vulnerable to misbinding attacks {{MM24}}. To mitigate this risk, if only the TLS server is authenticated using the RawPublicKey certificate type, the server SHOULD enforce strict validation of the "server_name" extension received in the ClientHello to confirm that the client intended to establish a connection with the server. Alternatively, or when both the client and server authenticate using the RawPublicKey certificate type, the endpoints SHOULD perform identity verification using either the "external_id_hash" TLS extension {{RFC8844}}, or at the application layer after the TLS connection is established.
```
martinthomson, line 6142: Wrap this paragraph for consistency with the rest of the document.

martinthomson (commented):
I'm not sure what "strict validation of the "server_name" extension" is supposed to entail based on this text.

ms-s:
> I'm not sure what "strict validation of the "server_name" extension" is supposed to entail based on this text.

Good point. Perhaps: "the server SHOULD check that any name received in the "server_name" extension of the ClientHello is something that it recognizes and is willing to respond to, ensuring that the client intended to establish a connection with the server receiving the ClientHello."

danwing:
Client has no way to know if SNI is validated by the server, so only external_id_hash is a reliable check.

draft-ietf-tls-rfc8446bis.md
danwing, line 6142: In many applications, only the server identity is verified, so we want to allow the client to validate the server's identity (without requiring mutual TLS).  Suggested change below.

```suggestion
When TLS 1.3 is used with raw public keys {{RFC7250}} for peer authentication, it may be vulnerable to misbinding attacks {{MM24}}. To mitigate this risk, if only the TLS server is authenticated using the RawPublicKey certificate type, the server SHOULD enforce strict validation of the "server_name" extension received in the ClientHello to confirm that the client intended to establish a connection with the server. Alternatively, the endpoints SHOULD perform identity verification of the peer using the "external_id_hash" TLS extension {{RFC8844}}, or at the application layer after the TLS connection is established.
```

ekr:
@ms-s @martinthomson @danwing see PR #1371 for proposed revised text.

PR #1369, opened 2025-02-02
Editorial updates to references

emanjon:
Editorial updates to references

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -55,7 +55,8 @@ normative:
        author:
          org: ITU-T
        seriesinfo:
-         ISO/IEC 8824-1:2021
+         ITU-T X.690
+       target: https://www.itu.int/rec/T-REC-X.690-202102-I/en
   GCM:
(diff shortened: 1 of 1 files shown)


ekr:
I agree that the reference to [DH76] is better as informational but I think we do need a normative reference for the shared key computation in that section.. Do you have a proposal?

FWIW, `main` does compile for me, and it seems to be compiling properly in CI.

emanjon:
I think [KEYAGREEMENT] is a good reference for FFDH computation as well. It even refers to the TLS RFC719 groups. I made this change in the PR.

>FWIW, main does compile for me, and it seems to be compiling properly in CI.
Must have been some temporary CI, problem.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -93,6 +95,15 @@ informative:
...
 
+  DH76:
+       title: "New directions in cryptography"
+       author:
+       -
+         ins: W. Diffie
+       -
+         ins: M. Hellman
+       target: https://ee.stanford.edu/~hellman/publications/24.pdf
+       date: November 1976
```
martinthomson, line 106:
```suggestion
  DH76: DOI.10.1109/TIT.1976.1055638
```

This works well enough.

PR #1370, opened 2025-02-03
Remove normative references to paywalled crypto

emanjon:
Remove normative references to paywalled crypto with the open access SP 800-56A [KEYAGREEMENT], which is already referenced for ECC public key validation.

Paywalled crypto specifications is a cybersecurity risk and goes against everything the IETF stands for. People should not have to pay to implement or analyze TLS 1.3.

Removed text on parameter generation as all the curves are named curves. This is likely a left over from an earlier version of TLS.

An alternative/additional open-access reference for ECC key generation and shared secret calculation would be SECG SEC 1.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -63,13 +63,6 @@ normative:
           - ins: M. Dworkin
         seriesinfo:
           NIST: Special Publication 800-38D
-  KEYAGREEMENT:
-        title: "Recommendation for Pair-Wise Key-Establishment Schemes Using Discrete Logarithm Cryptography"
-        date: 2018-04
-        author:
-          - ins: NIST
-        seriesinfo:
-          NIST: Special Publication 800-56Ar3
 
 informative:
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc8446bis.md
```diff
@@ -2432,7 +2439,7 @@ For the curves secp256r1, secp384r1, and secp521r1,
 peers MUST validate each other's public value Q by ensuring
 that the point is a valid point on the elliptic curve.
 The appropriate validation procedures are defined in Appendix D.1 of {{ECDP}}
-and alternatively in Section 5.6.2.3 of {{?KEYAGREEMENT=DOI.10.6028/NIST.SP.800-56Ar2}}.
```
martinthomson, line 2435: You could use the same DOI method to get SP 800-56Ar3. So why not?

emanjon: Yes, after all, why not?

ekr (approved):
LGTM

PR #1371, opened 2025-02-09
Rpk misbinding

ekr:
This is a revised version of the text from @ms-s in #1366 that I think makes the point more sharply.


ms-s:
Looks good to me!

PR #1373, opened 2025-02-10
Suggest getrandom interface which is available on newer OSes.

loganaden:


draft-ietf-tls-rfc8446bis.md
```diff
@@ -5213,7 +5213,10 @@ provides several recommendations to assist implementors.
 
 TLS requires a cryptographically secure pseudorandom number generator (CSPRNG).
 In most cases, the operating system provides an appropriate facility such
-as /dev/urandom, which should be used absent other (e.g., performance) concerns.
+as /dev/urandom, which should be used absent other (e.g., performance) concerns. On 
+newer Operating Systems, the getrandom interface SHOULD be used as
+it can be used in sandboxing environments.
```
davidben, line 5218: Not an objection per se, but I dunno if we want to get into a whole digression in the spec about different APIs to access the PRNG. E.g. getrandom is Linux-specific, then there's getentropy, and then there's the mess about which versions of Linux have which early boot behavior.

Makes me wonder if we should say anything at all here.

tomato42: yes, while the intention is good, I think it's too Linux specific

davidben: Oh yeah, +1 to keeping the mention of the OS. Sorry, "say anything at all" wasn't right. I was thinking just not saying anything about the specific API, but wrote something broader.

loganaden: @davidben @kaduk would something like "On newer operating systems, other interfaces, which do not require opening a file descriptor and can be used in in sandboxing environment, SHOULD be used" ?

davidben: I don't think the TLS spec should opine about something as implementation-specific as which Linux-specific APIs are or aren't friendly for sandboxing. _Definitely_ not at the level of a SHOULD.

I mean, depending the sandboxing strategy and the TLS stack's structure, `/dev/urandom` might even be _better_ than `getrandom` for a sandbox. Imagine a syscall-filtering sandbox that hasn't been updated for `getrandom`, paired with a [capabilities-heavy](https://en.wikipedia.org/wiki/Capability-based_security) TLS stack that actually happily takes an externally-supplied `/dev/urandom` file descriptor.

(Though, yes, `getrandom` is probably generally more sandboxing-friendly than `/dev/urandom` in most cases. I wouldn't expect most TLS libraries to happily take an externally-supplied fd in lieu of `/dev/urandom`. But it's a plausible enough design to disqualify this guidance IMO. And then you've got Windows, where the APIs and sandboxing considerations are completely different.)

davidben: Also, practically speaking, the TLS library is probably calling into its cryptography library's general-purpose PRNG interface, so it's not like it's going to do anything special anyway.

martinthomson, line 5218: I can't suggest this, but here's what I'd say:

> A performant and appropriately-secure CSPRNG is provided
> by most operating systems or can be sourced from a cryptographic library.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5223,7 +5226,7 @@ TLS uses random values (1) in public protocol fields such as the
 public Random values in the ClientHello and ServerHello and (2) to
 generate keying material. With a properly functioning CSPRNG, this
 does not present a security problem, as it is not feasible to determine
-the CSPRNG state from its output. However, with a broken CSPRNG, it
+the CSPRNG state from its output. However, with a broken or compromised CSPRNG, it
```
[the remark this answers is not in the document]
seanturner: Broken seems to be sufficient. This sentence is a counterpoint to `properly functioning CSPRNG` in the previous sentence.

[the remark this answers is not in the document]
loganaden: Understood. Will remove this.

draft-ietf-tls-rfc8446bis.md
loganaden: got it.

loganaden: Would something like this be better @davidben @kaduk : A performant and appropriately-secure CSPRNG is provided by most operating systems or can be sourced from a cryptographic library.

davidben: SGTM

ekr: I'm fine with this proposed text if yo want to prepare an appropriate PR.

In addition, please send a PR adding yourself to the contributor list.

loganaden: working on it.

loganaden: @ekr  PR opened: https://github.com/tlswg/tls13-spec/pull/1375

PR #1381, opened 2025-05-29
external psk selfie attack

ms-s:
Acknowledging the Seflie attack when using external PSKs and adding pointers to RFC 9257/9258 for guidance on how to avoid.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -6140,7 +6140,7 @@ this form of attack by treating the cases where there is no
 valid PSK identity and where there is an identity but it has an
 invalid binder identically.
 
-## Sharing PSKs across protocol versions
+## Sharing PSKs Across Protocol Versions
 
 TLS 1.3 takes a conservative approach to PSKs by binding them to a
 specific KDF.  By contrast, TLS 1.2 allows PSKs to be used with any


draft-ietf-tls-rfc8446bis.md
```diff
@@ -6128,7 +6140,7 @@ this form of attack by treating the cases where there is no
 valid PSK identity and where there is an identity but it has an
 invalid binder identically.
 
-## Sharing PSKs
+## Sharing PSKs across protocol versions
```
ekr, line 6143:
```suggestion
## Sharing PSKs Across Protocol Versions
```

ekr (approved):

draft-ietf-tls-rfc8446bis.md
```diff
@@ -6142,6 +6154,19 @@ analysis has been done.  Implementations can ensure safety from
...
+one server. However, as noted in {{RFC9257}}, there are use cases where
+PSKs are shared between more than two entities. In such scenarios, in
+addition to the expected security weakness where a compromised group
+member can impersonate any other member, a malicious non-member can
+reroute handshakes between honest group members to connect them in
+unintended ways. This rerouting is a type of identity misbinding
+attack, and the Selfie attack {{Selfie}} is a special case targeting a
+group member that can act as both a TLS client and server. To avoid
+such attacks, {{RFC9257}} provides recommendations for external PSK
+usage, including the use of external PSK importers as defined in {{RFC9258}}.
```
davidben, line 6168: I'm not sure this is quite right.

Having a PSK shared between multiple servers is similar to having a private key shared between multiple servers. If the servers all play the same role, that's fine. Indeed it's quite common for TLS server deployments to be made up of multiple instances, and it *is* possible for a network attacker to rebind you from one server instance to another, at least at the TLS level. We have generally not considered this notable.

The Selfie "attack" is just a misuse of the TLS + PSK interface to treat it as a way to authenticate individual entities *within* some group, when, like TLS + signing key, it only authenticates the "group" itself.

And so the Selfie "attack" has one group member act as client and server, and yet somehow its server role is distinct role from other other group members in their server roles, such that it is upset about being rebound back to itself. This is just not what the PSK interface does.

That's why the "fix" in RFC9258 is not really a protocol change. It's just artificially making it so that every ordered pair of members of the group has a distinct PSK. (But, to the point of the original text being misleading, the distinct PSKs are still "known" to all members of the group, which suggests that's not quite the property here.) You could also "fix" it by having your application protocol say the identities within the group.

(But TBH I don't think Selfie is an attack, just a misunderstanding of TLS, and thus doesn't necessarily need to be mentioned if it'll take too long to find the right text.)

ms-s:
I do think Selfie is an attack but I agree that there is no point spending a long time to find the right text at this stage. I believe a reference to RFC 9258 would be beneficial. Does it work if we remove the reference to Selfie and leave the rest of the appendix as is? (Selfie is reference in RFC9257 in any case):

`External PSKs in TLS are designed to be known to exactly one client and one server. However, as noted in {{RFC9257}}, there are use cases where PSKs are shared between more than two entities. In such scenarios, in addition to the expected security weakness where a compromised group member can impersonate any other member, a malicious non-member can reroute handshakes between honest group members to connect them in unintended ways. This rerouting is a type of identity misbinding attack and to avoid such attacks, {{RFC9257}} provides recommendations for external PSK
usage, including the use of external PSK importers as defined in {{RFC9258}}.`

ekr:
Updated and merged by hand.

PR #1382, opened 2025-05-29
rpk misbinding and clarification

ms-s:
Changes in PR:

Section 2: Protocol Overview: It is technically incorrect to say that if the Certificate message contains a raw public key then the message will not contain a certificate. Especially when we later say "If the RawPublicKey certificate type was negotiated". Adding the X.509 makes it explicit what is intended.

Typo: remove a stray "or" in Appendix.

Add self as contributor.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -767,8 +767,8 @@ Certificate:
   authenticate with a certificate). Note that if raw
   public keys {{RFC7250}} or the cached information extension
   {{?RFC7924}} are in use, then this message will not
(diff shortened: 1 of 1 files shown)


ekr:
I'm not sure I understand the distinction you're drawing here. It won't contain any kind of certificate, X.509 or otherwise. It will contain a raw public key stuffed into a certificate payload.

ekr:
I've fixed the 8444 -> 8844 issue separately. If you want to send a separate PR adding yourself to acknowledgements, I'll add you.

ekr (approved):

PR #1388, opened 2025-05-30
First stage instead of full key schedule

muhammad-usama-sardar:
Figure in Sec. 7.1 is only first stage of key schedule (as discussed on the list)


ekr:
I addressed this slightly differently in https://github.com/tlswg/tls13-spec/pull/1390/files

PR #1390, opened 2025-05-30
Sardar comments

ekr:
@martinthomson @beurdouche

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -4356,8 +4356,8 @@ input secrets are:
   the resumption_secret value from a previous connection)
 - (EC)DHE shared secret ({{ecdhe-shared-secret-calculation}})
 
-This produces a full key derivation schedule shown in the diagram below.
-In this diagram, the following formatting conventions apply:
+This produces the overall key schedule shown in the diagram below
+({{key-schedule-diagram}}. In this diagram, the following formatting conventions apply:
 
 - HKDF-Extract is drawn as taking the Salt argument from the top and
   the IKM argument from the left, with its output to the bottom and
@@ -4371,6 +4371,10 @@ In this diagram, the following formatting conventions apply:
 Note: the key derivation labels use the string "master" even though
 the values are referred to as "main" secrets.  This mismatch is a
 result of renaming the values while retaining compatibility.
+
+Note: this does not show all the leaf keys such as the separate
+AEAD and IV keys but rather the first set of secrets derived
+from the handshake.
 
 ~~~~ aasvg
                  0
@@ -4426,6 +4430,7 @@ result of renaming the values while retaining compatibility.
                                        ClientHello...client Finished)
                                 = resumption_secret
 ~~~~
+{: #key-schedule-diagram title="Overall TLS 1.3 Key Schedule"}
 
 The general pattern here is that the secrets shown down the left side
 of the diagram are just raw entropy without context, whereas the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc8446bis.md
```diff
@@ -4356,8 +4356,8 @@ input secrets are:
   the resumption_secret value from a previous connection)
 - (EC)DHE shared secret ({{ecdhe-shared-secret-calculation}})
 
-This produces a full key derivation schedule shown in the diagram below.
-In this diagram, the following formatting conventions apply:
+This produces the overall key schedule shown in the diagram below
```
muhammad-usama-sardar, line 4359: I still have trouble with calling it "overall" which gives an impression of completeness. I still propose calling it `first stage of key schedule` because this is exactly what it is.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -4426,6 +4430,7 @@ result of renaming the values while retaining compatibility.
                                        ClientHello...client Finished)
                                 = resumption_secret
 ~~~~
+{: #key-schedule-diagram title="Overall TLS 1.3 Key Schedule"}
```
muhammad-usama-sardar, line 4433: Same as above comment.

muhammad-usama-sardar (changes requested):
"Overall" still does not seem good.

draft-ietf-tls-rfc8446bis.md
ekr: I understand your point, but the idea here is that it is a consolidated picture of that first stage, so I don't think that that change is helpful in that respect.

muhammad-usama-sardar: > it is a consolidated picture of that first stage

I agree with this. But this is not what the current caption and text is saying. Unless it says "first stage" or something similar in caption, it remains misleading.

ekr: I don't think it's misleading, but I'd be willing to use "main".

ekr: Perhaps. On further reflection, I however, I don't think that's best.

muhammad-usama-sardar: You actually agreed ([source](https://mailarchive.ietf.org/arch/msg/tls/h2TfNJ50YgbEDxc1LJRvGX4t8Zo/)). Anyway, I understand that you want to wrap it up. But would you mind sharing the reasoning? I would like to understand your reasoning and may also be fine with "main" to wrap it up. But I am obviously against anything -- including "overall" -- which gives a sense of completeness.

ekr: > You actually agreed ([source](https://mailarchive.ietf.org/arch/msg/tls/h2TfNJ50YgbEDxc1LJRvGX4t8Zo/)). 

I'm not sure what your point is.  Like I said, *on further reflection* I don't think that's best.


> Anyway, I understand that you want to wrap it up. But would you mind sharing the reasoning? I would like to understand your reasoning and may also be fine with "main" to wrap it up. But I am obviously against anything -- including "overall" -- which gives a sense of completeness.

The problem with "first stage" is it creates additional confusion. For instance, you might think of the first stage as temporal (e.g., just early secret) rather than the depth of the tree. The relevant notion here is that we generate a bunch of secrets and then we generate leaf keys out of those secrets. And you'll note that there are arbitrary confusing stages after that, such as induced by KeyUpdate. 

More generally, I don't think it's helpful to present that as if it was all one conceptual unit, because it's not. Rather, we have a modular system, and this is one module, albeit with somewhat messy cut points.

muhammad-usama-sardar: Not a very convincing argument. But go ahead with "main" to wrap it up.

PR #1394, opened 2025-06-14
Hash in Transcript-Hash

muhammad-usama-sardar:
Hash algorithm negotiated by the peers in cipher suite is also used in Transcript-Hash.


ekr:
I accepted the technical change but reverted the i.e., changes. This text was in 8446 and so has already been copy edited. If the RPC feels it needs to change in publication, we can address it then.

muhammad-usama-sardar:
> I accepted the technical change 

Thanks. That's all I care about. 
> but reverted the i.e., changes. This text was in 8446 and so has already been copy edited. If the RPC feels it needs to change in publication, we can address it then.

Sure, works for me. I just never happened to see it like that in formal writings before.

PR #1398, opened 2025-11-03
2^24.5 full-size records is a limit for keys not the connection

emanjon:
This seems like an error in the text. 2^24.5 full-size records is a limit for the set of keys, not the connection. My understanding is that you rekey inside a connection


tomato42 (approved):

ekr:
I agree with this but tagging @martinthomson for double check.

martinthomson (approved):

ekr:
Merged by hand

PR #1399, opened 2025-11-17
Remove requirement to use PSS, consistent with suggestion from DavidBen

ekr:


davidben (approved):
LGTM.

For posterity, what this does is make PSS's allowed status implicit from the semantics of the particular SignatureScheme codepoints which are active. For the codepoints defined in this document, the net effect of this PR is a no-op, because the PKCS1 algorithms are already defined as (_emphasis_ mine):

> Indicates a signature algorithm using RSASSA-PKCS1-v1_5 [[RFC8017](https://tlswg.org/tls13-spec/draft-ietf-tls-rfc8446bis.html#RFC8017)] with the corresponding hash algorithm as defined in [[SHS](https://tlswg.org/tls13-spec/draft-ietf-tls-rfc8446bis.html#SHS)]. _These values refer solely to signatures which appear in certificates (see [Section 4.4.2.2](https://tlswg.org/tls13-spec/draft-ietf-tls-rfc8446bis.html#certificate-selection)) and are not defined for use in signed TLS handshake messages_, although they MAY appear in "signature_algorithms" and "signature_algorithms_cert" for backward compatibility with TLS 1.2.

However, this opens the door for other SignatureSchemes to define other semantics, notably draft-ietf-tls-tls13-pkcs1.

ekr:
Note that this document is with the RPC, so this PR is going to be held pending Auth48

paulwouters (changes requested):
Please note this change of behaviour in the "Relationship to RFC 8446" Section as well

OR13 (approved):

boucadair (approved):

PR #1401, opened 2026-02-11
Clarify that the message sequence may change with extensions

davidben:
As previously written, it suggests the message sequence does not change at all, but it is actually (as the text suggests), the transcript of everything sent or received.

This still doesn't quite make us robust to extensions. The CertificateVerify and Finished messages are defined in terms of this "Handshake Context" that is written using phrases like

* "later of EncryptedExtensions/CertificateRequest"
* "later of server Finished/EndOfEarlyData"
* "ClientHello ... client Finished + CertificateRequest"

And then sometimes Certificate and CertificateVerify are explicitly appended to it. Interestingly, that is also a bit goofy because we don't always account for those messages not existing in the prose.

This PR leaves all that alone for now, because I haven't quite written something down for that yet. (I think the right wording is to say "hash all the messages up to but not including Finished/CertificateVerify.)

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -2995,13 +2995,16 @@ stateless HelloRetryRequest by storing just the hash of ClientHello1
 in the cookie, rather than requiring it to export the entire intermediate
 hash state (see {{cookie}}).
 
-For concreteness, the transcript hash is always taken from the
-following sequence of handshake messages, starting at the first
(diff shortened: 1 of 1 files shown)


davidben:
>  (I think the right wording is to say "hash all the messages up to but not including Finished/CertificateVerify.)

Ah hmmf. Except we kinda made a mess of post-handshake messages. That's a little awkward to describe. (Not relevant for this PR.)

PR #1403, opened 2026-02-11
Rewrite CertificateVerify and Finished transcript hash definitions

davidben:
(This is a standalone PR because GitHub is bad at stacked PRs, but you probably want to look at https://github.com/tlswg/tls13-spec/pull/1401 first.)

Rather than trying to define in terms of a funky Handshake Context and then a set of messages that are appended throughout, define this the way it is actually implemented: you stop the transcript just before the message you are working on.

One cloud darkens this simple description: post-handshake authentication acts on a funny transcript where the main transcript is forked off and then we get extra messages appended to it.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -2932,17 +2932,7 @@ sent under certain circumstances, as defined below. The Finished
 message is always sent as part of the Authentication Block.
 These messages are encrypted under keys derived from the
 \[sender]_handshake_traffic_secret,
-except for post-handshake authentication.
-
-The computations for the Authentication messages all uniformly
-take the following inputs:
-
-- The certificate and signing key to be used.
-- A Handshake Context consisting of the list of messages to be
-  included in the transcript hash.
-- A Base Key to be used to compute a MAC key.
-
-Based on these inputs, the messages then contain:
+except for post-handshake authentication:
 
 Certificate
 : The certificate to be used for authentication, and any
@@ -2951,23 +2941,11 @@ Certificate
   (including 0-RTT).
 
 CertificateVerify:
-: A signature over the value Transcript-Hash(Handshake Context, Certificate)
+: A signature over the transcript hash up to this point.
 
 Finished:
-: A MAC over the value Transcript-Hash(Handshake Context, Certificate, CertificateVerify)
+: A MAC over the transcript hash up to this point,
   using a MAC key derived from the Base Key.
-{:br}
-
-
-The following table defines the Handshake Context and MAC Base Key
-for each scenario:
-
-| Mode | Handshake Context | Base Key |
-|------|-------------------|----------|
-| Server | ClientHello ... later of EncryptedExtensions/CertificateRequest | server_handshake_traffic_secret |
-| Client | ClientHello ... later of server Finished/EndOfEarlyData | client_handshake_traffic_secret |
-| Post-Handshake | ClientHello ... client Finished + CertificateRequest | [sender]_application_traffic_secret_N |
-{: #hs-ctx-and-keys title="Authentication Inputs"}
 
 ### The Transcript Hash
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-rfc8446bis.md
```diff
@@ -3360,18 +3362,41 @@ Structure of this message:
...
+       verify_data =
+           HMAC(finished_key, Transcript-Hash(ClientHello...Previous))
+
+In this document, the transcript hash will end at:
+
+* For server Finished, Previous will be EncryptedExtensions, server
+  CertificateRequest, server Certificate, or server CertificateVerify,
+  depending on which is present and comes last.
+
+* For client Finished, Previous will be server Finished, EndofEarlyData,
```
martinthomson:
```suggestion
* For client Finished, Previous will be server Finished, EndOfEarlyData,
```

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3360,18 +3362,41 @@ Structure of this message:
        } Finished;
 
 
-The verify_data value is computed as follows:
+During the handshake, the verify_data value is computed using the
+transcript hash (see {{the-transcript-hash}}), ending at the message
+immediately preceding this Finished message:
+
+       verify_data =
+           HMAC(finished_key, Transcript-Hash(ClientHello...Previous))
```
martinthomson: Is "Previous" a good name?  "Last" might be better.  Or "end_of_transcript" or "last_in_transcript" (see, different casing to indicate it's a variable).  The title case word is too much like a real Handshake message.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -3360,18 +3362,41 @@ Structure of this message:
        } Finished;
 
 
-The verify_data value is computed as follows:
+During the handshake, the verify_data value is computed using the
+transcript hash (see {{the-transcript-hash}}), ending at the message
+immediately preceding this Finished message:
+
+       verify_data =
```
martinthomson:
```suggestion
~~~
verify_data =
```

draft-ietf-tls-rfc8446bis.md
martinthomson:
```suggestion
    HMAC(finished_key, Transcript-Hash(ClientHello...Previous))
~~~
```

martinthomson (commented):
Lots of changes, but I don't see a major problem.

ekr:
This is an interesting suggestion and it might have been a better way to write things, but I don't think we should make it now, for two reasons:

1. RFC 9261 (Exported Authenticators) explicitly uses the concept of Handshake Context (in fact, I think this may have been why we added it), so we'd be breaking that.
2. It's just a big change at this point in the process, so it's an opportunity to break things.

I think the most expedient thing to do is to take #1401, #1402, and #1404 but not #1403. This addresses the ambiguity while minimizing changes.

@davidben I'm happy to take care of that, but I wanted to see if people wanted to argue differently before I made those changes.

davidben:
> RFC 9261 (Exported Authenticators) explicitly uses the concept of Handshake Context (in fact, I think this may have been why we added it), so we'd be breaking that.

I *think* that one is fine? From a brief skim, RFC 9261 already redefines the calculations anyway, to account for its "authenticator request" thing.
https://www.rfc-editor.org/rfc/rfc9261.html#section-5.2.2-8
https://www.rfc-editor.org/rfc/rfc9261.html#name-finished

Aside: this is a much, much better Handshake Context than post-handshake auth. Post-handshake auth forces you to retain the running hash, which contains a partial block. This one falls out of the exporter secret so we don't increase per-connection memory.

> It's just a big change at this point in the process, so it's an opportunity to break things.

Totally fair. I wasn't sure whether we wanted this much surgery myself. :-)

draft-ietf-tls-rfc8446bis.md
davidben: "Last" is a bit ambiguous between "the final message" vs "the message before the current one". In this context, it seemed good to be precise. Good call about the title case. Will fix.

draft-ietf-tls-rfc8446bis.md
davidben: (GH suggestions got a bit mangled, but applied.)

davidben:
Addressed @martinthomson's comments.

muhammad-usama-sardar:
> Aside: this is a much, much better Handshake Context than post-handshake auth. 

I am not following the comparison you are trying to make here. Are you comparing the `Handshake Context` in RFC8446bis vs. `Handshake Context` in RFC9261? If so, I don't think the comparison is useful.
> Post-handshake auth forces you to retain the running hash, which contains a partial block. This one falls out of the exporter secret so we don't increase per-connection memory.

I don't understand this. Could you please explain? In particular, what exactly do you mean by "partial block"?

PR #1405, opened 2026-02-12
"(EC)DHE" -> "asymmetric key exchange"

martinthomson:
...or similar phrasing, throughout.

The use of (EC)DHE throughout really doesn't reflect actual usage any more, with the use now of KEMs for a lot of TLS usage (all, if things work out).

I know that DH is symmetric in a sense, but this seemed like the best phrasing to use. Happy to take suggestions for alternatives.

Looking at those KEM drafts, this would be helpful for them.  Otherwise, they have to explain very carefully that their shared secret goes into the slot labeled "(EC)DHE" even though none of those letters applies to their thing.

### DIFF
--- a/draft-ietf-tls-rfc8446bis.md
+++ b/draft-ietf-tls-rfc8446bis.md
@@ -572,6 +572,10 @@ specific technical changes:
   is possible to send no extensions, which results in
   length 0.
 
+- Used more generic language for the asymmetric key exchange,
+  which was previously exclusively (EC)DHE,
+  to reflect the use of KEM-based key exchange.
+
 In addition, there have been some improvements to the
 security considerations, especially around privacy.
 
@@ -673,11 +677,15 @@ termination of the connection, optionally preceded by an alert message
 
 TLS supports three basic key exchange modes:
 
-- (EC)DHE (Diffie-Hellman over either finite fields or elliptic curves)
+- asymmetric key exchange
 
-- PSK-only
+- PSK
 
-- PSK with (EC)DHE
+- combined PSK and asymmetric key exchange
+
+In this document, asymmetric key exchange includes Diffie-Hellman
+over either finite fields or elliptic curves.
+Other extensions define the use of key-encapsulation mechanisms (KEMs).
 
 {{tls-full}} below shows the basic full TLS handshake:
 
@@ -743,14 +751,14 @@ The server processes the ClientHello and determines the appropriate
 cryptographic parameters for the connection. It then responds with its
 own ServerHello ({{server-hello}}), which indicates the negotiated connection
 parameters. The combination of the ClientHello
-and the ServerHello determines the shared keys. If (EC)DHE
+and the ServerHello determines the shared keys. If asymmetric
 key establishment is in use, then the ServerHello
 contains a "key_share" extension with the server's ephemeral
 Diffie-Hellman share; the server's share MUST be in the same group as one of the
 client's shares. If PSK key establishment is
 in use, then the ServerHello contains a "pre_shared_key"
 extension indicating which of the client's offered PSKs was selected.
-Note that implementations can use (EC)DHE and PSK together, in which
+Note that implementations can use asymmetric key exchange and PSK together, in which
 case both extensions will be supplied.
 
 The server then sends two messages to establish the Server Parameters:
@@ -867,7 +875,7 @@ instead of a full handshake.
 In TLS 1.2 and below, this functionality was provided by "session IDs" and
 "session tickets" {{RFC5077}}. Both mechanisms are obsoleted in TLS 1.3.
 
-PSKs can be used with (EC)DHE key exchange to provide forward
+PSKs can be used with asymmetric key exchange to provide forward
 secrecy in combination with shared keys, or can be used alone, at the
 cost of losing forward secrecy for the application data.
 
@@ -917,7 +925,7 @@ via a PSK, it SHOULD also supply a "key_share" extension to the server to
 allow the server to decline resumption and fall back
 to a full handshake, if needed. The server responds with a "pre_shared_key"
 extension to negotiate the use of PSK key establishment and can (as shown here)
-respond with a "key_share" extension to do (EC)DHE key establishment, thus
+respond with a "key_share" extension to do asymmetric key establishment, thus
 providing forward secrecy.
 
 When PSKs are provisioned externally, the PSK identity and the KDF hash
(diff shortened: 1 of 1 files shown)


emanjon:
Good to remove (EC)DH fom the spec. "asymmetric key exchange" is a good modernization of (EC)DH but removes the important ending E which stands for ephemeral. The ephemeral part should not be removed.

"ephemeral asymmetric key exchange" would be a good replacement for (EC)DHE

martinthomson:
While I agree with you, that would need to be a consensus position.  I'm trying to stick within the painted lines.

(Apparently, we can't get consensus on the ephemeral thing and so we get specs with explicit provisions for key reuse instead.)

emanjon:
I think removing an already existing "ephemeral" is a much smaller change than removing it, but I can live with "asymmetric key exchange".

Removing "ephemeral" in RFC 8446 is maybe even an improvement. The problem today is that static keys are common in visibility solutions, while RFC 8446 only talks about ephemeral keys and do not have any security considerations about static keys. I think it make sense to allow people to register key exchange with static keys, as long as they describe the security considerations.

muhammad-usama-sardar:
> I know that DH is symmetric in a sense, but this seemed like the best phrasing to use. Happy to take suggestions for alternatives.

I think there are two things here:

1. Process of key exchange
2. The resulting secret (namely `g^xy`) 

While the process of key exchange is asymmetric, I believe (EC)DHE (namely `g^xy`) itself should be characterized as symmetric in the sense that both endpoints should end up with the same value. How about "_shared ephemeral secret_" instead of "asymmetric key exchange"?

---

Moreover, for such a change, I believe we not only need to check the security considerations carefully, but also add security considerations for KEMs.

---
> Removing "ephemeral" in RFC 8446 is maybe even an improvement. The problem today is that static keys are common in visibility solutions, while RFC 8446 only talks about ephemeral keys and do not have any security considerations about static keys. 

Unless I am misunderstanding something, from a formal analysis perspective, this seems to be a regression, rather than an improvement. All proofs to my knowledge are based on the assumption that `g^x` and `g^y` are freshly generated for each connection. Removing "ephemeral" will invalidate these proofs.

draft-ietf-tls-rfc8446bis.md
```diff
@@ -4367,7 +4375,7 @@ input secrets are:
 
 - PSK (a pre-shared key established externally or derived from
   the resumption_secret value from a previous connection)
-- (EC)DHE shared secret ({{ecdhe-shared-secret-calculation}})
+- Asymmetric shared secret ({{ecdhe-shared-secret-calculation}})
```
ekr:
```suggestion
- Asymmetric shared secret resulting from the asymmetric key establishment (see {{ecdhe-shared-secret-calculation}})
```

I think this makes things clearer because as Usama points out, the secret is actually symmetric

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5753,20 +5761,20 @@ Forward secrecy limits the effect of key leakage in one direction
 (compromise of a key at time T2 does not compromise some key at time
 T1 where T1 < T2). Protection in the other direction (compromise at
 time T1 does not compromise keys at time T2) can be achieved by
-rerunning (EC)DHE. If a long-term authentication key has been
-compromised, a full handshake with (EC)DHE gives protection against
+rerunning an asymmetric key exchange. If a long-term authentication key has been
```
ekr: I think we should use "establishment" everywhere here.

ekr (changes requested):

draft-ietf-tls-rfc8446bis.md
```diff
@@ -2508,7 +2516,7 @@ psk_ke:
 supply a "key_share" value.
 
 psk_dhe_ke:
-: PSK with (EC)DHE key establishment. In this mode,
+: PSK with asymmetric key establishment. In this mode,
```
ekr:
```suggestion
: PSK with asymmetric key establishment (the label "dhe" is historical, dating from the original version of TLS 1.3, which did not support non-DHE key key establishment algorithms). In this mode,
```

ekr:
> I think removing an already existing "ephemeral" is a much smaller change than removing it, but I can live with "asymmetric key exchange".
> 
> Removing "ephemeral" in RFC 8446 is maybe even an improvement. The problem today is that static keys are common in visibility solutions, while RFC 8446 only talks about ephemeral keys and do not have any security considerations about static keys. I think it make sense to allow people to register key exchange with static keys, as long as they describe the security considerations.

Historically, in TLS, the distinction between static and ephemeral is whether the key is in a credential versus explicitly carried in the handshake (often signed by some key which is in a credential). That's what the E in DHE conveys, i.e., that it's potentially ephemeral, not that it's a one-time key [0]. This is of course a distinct question from whether key reuse is a good idea, whether it's allowed, etc, and TLS never took a position on how long you could reuse the keys. For example, you may recall that RFC 2246 contains an RSA_EXPORT mode in which the server supplied a temporary RSA key (explicitly described as "ephemeral RSA key" in RFC 2246), and these were often reused [1].

Now that we don't support static modes (in the sense above) at all, it's not clear to me how much the E is doing for us (and I don't think people say ML-KEME), so I favor keeping the text short, lest we have to mint a new acronym like "EAKE".

In agreement with Martin, I would be in favor of simply forbidding key reuse in TLS 1.3, but it seems we don't have WG consensus for that. However, I don't think whether the term "ephemeral" appears here changes things much one way or the other.


[0] Note that RFC 4949 defines ephemeral thus: "I) /adjective/ Refers to a cryptographic key or other cryptographic parameter or data object that is short-lived, temporary, or used one time. (See: session key. Compare: static.)"
[1] https://web.archive.org/web/20150304000214/http://blog.cryptographyengineering.com/2015/03/attack-of-week-freak-or-factoring-nsa.html "You see, it turns out that generating fresh RSA keys is a bit costly. So modern web servers don't do it for every single connection. In fact, Apache mod_ssl by default will generate a single export-grade RSA key when the server starts up, and will simply re-use that key for the lifetime of that server."

draft-ietf-tls-rfc8446bis.md
```diff
@@ -5753,20 +5761,20 @@ Forward secrecy limits the effect of key leakage in one direction
...
+compromised, any handshake with asymmetric key exchange gives protection against
 active attackers. Using the terms in {{RFC7624}}, forward secrecy
-without rerunning (EC)DHE does not stop an attacker from doing static
+without rerunning asymmetric key exchange does not stop an attacker from doing static
 key exfiltration. After key exfiltration of
 application_traffic_secret_N, an attacker can e.g., passively
 eavesdrop on all future data sent on the connection including data
 encrypted with application_traffic_secret_N+1,
-application_traffic_secret_N+2, etc. Frequently rerunning (EC)DHE
+application_traffic_secret_N+2, etc. Frequently rerunning asymmetric key exchange
```
martinthomson:
```suggestion
rerunning an asymmetric key establishment. If a long-term authentication key has been
compromised, a full handshake with asymmetric key establishment gives protection against
passive attackers. If the resumption_secret has been
compromised, a resumption handshake with asymmetric key establishment gives protection
against passive attackers and a full handshake with asymmetric key establishment gives
protection against active attackers. If a traffic secret has been
compromised, any handshake with asymmetric key establishment gives protection against
active attackers. Using the terms in {{RFC7624}}, forward secrecy
without rerunning asymmetric key establishment does not stop an attacker from doing static
key exfiltration. After key exfiltration of
application_traffic_secret_N, an attacker can e.g., passively
eavesdrop on all future data sent on the connection including data
encrypted with application_traffic_secret_N+1,
application_traffic_secret_N+2, etc. Frequently rerunning asymmetric key establishment
```

thomwiggers (approved):
Seems like a great idea

ekr:
Merged by hand.

PR #1410, opened 2026-03-16
Forbid key share reuse

ekr:
Per discussion on the list. Note that I put this in the section on KeyShares and then removed the text about the impact of reuse on tracking, as it is no longer applicable.


rfc9846.md
```diff
@@ -2366,6 +2368,11 @@ KeyShareEntry values for groups not listed in the client's
 these rules and abort the handshake with an "illegal_parameter" alert
 if one is violated.
 
+Clients and Servers MUST NOT reuse a key share for multiple
+connections.  Because {{RFC8446}} permitted reuse, receiving
```
muhammad-usama-sardar, line 2372: What exactly do you mean by receiving implementations?

muhammad-usama-sardar:
[Thread](https://mailarchive.ietf.org/arch/msg/tls/jpSC_G9chvSpL34X7pH3oCKh6cE/) for reference and easier finding in future

rfc9846.md
ekr: Implementations that receive key shares. I believe this to be adequately clear.

martinthomson (approved):
Very good.  Ship it.

thomwiggers (changes requested):
The proposed changes are good for the main body. I do think we would do well to consider Nick Sullivan's clarification on the list:

> Just to be clear about the scope of this change, it only prevents literal reuse of the same share. It does not rule out implementations generating related shares from shared secret material since that is not visible to the client. This change enforces non-reuse, not independence of key shares.

This is actually an important detail when considering security modeling of TLS 1.3. Putting this clarification that we're still not truly mandating ephemeral key exchange in the security considerations may be useful for future analysts.

ekr:
Well, we don't take a position on how the key shares are generated at all, other than to say:

>  The  key_exchange values for each KeyShareEntry MUST be generated  independently.

Separately, I don't think it's helpful to use the terms "ephemeral" and "static". Historically, in TLS, "ephemeral" really meant that the key was explicit in the handshake rather than embedded in a credential. For example, TLS 1.0's "ephemeral RSA" keys were routinely reused.

thomwiggers:
I am willing to bet that if I ask 10 cryptography academics what "generated independently" means, all of them will assume you are referring to ephemeral key exchange. 

On the other hand maybe it's better to not have "this doesn't _actually_ mean you need to do ephemeral key exchange" in the document at all, to avoid giving people ideas.

ekr:
> I am willing to bet that if I ask 10 cryptography academics what "generated independently" means, all of them will assume you are referring to ephemeral key exchange.

This text doesn't refer to the relationship between KeyShares in distinct connections *at all*. It refers to the relationship between the KeyShares in the same connection.

thomwiggers:
Ah right, that was from another section of the draft. Thanks for the clarification.

muhammad-usama-sardar:
> On the other hand maybe it's better to not have "this doesn't actually mean you need to do ephemeral key exchange" in the document at all, to avoid giving people ideas.

@thomwiggers I don't find the text you quote anywhere in 8446bis or on the [list](https://mailarchive.ietf.org/arch/browse/tls/?q=%22this%20doesn%27t%20actually%20mean%20you%20need%20to%20do%20ephemeral%20key%20exchange%22). Could you please clarify which document and section you are referring to?

emanjon (approved):

seanturner:
There is consensus to make this change; see [email](https://mailarchive.ietf.org/arch/msg/tls/0CVVIw9nR9NE0Ua_4HFwtoi48eE/).

PR #1411, opened 2026-03-27
Fix record size limit reference

ekr:


ekr:
@martinthomson for obvious double check
@debcooley for awareness.

debcooley:
possibly a dumb question:  Will you add RFC8449 to the refs?  (I don't see it, but my vision isn't what it used to be)

ekr:
There are other inline references such as `{{?RFC8449}}` that automatically add it.

martinthomson (approved):

PR #1415, opened 2026-04-20
Harmonize the requirements for legacy_version. Fixes #1412

### LINKED ISSUE #1412: Inconsistent requirement on client regarding ServerHello.legacy_version
The current draft states both

> A client which receives a TLS 1.3 Server Hello with a legacy_version value not equal to 0x0303 MUST abort the handshake with an "illegal_parameter" alert.

and

> Any endpoint receiving a Hello message with ClientHello.legacy_version or ServerHello.legacy_version set to 0x0300 MUST abort the handshake with a "protocol_version" alert.

Which would seem to require sending two different alerts if 0x0300 is sent in a `ServerHello.legacy_version`. This seems unlikely to be intentional. If it is intentional it probably deserves more text. If it's not intentional the alerts should be the same for both cases. Given that RFC 8446 already required `protocol_version` be sent (for the second requirement) I'd suggest the first requirement (which does not appear in 8446) be modified to also require sending `protocol_version`.

ekr:
RFC 8446 required implementations to send a `protocol_version` alert for `legacy_version=0x0300` and allowed TLS 1.0 and TLS 1.1. This draft forbids those and locks down the `legacy_version` to 0x0303, but then required `illegal_parameter`. This change harmonizes the text and requires `protocol_version` to avoid requiring implementations to change.

I see an argument for `illegal_parameter`, but I think avoiding change is better.


ekr:
@debcooley for approval.

@martinthomson @davidben for awareness.

debcooley:
I approve....  (I'm sure there is some fancy way to do that, but this is quicker).

kaduk (approved):
I kind of remember some discussions about only rejecting 0x300 (vs rejecting everything other than 0x0303 as we now do) back in the RFC 8446 days, but I am pretty sure that any points that were raised then have since been overtaken by events.

ekr:
Note that this doesn't change whether you reject them, just the alert you send.

PR #1421, opened 2026-05-05
You don't need to support GREASE

ekr:
@emanjon @davidben


davidben (approved):
Fair enough. I would say that the GREASE codepoints are defined to be vacuously supported by RFC 8701, but also no objections to saying it explicitly. :-)

PR #1422, opened 2026-05-06
Editorial changes

emanjon:
- Fixing the inconsistencies between abstract and header:
  - adding all obsolete drafts from the abstract to the heading 
  - fixing that 8422 is not both updated and obsoleted
- Changed "Negotiated Groups" to "Supported Groups". The term "Negotiated Groups" is only used once and never again.

Eric Rescorla wrote:
>I'm now trying to recall why we did this. ISTM that given that we are
>obsoleting 5246 (already done in 8446), we should obsolete all the
>other specs that only meaningfully apply to 5246. Here's the
>list:
>
> * RFC 5077: Transport Layer Security (TLS) Session Resumption without
>Server-Side State
> * RFC 5246: The Transport Layer Security (TLS) Protocol Version 1.2
> * RFC 5705: Keying Material Exporters for Transport Layer Security (TLS)
> * RFC 6066: Transport Layer Security (TLS) Extensions: Extension
>Definitions
> * RFC 6961: The Transport Layer Security (TLS) Multiple Certificate Status
>Request Extension
> * RFC 7627: Transport Layer Security (TLS) Session Hash and Extended
>Master Secret Extension
> * RFC 8422: Elliptic Curve Cryptography (ECC) Cipher Suites for Transport
>Layer Security (TLS)
>   Versions 1.2 and Earlier

Note that 5705, 6066, and 7627 are still listed as updated and not obsoleted. My understanding from your answer above is that they should all be obsoleted.

### DIFF
--- a/rfc9846.md
+++ b/rfc9846.md
@@ -4,8 +4,8 @@ abbrev: TLS
 docname: draft-ietf-tls-rfc8446bis-latest
 submissiontype: IETF
 category: std
-updates: 5705, 6066, 7627, 8422
-obsoletes: 8446
+updates: 5705, 6066, 7627
+obsoletes: 5077, 5246, 6961, 8422, 8446
 
 v: 3
 ipr: pre5378Trust200902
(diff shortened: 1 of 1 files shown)


seanturner:
Please also add something to the last paragraph in s1 to indicate how/why RFC 8442 is obsoleted.

ekr:
Superseded by PR #1426

PR #1447, opened 2026-05-17
Editorial change: add missing parenthesis

interrato:


ekr:
This is already fixed in main

PR #1449, opened 2026-05-21
Update extensions list

ekr:
This PR updates the extension list to match the now-current extensions that are Recommended=Y.

I note that this includes a bunch of non-core extensions, but those were also in 8846, so I just followed the same nominal policy. 

@martinthomson  thoughts?

### DIFF
--- a/rfc9846.md
+++ b/rfc9846.md
@@ -1829,8 +1829,8 @@ A number of TLS messages contain tag-length-value encoded extensions structures.
            (65535)
        } ExtensionType;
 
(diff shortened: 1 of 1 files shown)


ekr:
@seanturner I think I successfully pulled in most of your proposed changes. Here are the ones I did not pull in that I know if:

- The citations to 8422 and 8449 as I think they are superfluous
- The addition of renegotiation_info. Instead I clarified things were TLS 1.3 only

I also fixed the build by adding a placeholder citation. PTAL.

seanturner (approved):
Thanks!

PR #1450, opened 2026-05-21
Update changelog to be more complete.

ekr:
@martinthomson


seanturner:
Do you also need to add NewSessionTicket.extension here because the upper bounds went from 2^16-2 to 2^16-1?

martinthomson (approved):

PR #1451, opened 2026-05-27
IANA alignment

seanturner:
Not entirely sure about all of these.

- ExtensionType presentation language (front and back):
    - `server_name`: RFC 9261 adds "CR" to the TLS 1.3 column. - Added Informative reference to RFC 9261?
    - `max_fragment_length`: IANA also references 8449.
    - `status_request`: RFC 9846 updates how to do it so we should add a reference to IANA entry.
    - `supported_algorithms`: RFC 9846 updates how to do it so we should add a reference to IANA entry.
- TLS Extensions table:
    - `supported_versions` & `psk_key_exchange_modes`: swapped order to match IANA registry order.
    - `ech_outer_extensions`: added as it is now marked Recommended.
    - `renegotiation_info`: added because it is marked as Recommended - maybe it shouldn’t be Recommended=Y?
- IANA Considerations
    - To avoid different rules for the SignatureScheme and PskKeyExchangeMode registries refer (informatively?) to RFC 9847.
    - This documents updates `status_request` and `supported_groups` so asked IANA to add a reference to this RFC.
    - In the IANA registry, the `cached_info` TLS 1.3 column is currently "-". This document makes it "CH, EE".
    - Added “CH, SH" to `renegotiation_info` TLS 1.3 column, *BUT* see earlier comment.

### DIFF
--- a/rfc9846.md
+++ b/rfc9846.md
@@ -5084,8 +5084,8 @@ alert in the TLS Alerts registry with the value given in {{alert-protocol}}.
 IANA has updated added references to this document for the status_request and
 supported_groups entries in the TLS ExtensionType Values registry.
 
-IANA has updated the TLS ExtensionType values in the "TLS 1.3" column for 
-cached_info to "CH, EE" and renegotiation_info to "CH, SH".
+IANA has updated the TLS ExtensionType value in the "TLS 1.3" column for 
+cached_info to "CH, EE".
 
 --- back
 
--- a/rfc9846.md
+++ b/rfc9846.md
@@ -1892,7 +1892,7 @@ appears, it MUST abort the handshake with an "illegal_parameter" alert.
 | quic_transport_parameters {{RFC9001}} | CH, EE |
 | ticket_request {{RFC9149}} | CH, EE|
 | ech_outer_extensions {{RFC9849}} | CH |
-| renegotiation_info {{RFC5746}} | CH, SH |
+| renegotiation_info {{RFC5746}} | - |
 {: #extensions-list title="TLS Extensions"}
 
 Note: This table includes only extensions marked
--- a/rfc9846.md
+++ b/rfc9846.md
@@ -97,6 +97,7 @@ informative:
   RFC9149:
   RFC9257:
   RFC9258:
+  RFC9261:
   RFC9345:
   RFC9849:
 
@@ -1784,11 +1785,11 @@ A number of TLS messages contain tag-length-value encoded extensions structures.
        } Extension;
 
        enum {
-           server_name(0),                             /* RFC 6066 */
-           max_fragment_length(1),                     /* RFC 6066 */
-           status_request(5),                          /* RFC 6066 */
-           supported_groups(10),                       /* RFC 8422, 7919 */
-           signature_algorithms(13),                   /* RFC 8446 */
+           server_name(0),                             /* RFC 6066, 9261 */
+           max_fragment_length(1),                     /* RFC 6066, 8449 */
+           status_request(5),                          /* RFC 9846, 6066 */
+           supported_groups(10),                       /* RFC 9846, 8422, 7919 */
+           signature_algorithms(13),                   /* RFC 9846 */
            use_srtp(14),                               /* RFC 5764 */
            heartbeat(15),                              /* RFC 6520 */
            application_layer_protocol_negotiation(16), /* RFC 7301 */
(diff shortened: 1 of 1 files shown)


rfc9846.md
```diff
@@ -5062,21 +5064,29 @@ by IANA:
...
 in the TLS ExtensionType Values registry to "extended_main_secret".
 
 IANA has created  a value for the "general_error"
 alert in the TLS Alerts registry with the value given in {{alert-protocol}}.
 
+IANA has updated added references to this document for the status_request and
+supported_groups entries in the TLS ExtensionType Values registry.
+
+IANA has updated the TLS ExtensionType values in the "TLS 1.3" column for 
+cached_info to "CH, EE" and renegotiation_info to "CH, SH".
```
davidben: I think `renegotiation_info` was correct as-is. `renegotiation_info` is a bugfix extension for TLS 1.2 (and below) renegotiation. It's not applicable to TLS 1.3 because we don't have renegotiation in the first place. (And if we did, I sure hope we'd include the bugfix from day 1!) That means:

* A TLS-1.3-_only_ client would not send it in ClientHello
* A TLS-1.2-or-1.3 client would send it, but only for the benefit of TLS 1.2 servers
* A TLS 1.3 server would never send it in ServerHello (or any other message)
* A TLS 1.3 server would ignore it in ClientHello, because it might be talking to a TLS-1.2-or-1.3 client

I think either `-` or `CH` are defensible in that case, depending on how we want to handle the 1.2-or-1.3 thing, but probably `-` is fine?

seanturner: Thanks for working through this, I'll admit I didn't. Also, I hadn't considered just putting a `-` Table to align it with the IANA tables. Let's go for the KISS approach.

seanturner:
```suggestion
IANA has updated the TLS ExtensionType value in the "TLS 1.3" column for 
cached_info to "CH, EE".
```

rfc9846.md
```diff
@@ -1874,22 +1875,24 @@ appears, it MUST abort the handshake with an "illegal_parameter" alert.
...
+| signature_algorithms_cert {{RFC9846}}|      CH, CR |
+| key_share {{RFC9846}}            | CH, SH, HRR |
 | transparency_info {{RFC9162}}    | CH, CR, CT |
 | connection_id {{RFC9146}}        | CH, SH |
 | external_id_hash {{RFC8844}}     | CH, EE |
 | external_session_id {{RFC8844}}  | CH, EE |
 | quic_transport_parameters {{RFC9001}} | CH, EE |
 | ticket_request {{RFC9149}} | CH, EE|
+| ech_outer_extensions {{RFC9849}} | CH |
+| renegotiation_info {{RFC5746}} | CH, SH |
```
seanturner:
```suggestion
| renegotiation_info {{RFC5746}} | - |
```

rfc9846.md
```diff
@@ -1784,28 +1785,28 @@ A number of TLS messages contain tag-length-value encoded extensions structures.
...
 
        enum {
-           server_name(0),                             /* RFC 6066 */
-           max_fragment_length(1),                     /* RFC 6066 */
-           status_request(5),                          /* RFC 6066 */
-           supported_groups(10),                       /* RFC 8422, 7919 */
-           signature_algorithms(13),                   /* RFC 8446 */
+           server_name(0),                             /* RFC 6066, 9261 */
+           max_fragment_length(1),                     /* RFC 6066, 8449 */
+           status_request(5),                          /* RFC 9846, 6066 */
```
ekr, line 1790: I'm not so sure about these 9846 updates. You think it should be the most recent version and not the original one?

rfc9846.md
```diff
@@ -1784,28 +1785,28 @@ A number of TLS messages contain tag-length-value encoded extensions structures.
...
+           supported_groups(10),                       /* RFC 9846, 8422, 7919 */
```
ekr, line 1791: Why do we need 9846? Is it not completely defined in 8422? That's a general question for others too

rfc9846.md
```diff
@@ -1857,11 +1858,11 @@ appears, it MUST abort the handshake with an "illegal_parameter" alert.
 
 | Extension                                |   TLS 1.3   |
 |:-----------------------------------------|------------:|
-| server_name {{RFC6066}}                    |      CH, EE |
-| max_fragment_length {{RFC6066}}            |      CH, EE |
-| status_request {{RFC6066}}                 |  CH, CR, CT |
-| supported_groups {{RFC7919}}               |      CH, EE |
-| signature_algorithms {{RFC8446}}           |      CH, CR |
+| server_name {{RFC6066}} {{RFC9261}}        |  CH, EE, CR |
+| max_fragment_length {{RFC6066}} {{RFC8449}}|      CH, EE |
```
ekr, line 1862: Why does this need a cite to 8449? That replaces this extension but doesn't redefine it AFAICT.

ekr:
@seanturner I'm still confused about the cites to 9846. Is that really the convention?

ekr:
@seanturner ping

seanturner:
Not sure it's a convention per se, but I seem to remember that RFC 8446 was supposed to be kind of a one stop shop for implementers. If RFC 9846 obsoletes RFC 8446, then RFC 9846 becomes that one stop shop and there's no need to refer to RFC 8446 in the presentation syntax ([here](https://tlswg.org/tls13-spec/rfc9846.html#name-extensions) and [here](https://tlswg.org/tls13-spec/rfc9846.html#name-key-exchange-messages-2)) or the [table of extensions](https://tlswg.org/tls13-spec/rfc9846.html#table-1).

Further, if somebody, albeit oddly, comes at TLS 1.3 via the [IANA tables](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#tls-extensiontype-values-1) and then jumps into the RFC what they will see is that the "references" do not match. In the IANA tables they will see RFC 9846 and in the RFC they will see RFC 8446 - and that could be confusing (and will invite silly errata).

My thinking follows for the other extension references:
- Adding `9261` to `server_name(0)`: The IANA tables include the reference. We could make this an informative reference.
- Adding `8449` to `max_fragment_length(1)`: The IANA tables include the reference. No need for a new reference RFC 9846 already refers to RFC 8449.
- Adding `9846` to `status_request(5)`: When I read the following in the text, I think it says "read this RFC for information about `status_request(5)` for TLS 1.3":

>It also changes how Online Certificate Status Protocol (OCSP) messages are carried and therefore updates [[RFC6066](https://tlswg.org/tls13-spec/rfc9846.html#RFC6066)] and obsoletes [[RFC6961](https://tlswg.org/tls13-spec/rfc9846.html#RFC6961)] as described in [Section 4.5.1.1](https://tlswg.org/tls13-spec/rfc9846.html#ocsp-and-sct).

- Adding `9846` to `support_groups(2)`: When I read the following in the text, I also think it says "read this RFC for information about `support_groups(2)`":

> This document supersedes and obsoletes previous versions of TLS, including version 1.2 [[RFC5246](https://tlswg.org/tls13-spec/rfc9846.html#RFC5246)], the extended master secret extension [[RFC7627](https://tlswg.org/tls13-spec/rfc9846.html#RFC7627)], and elliptic curve definitions for TLS 1.2 [[RFC8422](https://tlswg.org/tls13-spec/rfc9846.html#RFC8422)].

If you buy the last two points, then I also think we should update the IANA considerations sections to add instructions to add a reference to RFC 9846 for the `status_request(5)` and `support_groups(2)` entries.

seanturner:
Addressed by #1449.

